PR #21, opened 2021-10-01
fix: remove source maps reference from the production build

Sakchai-Refinitiv:
close: https://jira.refinitiv.com/browse/ELF-1564

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,8 +47,8 @@
         "yargs": "^17.0.1"
       },
       "engines": {
-        "node": ">=14.15.4",
-        "npm": ">=7.19.1"
+        "node": ">=15.0.0",
+        "npm": ">=7.0.0"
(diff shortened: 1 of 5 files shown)


packages/elements/scripts/cmd/build.js
```diff
@@ -10,16 +10,32 @@ exports.builder = yargs => {
       type: 'boolean',
       default: false,
       description: 'Watch file change'
+    })
+    .option('sourceMap', {
+      alias: 'sm',
```
goremikins, line 15: Do not use shorthands in this case, they do not make sense and are not the same in TS

Sakchai-Refinitiv: I agree with you. I remove it already.

goremikins (approved):

PR #32, opened 2021-10-12
feat: add role attribute to base class

wsuwt:
Adding a small utility to help the other elements set role attribute.

For example, to set role in an element. `defaultRole` will be inherited down to all elements, each element then can override its own role.

```ts
// button
class Button {
   protected readonly defaultRole = 'button';
}
```

- `role` should be string
- `role` can be override by user setting.
- `role` should reflect to the element DOM

### DIFF
--- a/packages/solar-theme/src/shared-styles/input.less
+++ b/packages/solar-theme/src/shared-styles/input.less
@@ -52,7 +52,7 @@
   &[readonly] {
     .border(@input-readonly-border-color);
   }
-  &:not([focused]):not(:focus):hover {
+  &:not([focused]):not(:focus):not([disabled]):not([readonly]):hover {
     .border(@input-hover-border-color);
   }
 }


packages/core/src/elements/BasicElement.ts
```diff
@@ -175,6 +180,10 @@ export abstract class BasicElement extends LitElement {
       this.tabIndex = this.defaultTabIndex;
     }
 
+    if (!this.hasAttribute('role') && typeof this.defaultRole === 'string') {
```
goremikins, line 183: `typeof` is up to 50% faster than `hasAttribute` as `typeof` does not need to hit the DOM tree. Can you please switch the order here and for tabindex as well

wsuwt: Thanks. It's done.

PR #35, opened 2021-10-14
overlay and overlay-menu fixes

goremikins:


packages/elements/src/overlay-menu/index.ts
```diff
@@ -473,6 +483,7 @@ export class OverlayMenu extends Overlay {
   private opening (): void {
     const parentMenuItem = OpenedMenusManager.getParentMenuItem(this);
     this.dataDisconnectThrottler.cancel();
+    this.restoreNestedProperties();
```
Theeraphat-Sorasetsakul, line 486: Why do we need this? It's already restored when the overlay closed.

packages/elements/src/overlay-menu/index.ts
```diff
@@ -433,8 +434,17 @@ export class OverlayMenu extends Overlay {
     });
 
     this.setItemHighlight();
+    this.restoreNestedProperties();
 
-    if (OpenedMenusManager.isNested(this)) {
+    super.onClosed();
```
Theeraphat-Sorasetsakul, line 439: `super.onClosed()` at Overlay did nothing. Do we need this line? And is this line duplicated to line 457 in `restoreNestedProperties` function?

PR #40, opened 2021-10-18
feat: remove default aria disabled and aria readonly set by ControlElement

wsuwt:
According to [w3](https://www.w3.org/TR/html-aria/#docconformance-attr)

- `aria-disabled` does not have any use when element is also `disabled`
- Screen reader should be able to know disabled state with just `disabled` attribute, the same applies to `readonly`.
- It could serve a different purpose from `disabled`. For example, https://a11y-101.com/development/aria-disabled which is up to users to decide.


wsuwt:
Closing this in favour of https://github.com/Refinitiv/refinitiv-ui/pull/101

PR #53, opened 2021-10-28
fix(tree-select): update condition for selectAll

Theeraphat-Sorasetsakul:
closes: https://jira.refinitiv.com/browse/ELF-1373

[emerald-multi-select] Incorrect behaviour appears when user searched and clicked on deselect all

Step for reproduce:
1. Click on “Select All” checkbox.
2. Type some unrelated search query and have “No matches” message.
3. Click on “Deselect All” checkbox
4. Remove search query.

Expected result:
1. Continues reproduce step, Checkbox “Select All” is displaying and checked.
2. In step 2 and 3, the component must show only  `no results found` and DONE / CANCEL buttons.


packages/elements/src/tree-select/index.ts
```diff
@@ -284,7 +284,7 @@ export class TreeSelect extends ComboBox<TreeSelectDataItem> {
    * @returns Is all selected
    */
   protected get allSelected (): boolean {
```
Sarin-Udompanish, line 286: Should we change the function name to isAllSelected ?

packages/elements/src/tree-select/index.ts
```diff
@@ -284,7 +284,7 @@ export class TreeSelect extends ComboBox<TreeSelectDataItem> {
...
-    return this.memo.selected === this.memo.selectable;
+    return this.memo.selectable > 0 && this.memo.selected === this.memo.selectable;
```
Sarin-Udompanish, line 287: Can we create a new boolean variable to store this condition because it has been use in many places ?

Sarin-Udompanish (approved):

PR #58, opened 2021-11-02
Bugfix/polyfill unit test

Sakchai-Refinitiv:
## Description
Fix unit-test in IE 11 does not work because missing import `@formatjs` polyfill

Fixes [ELF-1512](ELF-1512)

## Type of change

## Checklist


packages/elements/src/pagination/__test__/pagination.test.js
```diff
@@ -6,21 +6,29 @@ import {
...
+  aTimeout,
+  waitUntil
 } from '@refinitiv-ui/test-helpers';
 
+// Translations polyfills
+import '@formatjs/intl-locale/polyfill.iife';
+import '@formatjs/intl-getcanonicallocales/polyfill.iife';
+import '@formatjs/intl-pluralrules/polyfill.iife';
+import '@formatjs/intl-pluralrules/locale-data/en';
+import '@formatjs/intl-pluralrules/locale-data/zh';
```
goremikins, line 18: There are no tests for languages apart of English. You do not need to include all locales.

To make tests consistent across all locales, please provide `lang` for every case.

packages/elements/src/pagination/__test__/pagination.test.js
```diff
@@ -397,56 +388,35 @@ describe('pagination/Pagination', () => {
...
-        })
-        .then(() => {
-          expect(el.page).to.equal('7');
-          expect(eventNotFired).to.equal(true);
-        });
-    });
-    it('Should not fire page-changed event when page changes through attribute', () => {
-      let eventNotFired = false;
+    it('Should not fire page-changed event when page changes through attribute', async () => {
+      let eventFired = null;
```
goremikins, line 393: An event can be either fired or not: `let eventFired = false`

packages/elements/src/pagination/__test__/pagination.test.js
```diff
@@ -397,56 +388,35 @@ describe('pagination/Pagination', () => {
...
-        .then(() => {
-          expect(el.page).to.equal('2');
-          expect(eventNotFired).to.equal(true);
-        });
+      expect(el.page).to.equal('2');
+      expect(eventFired).to.equal(false, 'Event must not be fired');
     });
     it('Should not fire page-changed event when page is changes through property', async () => {
-      let eventNotFired = false;
+      let eventFired = null;
```
goremikins, line 407: As above, use boolean

packages/translate/__test__/elf-translate-element.test.js
```diff
@@ -1,9 +1,16 @@
 import { fixture, expect, isIE, elementUpdated, nextFrame } from '@refinitiv-ui/test-helpers';
```
goremikins, line 1: There is not need to change tests for `translate`. To start with, we do not test 3rd party software (format.js polyfills in this case), but just ensure that the translation package works. And then, if you want to include these, I can see that translations are checked for Russian locale, which is not included.

goremikins (changes requested):
As per comments, remove the code which is not required and add `lang="en-gb"` to make sure tests are consistent across different test machines.

packages/elements/src/pagination/__test__/pagination.test.js
Sakchai-Refinitiv: In this case, I think the `null` value might be better because it makes sure 100% of the `false` value will come from the `try-catch` only and it does not have two places to assign the `eventNotFired = false`

packages/elements/src/pagination/__test__/pagination.test.js
Sakchai-Refinitiv: Do you mean to provide the `lang="en-gb"` for every translation test cases only?

packages/translate/__test__/elf-translate-element.test.js
Sakchai-Refinitiv: Already reverted

Sakchai-Refinitiv: OK, I already reverted the change.

packages/elements/src/pagination/__test__/pagination.test.js
goremikins: It should import these (as we test English only):

```javascript
import '@formatjs/intl-getcanonicallocales/polyfill.iife';
import '@formatjs/intl-pluralrules/polyfill.iife';
import '@formatjs/intl-pluralrules/locale-data/en';
```

Every sample must enclude lang, e.g.: `<ef-pagination lang="en-gb"></ef-pagination>`

packages/elements/src/pagination/__test__/pagination.test.js
goremikins: There are three possible scenarios:
`try`, `catch` and test timeout. For `try` `eventFired = true`; for catch `eventFired = false` and if test times out, the generic error is thrown. If you care about double assignment, you can leave as undefined: `let eventFired;`

packages/elements/src/tree-select/__test__/tree-select.test.js
```diff
@@ -1,5 +1,14 @@
 import { expect, fixture } from '@refinitiv-ui/test-helpers';
 
+// Translations polyfills
+import '@formatjs/intl-locale/polyfill.iife';
+import '@formatjs/intl-getcanonicallocales/polyfill.iife';
+import '@formatjs/intl-pluralrules/polyfill.iife';
+import '@formatjs/intl-pluralrules/locale-data/en';
+import '@formatjs/intl-pluralrules/locale-data/zh';
```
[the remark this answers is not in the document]
Sakchai-Refinitiv: Fixed

packages/elements/src/pagination/__test__/pagination.test.js
Sakchai-Refinitiv: Fixed, I prefer to leave as undefined.

packages/elements/src/pagination/__test__/pagination.test.js
Sakchai-Refinitiv: Fixed

packages/elements/src/tree-select/__test__/tree-select.test.js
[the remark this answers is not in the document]
Sakchai-Refinitiv: Fixed

packages/elements/src/pagination/__test__/pagination.test.js
Sakchai-Refinitiv: Updated

goremikins (approved):

PR #73, opened 2021-11-09
feat(calendar): add accessibility

goremikins:
## Description
Improves screen reader support

## Type of change

## Checklist


packages/elements/src/calendar/index.ts
```diff
@@ -378,16 +455,44 @@ export class Calendar extends ControlElement implements MultiValue {
...
+   * @param changedProperties Properties which have changed
+   * @return {void}
+   */
+  protected updated (changedProperties: PropertyValues): void {
+    super.updated(changedProperties);
+
+    // This code is here to ensure that focus is not lost
+    // while navigating through the render views using keyboard
+    if (this.focused && changedProperties.has('renderView') && this.activeElement !== this.viewBtnRef.value) {
+      this.viewBtnRef.value!.focus();
```
wsuwt, line 483: I know view button is always there, but asserting null throws a warning. Should we use `?` instead to only focus when button is really there ?

Jidapa-Pai:
I'm already checked issue about tap on year after merge this code https://github.com/Refinitiv/refinitiv-ui/pull/81 from develop, it works now.

packages/elements/src/calendar/index.ts
goremikins: Done

wsuwt (approved):

goremikins:
@TremayneChrist , @wsuwt , I think this need another walkthrough. Most notably I have changed the following:
1.  IE11, navigation did not work
2. IE11, it took absolutely ages to load the calendar. The reason is because polyfills are required to render formatted dates. So I removed `aria-label` for all elements that cause problem in IE11
3. Simplified some types
4. Changed the way how navigation maps are calculated to keep the code cleaner

PR #80, opened 2021-11-17
feat(select): improve screen reader support

wsuwt:
## Description

Improves screen reader support for `select`. 

https://github.com/Refinitiv/refinitiv-ui/pull/74 must be merge before merging this pull request. 

## Type of change

## Checklist


Jidapa-Pai:
I tested with the following tools on Windows and get result below.
- **Jaws** - The key control works fine.
- **NVDA** - Cannot use spacebar/enter to open popup list for select that using data object.
- **Windows's Narrator** - It's not work at all when tap spacebar/enter to open popup list.

Note: Test on Chrome.

packages/elements/src/select/index.ts
```diff
@@ -192,11 +185,24 @@ export class Select extends ControlElement implements MultiValue {
...
   * Set state to error
+  * @param value error value
   */
   @property({ type: Boolean, reflect: true })
-  public error = false;
+  public set error (value: boolean) {
+    const oldValue = this._error;
+    if (oldValue !== value) {
+      this._error = value;
+      this.ariaInvalid = String(value);
```
Jidapa-Pai, line 199: It would set `aria-invalid="null"` when I set `error` as `null`, The some screen reader e.g. NVDA also read as invalid.

packages/elements/src/select/index.ts
```diff
@@ -295,6 +301,28 @@ export class Select extends ControlElement implements MultiValue {
...
+   */
+  @property({ type: String, reflect: true, attribute: 'aria-haspopup' })
+  public ariaHasPopup = 'listbox';
+
+  /**
+   * Aria indicating open state of listbox popup
+   * @ignore
+   */
+  @property({ type: String, reflect: true, attribute: 'aria-expanded' })
+  public ariaExpanded = String(this.opened);
```
Jidapa-Pai, line 316: `aria-expanded` didn't change when set `opened` property manually.

wsuwt: Fixed

packages/elements/src/select/index.ts
```diff
@@ -665,31 +700,48 @@ export class Select extends ControlElement implements MultiValue {
...
 
   /**
-   * Focus and highlight the next/previous element
-   * @param direction Focus next or previous element
+   * Focus and highlight element according to specified direction
+   * @param direction previous, next, first or last focusable element
    * @returns {void}
    */
-  private focusElement (direction: number): void {
+  private focusElement (direction: 'previous' | 'next' | 'first' | 'last'): void {
```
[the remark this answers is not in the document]
wsuwt: Agreed. Fixed.

Jidapa-Pai:
> I tested with the following tools on Windows and get result below.
> 
> * **Jaws** - The key control works fine.
> * **NVDA** - Cannot press spacebar/enter to open popup list for select that using data object.
> * **Windows's Narrator** - It's not work at all when tap spacebar/enter to open popup list.
> 
> Note: Test on Chrome.

Actually, the problem occur when we put `aria-label` or `aria-labelledby` to the element, it's  intercept `tapstart` event.

packages/elements/src/select/index.ts
```diff
@@ -295,6 +321,28 @@ export class Select extends ControlElement implements MultiValue {
     return this.selectedSlotItems.map(item => this.getItemValue(item));
   }
 
+  /**
+   * Aria indicating that select has a popup of type listbox
+   * @ignore
+   */
+  @property({ type: String, reflect: true, attribute: 'aria-haspopup' })
+  public ariaHasPopup = 'listbox';
```
goremikins, line 329: Overriding AOM model does not work in Firefox (the properties are not reflected). It has to work directly with attributes (via setAttribute/removeAttribute). Probably we need a new util for that.

packages/elements/src/select/index.ts
```diff
@@ -53,6 +45,13 @@ const LABEL_SEPARATOR = ', '; // TODO: for multiselect
 const POPUP_POSITION = ['bottom-start', 'top-start'];
 const KEY_SEARCH_DEBOUNCER = 300;
 
+enum Direction {
```
goremikins, line 48: `Direction` essentially means moving towards something. `First`, `Last`, `Next` and `Previous` is not a direction, but destination :) so I may suggest to rename it to `enum Navigation`

wsuwt: Agree, navigation sounds better.

packages/elements/src/select/index.ts
wsuwt: On my testing, Firefox 94.0.2 does reflect these attributes correctly just like any other lit properties. Can you recheck again ?

wsuwt:
> > I tested with the following tools on Windows and get result below.
> > 
> > * **Jaws** - The key control works fine.
> > * **NVDA** - Cannot press spacebar/enter to open popup list for select that using data object.
> > * **Windows's Narrator** - It's not work at all when tap spacebar/enter to open popup list.
> > 
> > Note: Test on Chrome.
> 
> Actually, the problem occur when we put `aria-label` or `aria-labelledby` to the element, it's intercept `tapstart` event.

You need to press `Ctrl` + `Enter` / `Spacebar` instead when the screen reader is on.

packages/elements/src/select/__demo__/index.html
```diff
@@ -16,30 +16,29 @@
...
-    <ef-item value="8">Chocolate</ef-item>
-    <ef-item value="9">Honey &amp; Walnut</ef-item>
-    <ef-item value="10">Pistachio</ef-item>
-    <ef-item value="11">Salted Caramel</ef-item>
-    <ef-item value="12">Mint Choc Chip</ef-item>
-    <ef-item value="13">Hazelnut</ef-item>
-    <ef-item value="14">Strawberry</ef-item>
-    <ef-item value="15">Raspberry</ef-item>
+    <ef-item role="presentation" type="header">Drinks</ef-item>
+    <ef-item role="option" value="1">Cola</ef-item>
```
goremikins, line 20: A reminder: these option roles need to go into accessibility documentation

goremikins (approved):
Looks good to me. Please see my comment to not forget to update accessibility docs.

PR #90, opened 2021-11-24
fix(phrasebook): change ICU syntax on japanese language

bualoy-napat:
## Description
The change is related to the syntax of the ICU message that requires `other` argument.

Fixes # (issue)
ELF-1680  Japanese translation (ja) in tree-select does not work

## Type of change

## Checklist

### DIFF
--- a/packages/phrasebook/src/locale/ja/tree-select.ts
+++ b/packages/phrasebook/src/locale/ja/tree-select.ts
@@ -9,8 +9,8 @@ const translations = {
   // used as a toggle control
   FULL_LIST: 'すべて',
   SELECTED: '選択項目のみ',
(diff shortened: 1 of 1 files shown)


dtanp-rft:
Reference : https://github.com/formatjs/formatjs/commit/976c79a5df41130d5d032ada846fd7496776ce8a

This is a breaking change from formatjs. The new version requires `other` if it's conditional ICU format as they want to make sure that it always has a fallback option to translation.

For example, if it is `'{expansion, select, true {Collapse All} false {Expand All}}'` then it will return `undefined` if the value is neither `true` or `false`.

dtanp-rft (approved):

PR #99, opened 2021-11-30
Fix/add missing default value from api doc of all elements

Sarin-Udompanish:
## Description
Some properties of elements are missing default value and type.

Fixes # (issue)
1. Add missing default value
2. Add missing type
3. Add (readonly) description
4. Fix the case that default value is "" but shown 0 

## Checklist


packages/elements/src/overlay/elements/overlay.ts
```diff
@@ -621,6 +626,7 @@ export class Overlay extends ResponsiveElement {
   /**
    * Set focus boundary to restrict tabbing. Default's overlay itself.
    * If external focus is required, set to null
+   * @default overlay
```
dtanp-rft, line 629: Tricky. Will it look weird if we do `@default this`?

packages/elements/src/sparkline/index.ts
```diff
@@ -30,12 +30,14 @@ export class Sparkline extends ResponsiveElement {
...
   /**
    * Chart data as an array of number.
+   * @type {number[]}
    */
   @property({ type: Array })
   public data: number[] = [];
 
   /**
    * Chart previous data as an array of number.
+   * @type {array}
```
dtanp-rft, line 40: This should be array of number

packages/elements/src/overlay/elements/overlay.ts
Sarin-Udompanish: At first it show "this" as default value but I changed it to element name. 
I'm not sure as well which is better between 'this' or 'element-name'.

dtanp-rft: For correctness, I think `this` is more correct because the value is itself, not any `ef-overlay`.
Make sense?

packages/elements/src/sparkline/index.ts
Sarin-Udompanish: fixed

packages/elements/src/overlay/elements/overlay.ts
Sarin-Udompanish: Agree. Fixed

packages/elements/src/chart/index.ts
```diff
@@ -148,6 +148,7 @@ export class Chart extends BasicElement {
   /**
    * List of available chart colors
    * @type {string[]}
+   * @default []
    * @returns {string[]} List of available chart colors
```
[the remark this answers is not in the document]
Sarin-Udompanish: fixed

dtanp-rft (approved):

PR #102, opened 2021-11-30
feat(tree): improve screen reader support

wsuwt:
## Description

- `role=tree` on tree element. 
- `role=treeitem` on tree-item element. 
- Adds `aria-expanded` to indicate expand/collapsed state of each `tree-item`
- Adds `aria-selected` for single mode. 
- Adds `aria-checked` for multiple mode, to support tri-state i.e, `true`, `false` and `mixed`.
- Adds `aria-level` which is `depth+1` to each tree-item to indicate item depth.

## Type of change

## Checklist


wsuwt:
@TremayneChrist and @goremikins  Can you skim over this PR, it's not a lot of code but I'm not sure on some of the limitations on tree

- Ability to toggle `attribute` on a `treeitem` between group and default mode, it would re-render to previous  state on hover so I assume this is not supported and I did not add condition to remove `aria-expanded` when a parent changes to child node.
- Changing multiple to single mode would leave selected or half-selected parent node with selected state on parent node which is not correct so again I did not replace `aria-checked` with `aria-selected` to reflect this behaviour.

packages/elements/src/tree/elements/tree-item.ts
```diff
@@ -156,24 +157,44 @@ export class TreeItem<T extends TreeDataItem = TreeDataItem> extends ControlElem
     return this.checkedState === CheckedState.INDETERMINATE;
   }
 
+  /**
+   * Invoked whenever the element is updated
+   * @param {PropertyValues} changedProperties Map of changed properties with old values
```
TremayneChrist, line 162: Dont need the type here

packages/elements/src/tree/elements/tree-item.ts
```diff
@@ -156,24 +157,44 @@ export class TreeItem<T extends TreeDataItem = TreeDataItem> extends ControlElem
...
+   * @returns {void}
+   */
   protected update (changedProperties: PropertyValues): void {
     super.update(changedProperties);
+
     if (changedProperties.has('checkedState')) {
+      // Parent node in single-mode cannot be selected
+      if (this.parent && !this.multiple) {
+        return;
+      }
```
TremayneChrist, line 172: Elements can be reused, so I think it should still process and remove any attributes.

packages/elements/src/tree/elements/tree-item.ts
wsuwt: Fixed

packages/elements/src/tree/elements/tree-item.ts
wsuwt: Fixed.

packages/elements/src/tree/elements/tree-item.ts
```diff
@@ -156,24 +157,69 @@ export class TreeItem<T extends TreeDataItem = TreeDataItem> extends ControlElem
...
+          this.setAttribute('aria-checked', 'false');
+        }
+
+        // In single mode, only children nodes are selectable
+        if (!this.parent && !this.multiple) {
+          this.setAttribute('aria-selected', 'false');
+        }
+        break;
+    }
+  }
```
TremayneChrist, line 197: @wsuwt I feel this should always be linked with the selected state, so if `selected` attribute is being removed, the `aria-selected` should be set to `false`. The logic for setting the correct checked state can be handled in the renderer.

```suggestion
  private checkedChanged (): void {
    switch (this.checkedState) {
      case CheckedState.CHECKED:
        this.setAttribute('selected', '');
        this.setAttribute(this.multiple ? 'aria-checked' : 'aria-selected', 'true');
        break;
      case CheckedState.INDETERMINATE:
        this.setAttribute('selected', 'indeterminate');
        this.setAttribute('aria-checked', 'mixed');
        break;
      default:
        this.removeAttribute('selected');
        this.setAttribute(this.multiple ? 'aria-checked' : 'aria-selected', 'false');
        break;
    }
  }
```

packages/elements/src/tree/elements/tree-item.ts
```diff
@@ -156,24 +157,69 @@ export class TreeItem<T extends TreeDataItem = TreeDataItem> extends ControlElem
...
+   * aria-checked is used for multiple mode due to tri-state support
+   * @returns {void}
+   **/
+  private multipleChanged (): void {
+    this.removeAttribute(this.multiple ? 'aria-selected' : 'aria-checked');
+    this.checkedChanged();
+  }
+
+  /**
+   * Handles selectedd and arias attribute changes
```
TremayneChrist, line 171:
```suggestion
   * Handles selected and aria attribute changes
```

wsuwt: Fixed

packages/elements/src/tree/elements/tree-item.ts
wsuwt: Fixed

packages/elements/src/tree/helpers/renderer.ts
```diff
@@ -36,6 +37,19 @@ export class TreeRenderer extends Renderer {
...
+      el.setAttribute('aria-setsize', String(composer.getItemSiblings(item).length + 1));
+
+      if (el.parent) {
+        el.setAttribute('aria-expanded', el.expanded ? 'true' : 'false');
+      }
+
+      // In single mode, only children nodes are selectable
+      if (el.parent && !el.multiple) {
+        el.removeAttribute('aria-selected');
+      }
```
TremayneChrist, line 51: The element should set the aria tags, not the renderer. The renderer should just set the correct state of the tree item.

TremayneChrist: It's a bit confusing that some are written the renderer and some are in the item.

TremayneChrist: Not sure what to do about `aria-setsize` however....

packages/elements/src/tree/elements/tree-item.ts
```diff
@@ -156,24 +157,61 @@ export class TreeItem<T extends TreeDataItem = TreeDataItem> extends ControlElem
...
+        this.setAttribute('selected', '');
+        this.setAttribute(this.multiple ? 'aria-checked' : 'aria-selected', 'true');
+        break;
+      case CheckedState.INDETERMINATE:
+        this.setAttribute('selected', 'indeterminate');
+        this.setAttribute('aria-checked', 'mixed');
+        break;
+      default:
+        this.removeAttribute('selected');
+        this.setAttribute(this.multiple ? 'aria-checked' : 'aria-selected', 'false');
```
TremayneChrist, line 186: I think you could move the logic from the render to here.
```suggestion
        if (this.parent && !this.multiple) {
          this.removeAttribute('aria-selected');
        }
        else {
          this.setAttribute(this.multiple ? 'aria-checked' : 'aria-selected', 'false');
        }
```

packages/elements/src/tree/helpers/renderer.ts
wsuwt: I've looked into moving `aria-setsize` logics into `tree-item` and no luck so far. `tree-item` has no access to composer and is therefore unaware of adjacent nodes. This attribute is crucial for screen reader counting order. What do you think if go with this approach for now ?

TremayneChrist: We'd probably need to add a property to the list-tem for setsize and just set the property in the renderer. It's a bit tricky because all of our foundational elements now handle ARIA automatically, so the tree-item should follow, as it's part or the suite.

This gets even more interesting when people want to create their own fully custom renderer, because they will likely break the accessibility and have to implement the ARIA attributes themselves.

wsuwt: Good news. I just re-tested again on Windows, item order and set size seems to be automatically computed in both Narrator and NVDA. As long as items are in the DOM, it is fine. And these never worked with VoiceOver on MacOS anyways. I'll remove setsize and rely on browsers to compute these properties instead.

TremayneChrist (approved):

PR #146, opened 2022-01-07
Include ef documentation into monorepo

dtanp-rft:
Include documentation into monorepo.
* `documents` is added to be part of npm workspace
* Run `npm install` should do npm install in both `packages` and `documents`
* Run `npm run build` should build `documents` folder as well i.e. `dist` and `build` folder to be created
* `@refinitiv-ui/elements` which is a dependency of `documents` should be hoisted and use internal package instead of install it from npm
* Run`npm run build:docs` and `npm run start:docs` should build and start docs respectively
* Run `npm run build:docs` should trigger `api-analyzer` script so `custom-elements.md` should be generated in elements folder

TODO: Remove unused branches after this PR is merged.

- [x] `docs`
- [x] `new-docs`
- [x] Set readonly to elf-docs `v5` branch on gitlab


documents/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "ef-docs",
+  "version": "5.1.0",
```
wsuwt, line 3: Do we want to include `@refinitiv-ui/*` scope into this package as well ? Something like `@refinitiv-ui/docs` would match our naming convention.

wsuwt: Also, this is the first official release, should we version it as `5.0.0` ?

dtanp-rft: It shouldn't matter because it's not npm module but your proposal name could add more beauty to the project :)

wsuwt: Yep, just for the aesthetics. I'll update it.

wsuwt (approved):

PR #156, opened 2022-01-11
feat: upgrade `lit` to 2.1.1

wsuwt:
## Description

There seems to be an issue between `lit@2.1.0` and Skypack, see https://codepen.io/wsuwt/pen/bGoxPVw?editors=1011. This is to make sure that we use a patched lit version `2.1.1`.

Additionally, I think we could benefit from a new directive - `choose`, see https://lit.dev/docs/templates/directives/#choose. 

## Type of change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,40 +52,40 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.16.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.0.tgz",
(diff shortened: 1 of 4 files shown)


dtanp-rft (approved):
Feel a bit odd that there are same `lit` dependency in 3 different projects but I'm fine to release this as it is.

TremayneChrist:
@wsuwt do you know why lit is defined as a direct dependency in polyfills and translate libraries?

wsuwt:
> @wsuwt do you know why lit is defined as a direct dependency in polyfills and translate libraries?

In `translate` we are using a directive and in `polyfills` we are using one of the ployfills from `lit` required for legacy browsers

TremayneChrist (approved):

PR #160, opened 2022-01-12
Chore/remove eslint config from sample app

Sarin-Udompanish:
## Description
Remove eslint config of react & vue sample apps to prevent lint error when commit code and remove Bams registry from package.lock files.


TremayneChrist:
What about angular? We still have this in eslintignore https://github.com/Refinitiv/refinitiv-ui/blob/9a0ad709d71c243fe68dbc38318a04745ea5b23b/.eslintignore#L10

Sarin-Udompanish:
> 

I will remove it from eslintignore.

wsuwt (approved):

PR #161, opened 2022-01-12
feat(collapse): support screen reader

Sakchai-Refinitiv:
## Description
Support screen reader for collapse element

## Type of change

## Checklist


packages/elements/src/collapse/index.ts
```diff
@@ -195,13 +195,22 @@ export class Collapse extends BasicElement {
...
+        aria-expanded=${this.expanded}
+        aria-controls="content"
+      >
+        <ef-icon icon="right" slot="left" part="toggle" aria-hidden="true"></ef-icon>
         <slot slot="left" name="header-left"></slot>
         <slot slot="right" name="header-right"></slot>
         ${this.header}
       </ef-header>
-      <div part="content">
+      <div id="content" part="content" role="region" aria-labelledby="header" ?hidden=${!this.expanded}>
```
Jidapa-Pai, line 213: NVDA did not announce collapse/expanded state once it was expanded, It seem to have a problem `hidden`. Found on Chrome.

Sakchai-Refinitiv: @Jidapa-Pai, this is the one I was trying to solve; it works in Firefox but doesn't work in Chromium (e.g. Chrome, MS Edge).  I already try to use CSS `display: none` to hide the content but got the same result. Do you have any ideas?

Jidapa-Pai: Is the `hidden` using for skip the content region when collapse? If so, can we try `tabindex`?

packages/elements/src/header/index.ts
```diff
@@ -30,6 +31,8 @@ export class Header extends BasicElement {
     return VERSION;
   }
 
+  protected readonly defaultRole = 'heading';
```
Jidapa-Pai, line 34: We should define defaultRole as `protected readonly defaultRole: string | null = 'heading';` so it can be override.

packages/elements/src/header/index.ts
```diff
@@ -56,6 +59,13 @@ export class Header extends BasicElement {
   @property({ type: String, reflect: true })
   public level: '1'| '2'| '3' = '2';
 
+  protected updated (changedProperties: PropertyValues): void {
+    super.updated(changedProperties);
+    if (changedProperties.has('level')) {
+      this.setAttribute('aria-level', this.level);
```
Jidapa-Pai, line 65: I would suggest to define `ariaLevel` as reflect property and change `level` to setter/getter. We usually use this pattern in the others element e.g. `checkbox`.

packages/elements/src/header/index.ts
Sakchai-Refinitiv: We decided to developer handle attribute `role` and `aria-level` at the application level so the element header does not need to change.

packages/elements/src/header/index.ts
Sakchai-Refinitiv: We decided to developer handle attribute `role` and `aria-level` at the application level so the element header does not need to change.

bualoy-napat (approved):

PR #174, opened 2022-01-18
fix(slider): value-changed event is not fired

Jidapa-Pai:
## Description
Fixes # (issue)
Step to reproduce:

1. Programmatically set `value` by Javascript. e.g. set `value` from 0 to 10.
2. Drag the slider back to 0.
Result: The `value-changed` event is not fired. Try to reproduce here https://codepen.io/jidapa-pai/pen/XWeQdeO

The problem is when user programmatically set `value` the `valuePrevious` which is internal value didn't update so when drag the slider back to previous value, it not detect value change and not fires the `value-changed` event.
So, the solution is to update `valuePrevious` when the `value` was programmatically set.

## Type of change

## Checklist


packages/elements/src/slider/__test__/slider.test.js
```diff
@@ -1358,6 +1358,119 @@ describe('slider/Slider', () => {
...
+        // Darg end
+        setTimeout(() => window.dispatchEvent(new MouseEvent('mouseup', { clientX: dragValuePositionStart, clientY: 0 })));
+        await oneEvent(window, 'mouseup');
+        expect(el.dragging).to.be.false;
+        expect(el.value).to.equal(calculateValue(el, dragValuePositionStart).toString());
+
+        // Check call fire event
+        expect(callCountValue).to.equal(1);
+      });
+      it('Event from-changed should fires when from property was set via api and drag the slider back to previous value', async () => {
```
Sakchai-Refinitiv, line 1397: Making the line space between the test cases improves code readability.

packages/elements/src/slider/__test__/slider.test.js
```diff
@@ -1358,6 +1358,119 @@ describe('slider/Slider', () => {
         expect(toEventFiredCount, 'Event "to-changed" should not fired').to.equal(0);
         expect(slider.to).to.equal('90');
       });
+
+      // jira ELF-1729
```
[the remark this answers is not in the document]
Theeraphat-Sorasetsakul: I think no. I will remove it.

packages/elements/src/slider/__test__/slider.test.js
```diff
@@ -1358,6 +1358,119 @@ describe('slider/Slider', () => {
...
+        // Darg end
+        setTimeout(() => window.dispatchEvent(new MouseEvent('mouseup', { clientX: dragPositionStart, clientY: 0 })));
+        await oneEvent(window, 'mouseup');
+        expect(el.dragging).to.be.false;
+        expect(el.from).to.equal(calculateValue(el, dragPositionStart).toString());
+
+        // Check call fire event
+        expect(callCountValue).to.equal(1);
+      });
+      it('Event to-changed should fires when to property was set via api and drag the slider back to previous value', async () => {
```
Sakchai-Refinitiv, line 1435: Please add line space

packages/elements/src/slider/index.ts
```diff
@@ -917,6 +917,11 @@ export class Slider extends ControlElement {
         }
       }
     }
+
+    if (!this.dragging) {
+      // Update internal `valuePrevious` when `value` was programatically set by user.
+      this.valuePrevious = this.value;
+    }
```
Sakchai-Refinitiv, line 924: Why does this change add to the end of the function instead of the beginning?

Sakchai-Refinitiv (approved):

Sakchai-Refinitiv:
@TremayneChrist @goremikins Do you want to skim this PR before merging?

goremikins:
I do not understand what is going on there and why as the code looks very messy :) but my guess is that the `value-changed` event should be fired not on `blur`, but when the drag is released.

Sakchai-Refinitiv:
> I do not understand what is going on there and why as the code looks very messy :) but my guess is that the `value-changed` event should be fired not on `blur`, but when the drag is released.

If you think the code should be refactored, you can create a new story we can do this in next sprint.

Theeraphat-Sorasetsakul:
> I do not understand what is going on there and why as the code looks very messy :) but my guess is that the `value-changed` event should be fired not on `blur`, but when the drag is released.

I agreed the code like messy to read.
`value-changed` event is fired when drag is released. `onBlur` is for number-field.

PR #179, opened 2022-01-26
Revert "feat(collapse): support screen reader"

TremayneChrist:
Reverts Refinitiv/refinitiv-ui#161

@Sakchai-Refinitiv this should be reviewed properly by members of the team with deeper accessibility knowledge.

PR #189, opened 2022-01-28
feat(collapse): support screen reader

Sakchai-Refinitiv:
## Description
Support screen reader for collapse element.
- add `aria-level` attribute and delegate `value` to the header element for semantic meaning.
- add the `header-toggle` part which has a `role button` and wraps header text to make semantic meaning for a screen reader accessible
- hide the `arrow icon` from a screen reader because of using the button in the `header-toggle` part instead.
- add `id` and `aria attributes` to the part content to support accessibility when expanded/collapsed states changed.
- add `hidden` attribute to the `slot` to prevent users access the content when collapsed by using keyboard navigation.
- update theme by disabling outline from the `button` in `header-toggle`
## Type of change

## Checklist

Ref: https://github.com/Refinitiv/refinitiv-ui/pull/179


packages/elements/src/collapse/index.ts
```diff
@@ -101,6 +101,13 @@ export class Collapse extends BasicElement {
   @query('ef-panel', true)
   private panel!: HTMLElement;
 
+  /**
+   * Hierarchical level of element
+   * @ignore
+   */
+  @property({ type: String, reflect: true, attribute: 'aria-level' })
```
goremikins, line 108: Why do you need it? It break Firefox AOM model. Listen for attribute changes and set a private property.

packages/elements/src/collapse/index.ts
```diff
@@ -195,15 +202,17 @@ export class Collapse extends BasicElement {
    */
   protected render (): TemplateResult {
     return html`
-      <ef-header part="header" level="${this.level}" @tap="${this.handleTap}">
-        <ef-icon icon="right" slot="left" part="toggle"></ef-icon>
+      <ef-header part="header" level="${this.level}" @tap=${this.handleTap} role="heading" aria-level=${this.ariaLevel}>
```
goremikins, line 205: I think the entire `ef-header` is a clickable element, so it needs `role="button"`. As well, there is no need to set `aria-level` as the semantics is only valid for the light DOM.

This probably needs testing, but I think the entire `ef-collapse` element should have `role=heading`.

packages/elements/src/collapse/index.ts
```diff
@@ -195,15 +202,17 @@ export class Collapse extends BasicElement {
...
+        <div part="header-toggle" id="header-toggle" role="button" tabindex="0" aria-expanded="${this.expanded}" aria-controls="content">
+          ${this.header}
+        </div>
+        <ef-icon icon="right" slot="left" part="toggle" aria-hidden="true"></ef-icon>
```
goremikins, line 209: Why `aria-hidden`? Icon is not focusable by default so should not be visible by the screen reader.

packages/elements/src/collapse/index.ts
```diff
@@ -195,15 +202,17 @@ export class Collapse extends BasicElement {
...
         <slot slot="left" name="header-left"></slot>
         <slot slot="right" name="header-right"></slot>
-        ${this.header}
       </ef-header>
-      <div part="content">
+      <div id="content" part="content" role="region" aria-labelledby="header-toggle">
         <ef-panel ?spacing="${this.spacing}" transparent>
-          <slot></slot>
+          <slot ?hidden=${!this.expanded}></slot>
```
goremikins, line 215: I do not like setting `hidden` on slot itself, it might not work correctly in IE11, and slot is kind of a special element. I can see two solutions:
1) Make slot optional, e.g. `${this.expanded ? html`<slot></slot>` : undefined}`
2) Set hidden on the panel

packages/elements/src/collapse/index.ts
```diff
@@ -141,7 +148,7 @@ export class Collapse extends BasicElement {
    * @returns {boolean} true if target is ef-header
    */
   private static isHeader (element: HTMLElement): boolean {
```
goremikins, line 150: I guess this function is here to prevent hiding when the click happens on the slotted element. If so, why not to create a local reference to the `ef-header` itself?

goremikins: As well, `handleTap` should not be an arrow function.

goremikins (changes requested):
As mentioned in the comments, accessibility needs to be simplified, but please check my suggestions with the screen reader. Further, the code needs to work consistently in Firefox and Chrome.
Furthermore the inconsistencies with the code and formatting need to be resolved while doing refactoring.

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: Your solutions are not valid, I have already tested,

1. Entire `ef-collapse` cannot be a `role=heading` because it makes semantics incorrect, the header and the content must have a separate container place sibling with each other.
2. The `ef-header` cannot be a `role=button` because developers can put a button, link, or icon in the slot of `ef-collapse` in left or right slots which makes semantic meaning and screen reader announce incorrectly

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: For disabled people, it is not necessary to focus and navigate to the icon because you can use the entire header as a button to toggle, and the screen reader announces `expanded/collapsed` every time when focusing on the header that is the same meaning as the icons.

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: Yes, you understand `isHeader` correctly but I am not getting about `create a local reference` Can you provide more detail?

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: That is correct, it does not work on IE11.  I will go with solution 1 (Make slot optional) because solution 2 I was tested does not work on Chromium browsers.

Sakchai-Refinitiv: In solution one it might affect performance because it swaps content in/out DOM tree.

packages/elements/src/collapse/index.ts
goremikins: How can you focus on `ef-icon`?

packages/elements/src/collapse/index.ts
goremikins:
```
html`<ef-header ${ref(this.headerRef)}`

if (target === this.headerRef.value! || target === other-clickable-targets) { ... }

```

packages/elements/src/collapse/index.ts
goremikins: You made valid points, however as we have discussed these needs further research.

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: Actually, it is not a focus, but it is something like an `current element` for a screen reader (you can see a border around of active element).  if you use the `VoiceOver` on macOS you can use the keyboard command `Capslock + ShiftUp + Down Arrow` to navigate to inside the header content.

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: Thanks

packages/elements/src/collapse/index.ts
goremikins: Even if it does affect performance (which I doubt, as the slot acts as a placeholder and the actual content is already on the page), the effect can be disregarded. Anything that can be done within a single frame (~24ms) is not noticeable by the user eye.

packages/elements/src/collapse/index.ts
goremikins: Yes, the user can, but why is it a problem? If you can see the icon, the impaired user might want to see it as well via keyboard commands. You can set `aria-label` for the icon in that case

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: I need this to allow developers can set a heading level to the header inside the collapse element.  Do you mean should not override all aria properties? is it conflict with AOM?

goremikins: Yes, you cannot override aria properties

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: The problem is if no `aria-label` for the icon the screen reader will announce the heading with an icon that is not good because it is very long and make a user confused.
For example, it announces `"Heading level 2, 2 items, icon right image, Title xxx, expanded button"` instead of `"Heading level 2, Title xxx, expanded button"`

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: I agree with you.

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: I have already removed it.

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: I use a new solution, just add CSS `visibility hidden` to the content when collapsed. it works well.

packages/elements/src/collapse/index.ts
Sakchai-Refinitiv: We do not want this function anymore, we let developers handle events from content in slots by themselves.

Sakchai-Refinitiv:
Hi @goremikins, I have already updated the code, could you please continue to review it.  Thanks

Sakchai-Refinitiv:
Please use this PR instead https://github.com/Refinitiv/refinitiv-ui/pull/203

PR #191, opened 2022-02-01
feat(list): improve keyboard navigation

wsuwt:
## Description
A list should only have one focus point, meaning that items in the list shoud not be focusable by Tab key. For instance, a person with vision disability encounter a list with 1,000 items on a page should not have to Tab a thousand time just to skip to another object on the page. To do this, we can focus on the list itself and then utilise `aria-activedescendant` attribute with value of current item id to indicate current focused item. 

However, this only partially work in MacOS VoiceOver. VoiceOver annouces `role=tree` as `table` this is a known issue.

## Expected result

- A list should contain only one tab focus
- Items in list can be navigate with `Up` and `Down` arrow keys.
- List should have a style to indicates that it is in focus when it is focused by keyboard.
- An item should have `aria-checked` instead of `aria-selected` in multiple mode.

## Type of change

## Checklist


packages/elements/src/tree/helpers/renderer.ts
```diff
@@ -11,23 +12,23 @@ type RendererScope = {
...
     super((item: TreeDataItem, composer: CollectionComposer<TreeDataItem>, element = document.createElement('ef-tree-item')): HTMLElement => {
       const el = element as TreeItem;
       const multiple = !!scope && (scope as RendererScope).multiple === true;
       const noRelation = !!scope && (scope as RendererScope).noRelation === true;
       const mode = !multiple || !noRelation ? TreeManagerMode.RELATIONAL : TreeManagerMode.INDEPENDENT;
-      if (currentComposer !== composer || currentMode !== mode) {
-        currentMode = mode;
-        currentComposer = composer;
-        manager = new TreeManager(composer, mode);
-      }
```
wsuwt, line 28: Hi @TremayneChrist , I wonder what `currentMode` and `currentMode` are for and is it safe to remove. I want to add key property to `Renderer` but `super()` must be the first statement in constructor.

.commitlintrc.yml
```diff
@@ -1,2 +1,4 @@
 extends:
   - '@commitlint/config-conventional'
+rules:
+  header-max-length: [1, 'always', 150]
```
goremikins, line 4: What is this for?

wsuwt: It's commit message maximum length. I had one lengthy commit which I could not commit and it think 100 characters is too few considering that we also use our commit message for CHANGELOG.

packages/elements/src/tree/helpers/renderer.ts
TremayneChrist: This mode is for setting relational updates on/off. So if you toggle a parent it will decide whether to check/uncheck dependents.

TremayneChrist: This also probably affects the selected state of items in the renderer, which is why it is required.

packages/elements/src/item/index.ts
```diff
@@ -186,6 +182,12 @@ export class Item extends ControlElement {
     if (changedProperties.has('type')) {
       this.typeChanged();
     }
+    if (changedProperties.has('multiple')) {
+      this.multipleChanged();
+    }
+    if (changedProperties.has('selected')) {
+      this.selectedChanged();
+    }
     super.update(changedProperties);
```
goremikins, line 191: In lit.dev lifecycle there is new hook, called `willUpdate`. In majority of cases we should use that instead of `update`

packages/elements/src/list/helpers/list-renderer.ts
```diff
@@ -32,6 +39,7 @@ export class ListRenderer extends Renderer {
       el.label = composer.getItemPropertyValue(item, 'label') as string;
       el.subLabel = composer.getItemPropertyValue(item, 'subLabel') as string;
       el.value = composer.getItemPropertyValue(item, 'value') as string;
+      el.id = `${this.key}-${el.value}`;
```
goremikins, line 42: Do we need `key` here? Render is always rendered in shadowDom, is not it?

packages/elements/src/list/helpers/list-renderer.ts
```diff
@@ -42,6 +50,7 @@ export class ListRenderer extends Renderer {
 
       const itemRole = el.type === 'text' || !el.type ? 'option' : 'presentation';
       el.setAttribute('role', itemRole);
+      el.setAttribute('tabindex', '-1');
```
goremikins, line 53: `el.tabIndex = -1;` should do the same

packages/elements/src/list/helpers/list-renderer.ts
wsuwt: No, the items are in light DOM

packages/elements/src/list/helpers/list-renderer.ts
wsuwt: fixed

packages/elements/src/item/index.ts
wsuwt: fixed

packages/elements/src/tree/helpers/renderer.ts
wsuwt: But both `currentMode` and `currentComposer` are not used at all in the renderer, I've done some basic testing and see no difference.

packages/elements/src/item/index.ts
```diff
@@ -179,14 +157,51 @@ export class Item extends ControlElement {
...
+   */
+  protected willUpdate (changedProperties: PropertyValues): void {
     if (changedProperties.has('type')) {
       this.typeChanged();
     }
-    super.update(changedProperties);
+    if (changedProperties.has('multiple')) {
+      this.multipleChanged();
+    }
+    if (changedProperties.has('selected')) {
```
TremayneChrist: `multipleChanged` calls `selectedChanged` so use `else if` so as not to execute twice.
```suggestion
    else if (changedProperties.has('selected')) {
```

packages/elements/src/list/index.ts
```diff
@@ -253,7 +242,7 @@ export class List<T extends DataItem = ItemData> extends ControlElement {
    * @returns {void}
    */
   public first (): void {
-    const firstItem = this.itemMap.get(this.tabbableElements[0]);
+    const firstItem = this.itemMap.get(this.tabbableItems[0]);
```
TremayneChrist, line 246: `tabbableItems !== tabbableElements` and includes all tababble nodes in the tree. This can return you the wrong information.

packages/elements/src/list/index.ts
```diff
@@ -649,17 +644,14 @@ export class List<T extends DataItem = ItemData> extends ControlElement {
     this.addEventListener('focusout', this.onBlur);
   }
 
-  protected update (changeProperties: PropertyValues): void {
+  /**
+   * Invoked before update() to compute values needed during the update.
+   * @param changeProperties changed properties
+   * @returns {void}
+   */
+  protected willUpdate (changeProperties: PropertyValues): void {
```
TremayneChrist, line 660: Call super?

packages/elements/src/tree/elements/tree-item.ts
```diff
@@ -217,13 +217,11 @@ export class TreeItem<T extends TreeDataItem = TreeDataItem> extends ControlElem
...
-   * Invoked whenever the element is updated
-   * @param changedProperties Map of changed properties with old values
+   * Invoked before update() to compute values needed during the update.
+   * @param changedProperties changed properties
    * @returns {void}
    */
-  protected update (changedProperties: PropertyValues): void {
-    super.update(changedProperties);
-
+  protected willUpdate (changedProperties: PropertyValues): void {
```
TremayneChrist, line 224: Call super?

packages/elements/src/tree/helpers/renderer.ts
TremayneChrist: The mode switches the logic in the tree manager

TremayneChrist: This code creates a new tree manager is the data/mode has changed

TremayneChrist (changes requested):

packages/elements/src/tree/helpers/renderer.ts
wsuwt: Thanks, got it. I've reverted back the changes and some workaround regarding `super()` issue. By **TypeScript 4.6**, this rule will be more lenient https://devblogs.microsoft.com/typescript/announcing-typescript-4-6-rc/#code-before-super and we can remove this workaround.

packages/elements/src/list/index.ts
wsuwt: No, It is not required in both cases.

packages/elements/src/list/index.ts
wsuwt: Could you clarify a bit more ? It is going to be different anyway because in both cases `list` and `tree`, children won't be focusable.

TremayneChrist: ~~`tabbableElements` always returns the item element nodes (`ef-list > *`). `tabbableItems` goes through the whole tree I believe (including shadow root), so if you had a custom renderer with a focusable element in it, it would be included in the result.~~

TremayneChrist: Oh wait - it looks like it was modified before and you're reverting back to use the correct value.`tabbableItems` would be correct in this case.

TremayneChrist: So... `tabbableItems` always returns the item element nodes (`ef-list > *`). `tabbableElements` goes through the whole tree I believe (including shadow root), so if you had a custom renderer with a focusable element in it, it would be included in the result.

TremayneChrist: But like you said, this needs to change anyway now as the list items are no longer focuable. Custom renderers can render focusable items, however - I suppose that's down to the devs to manage though.

packages/elements/src/list/index.ts
TremayneChrist: So this doesn't break any of Lit's lifecycle by not being called?

packages/elements/src/tree/helpers/renderer.ts
TremayneChrist: I wouldn't worry about the key property too much. Renderers need to be revisited anyway.

packages/elements/src/list/index.ts
wsuwt: Yes, apparently so, https://lit.dev/docs/components/lifecycle/#reactive-update-cycle-performing. We should try using `willUpdate()` instead of `update()` where possible.

packages/elements/src/list/index.ts
```diff
@@ -353,22 +342,21 @@ export class List<T extends DataItem = ItemData> extends ControlElement {
...
   /**
    * Gets the available tabbable elements
    */
   protected get tabbableItems (): HTMLElement[] {
-    return Array.from(this.children)
-    .filter((el): el is HTMLElement => (el as HTMLElement).tabIndex >= 0);
+    return Array.from(this.children).filter((el): el is HTMLElement => {
+      const role = (el as HTMLElement).getAttribute('role');
+      return (role === 'option' || role === 'treeitem') && (el as HTMLElement).getAttribute('disabled') === null;
+    });
```
TremayneChrist: Make it a bit more readable
```suggestion
    return Array.from(this.children).filter((el): el is HTMLElement => {
      if (el instanceof HTMLElement) {
        const isEnabled = !el.hasAttribute('disabled');
        const isOption = ['option', 'treeitem'].includes( el.getAttribute('role'));
        return isOption && isEnabled;
      }
      return false;
    });
```

packages/elements/src/list/index.ts
```diff
@@ -353,22 +342,21 @@ export class List<T extends DataItem = ItemData> extends ControlElement {
    */
   protected highlightItem (item?: T, scrollToItem = false): void {
     if (item) {
-      const elementToFocus = this.elementFromItem(item);
-      const focus = this.activeElement && this.activeElement !== elementToFocus;
       this.clearHighlighted();
       this.composer.setItemPropertyValue(item, 'highlighted', true);
-      focus && elementToFocus?.focus({ preventScroll: true });
+      (this.tabIndex >= 0 && this.renderer.key && item.value) && this.setAttribute('aria-activedescendant', `${this.renderer.key}-${item.value}`);
```
TremayneChrist: Would be nice to make this more readable by moving the key generation to a helper.

```suggestion
      const activeKey = getItemKey(this.renderer.key, item.value);
      this.tabIndex >= 0 && activeKey && this.setAttribute('aria-activedescendant', activeKey);
```

TremayneChrist (commented):
Hey Phet,

Looks good. The only thing is the readability of some parts. Would be nice to make this a bit clearer for others.

Sarin-Udompanish (approved):

wsuwt:
> 1 review requesting changes and 1 approving review by reviewers with write access

Hi @TremayneChrist , I extracted `getItemId` into a function (not using `getItemKey` because it's duplicated in composer) and work the solution out for custom renderer by having to clone the key from original renderer. Please see, [cloning renderer key](https://github.com/Refinitiv/refinitiv-ui/pull/191/files#diff-0db67da8f50aa5ee56b9b7f1decfdc15e74ae463111d27a5d8e23ea048d920cdR118)

packages/elements/src/list/helpers/id.ts
```diff
@@ -1,10 +1,10 @@
...
- * Combines prefix and a string to give unique item key
+ * Combines prefix and a string to give unique item id used by aria-activedescendant
  * @param prefix a unique randomly generated string
  * @param value a string of valid item value
- * @returns unique item key
+ * @returns a unique item id
  */
-const getItemKey = (prefix: string, value: string): string => {
+const getItemId = (prefix: string, value: string | undefined): string => {
   const SEPARATOR = '-';
```
TremayneChrist: Can we remove this const and just place the separator in the string? You probably want to rename the file, too.`item-id.ts`?

wsuwt: Thanks, agreed and fixed.

TremayneChrist (approved):

PR #196, opened 2022-02-08
Fix cannot clear value when using free text mode

Sakchai-Refinitiv:
## Description
The value cannot be cleared by using programmatically when enabled free text mode and set custom text to the input.

Fixes https://jira.refinitiv.com/browse/ELF-1336

## Type of change

## Checklist
- [x]  My code follows the style guidelines of this project


goremikins:
It should be possible to not just clear the free text value, but to set it as well, if that value does not exist in the set. @TremayneChrist , do you agree?

TremayneChrist:
Yes. If free text mode is set, you should be able to do...

```js
el.value = '';
el.value = 'any string value I like';
```

goremikins:
For instance like this:
```
     else {
       // Clone value arrays
       const newValues = values.slice(0, this.multiple ? values.length : 1);
      const oldValues = this.values.slice();
       // Create comparison strings to check for differences
       const newComparison = newValues.sort().toString();
       const oldComparison = oldValues.sort().toString();
       // Should we update the selection state?
       if (newComparison !== oldComparison) {
         this.updateComposerValues(newValues);
        if (this.freeText) {
          this.freeTextValue = !this.composerValues.length ? newComparison : '';
       }

         this.requestUpdate('values', oldValues);
       }

```

Sakchai-Refinitiv:
Hi All, thank you for your review, I have already patched it.

Theeraphat-Sorasetsakul:
LGTM now. Could you please double check @goremikins @TremayneChrist ?

Theeraphat-Sorasetsakul (approved):

packages/elements/src/combo-box/index.ts
```diff
@@ -279,13 +279,19 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
...
       const newComparison = newValues.sort().toString();
       const oldComparison = oldValues.sort().toString();
-      // Should we update the selection state?
+
+      // Update the selection state when found new value
       if (newComparison !== oldComparison) {
         this.updateComposerValues(newValues);
+
+        if (this.freeText) {
+          this.freeTextValue = !this.composerValues.length ? newComparison : '';
```
goremikins, line 292: I think here is more correct to set newValues[0]: `this.freeTextValue = !this.composerValues.length ? newValues[0] : '';`

Sakchai-Refinitiv: Resolved

packages/elements/src/combo-box/index.ts
```diff
@@ -279,13 +279,19 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
...
       const newComparison = newValues.sort().toString();
       const oldComparison = oldValues.sort().toString();
-      // Should we update the selection state?
+
+      // Update the selection state when found new value
       if (newComparison !== oldComparison) {
         this.updateComposerValues(newValues);
+
+        if (this.freeText) {
+          this.freeTextValue = !this.composerValues.length ? newValues[0] : '';
```
TremayneChrist, line 292:
```suggestion
          // free text mode is only supported in single selection mode
          // so if there is no valid selection in the composer, we can assume
          // the first item can be used as the free text item.
          this.freeTextValue = !this.composerValues.length ? newValues[0] : '';
```

Theeraphat-Sorasetsakul:
Any update? Can this pull request be merged? @Sakchai-Refinitiv @TremayneChrist

Sakchai-Refinitiv:
> Any update? Can this pull request be merged? @Sakchai-Refinitiv @TremayneChrist

I am waiting for @TremayneChrist to confirm for merge.

TremayneChrist (approved):

PR #197, opened 2022-02-09
Feature/tab bar manage active state

Sarin-Udompanish:
## Description
From accessibility spec of tab-bar when user press tab button it should focus on current active tab and when press tab again it should focus out of this tab bar. That mean user can press tab button only once per tab. 

To achieve this we need to set tabIndex of other tab to -1 and when the active tab changed it should reassign all tabIndex. The problem is currently our active state of tab-bar is managed by user so user need to reassign tab index by themselves and that not what we want.

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/elements/src/tab-bar/index.ts
+++ b/packages/elements/src/tab-bar/index.ts
@@ -290,7 +290,6 @@ export class TabBar extends ResponsiveElement {
 
     const leftBtnStyle = scrollLeft > 0 ? 'flex' : 'none';
     const rightBtnStyle = scrollWidth - scrollLeft - elementWidth > 1 ? 'flex' : 'none';
-    
 
     this.leftBtn.style.setProperty('display', leftBtnStyle);
     this.rightBtn.style.setProperty('display', rightBtnStyle);
--- a/packages/elements/src/tab-bar/index.ts
+++ b/packages/elements/src/tab-bar/index.ts
@@ -169,11 +169,6 @@ export class TabBar extends ResponsiveElement {
    * @returns {void}
    */
   private onSlotChange (): void {
-    // wait element size updated
-    setTimeout(() => {
-      this.toggleScrollButton(this.content.clientWidth);
-    });
-
     const tabList = this.getFocusableTabs();
 
     if (tabList.length < 1) {
--- a/packages/elements/src/tab-bar/index.ts
+++ b/packages/elements/src/tab-bar/index.ts
@@ -12,8 +12,8 @@ import { property } from '@refinitiv-ui/core/lib/decorators/property.js';
 import { query } from '@refinitiv-ui/core/lib/decorators/query.js';
 import { VERSION } from '../version.js';
 import { tweenAnimate } from './helpers/animate.js';
-import { Tab } from '../tab';
-import type { Button } from '../button';
+import { Tab } from '../tab/index.js';
+import type { Button } from '../button/index.js';
 import '../button/index.js';
 
 const BAR_TRAVEL_DISTANCE = 150; // scroll distance
@@ -90,7 +90,7 @@ export class TabBar extends ResponsiveElement {
     const oldValue = this._value;
     if (value !== oldValue && this.isValidValue(value)) {
       this._value = value;
-      this.selectValue(value);
+      this.activateTab(value);
       this.requestUpdate('value', oldValue);
     }
   }
@@ -168,20 +168,22 @@ export class TabBar extends ResponsiveElement {
    * @returns {void}
    */
   private onSlotChange (): void {
-    this.setLevel();
-    const tabList = this.getFocusableTabs();
-    if (tabList.length < 1) {
-      return;
-    }
-    const activeTab = tabList.find(tab => tab.active) || tabList[0];
-
-    if (activeTab) {
-      this.value = this.getTabValue(activeTab);
-    }
     // wait element size updated
     setTimeout(() => {
       this.toggleScrollButton(this.content.clientWidth);
     });
+
+    const tabList = this.getFocusableTabs();
+
+    if (tabList.length < 1) {
+      return;
+    }
+    this.setLevel();
+    // get tab value from active tab
+    const activeTab = tabList.find(tab => tab.active) || tabList[0];
+    if (activeTab) {
+      this.value = this.getTabValue(activeTab);
+    }
   }
 
   /**
@@ -189,7 +191,7 @@ export class TabBar extends ResponsiveElement {
    * @param value value of tab to select
    * @returns {void}
    */
-  private selectValue (value: string): void {
+  private activateTab (value: string): void {
     if (!value) {
       return;
     }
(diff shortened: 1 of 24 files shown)


packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +154,147 @@ export class TabBar extends ResponsiveElement {
   }
 
   /**
-   * Hide all scroll buttons
-   * @returns {void}
+   * Return true if tab list have matched value
```
wsuwt, line 157:
```suggestion
   * Return true if incoming value matches one of the existing tab
```

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +154,147 @@ export class TabBar extends ResponsiveElement {
...
-    if (scrollLeft > 0) {
-      this.leftBtn.style.setProperty('display', 'flex');
+    if (activeTab) {
+      this.value = this.getTabValue(activeTab);
     }
-    else {
-      this.leftBtn.style.setProperty('display', 'none');
+    // wait element size updated
+    setTimeout(() => {
+      this.toggleScrollButton(this.content.clientWidth);
```
wsuwt, line 183: Ideally, I think this should be handled when element resizes

packages/elements/src/tab-bar/index.ts
```diff
@@ -11,15 +11,16 @@ import { customElement } from '@refinitiv-ui/core/lib/decorators/custom-element.
 import { property } from '@refinitiv-ui/core/lib/decorators/property.js';
 import { query } from '@refinitiv-ui/core/lib/decorators/query.js';
 import { VERSION } from '../version.js';
-import type { Tab } from '../tab';
 import { tweenAnimate } from './helpers/animate.js';
+import { Tab } from '../tab';
```
wsuwt, line 15: Please add .js extension.

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +154,147 @@ export class TabBar extends ResponsiveElement {
...
+      this.toggleScrollButton(this.content.clientWidth);
+    });
+  }
+
+  /**
+   * Mark tab as active
+   * @param value value of tab to select
+   * @returns {void}
+   */
+  private selectValue (value: string): void {
```
wsuwt, line 192: I think using `active` across the code would be easier to navigate, using `select` may cause confusion with the element state. Something like `activateTab`, `setActiveTab` etc. would sound better.

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +154,147 @@ export class TabBar extends ResponsiveElement {
...
+   */
+  private getTabLabel (tab: Tab): string {
+    return tab.label || tab.textContent || '';
+  }
+
+  /**
+   * Get Tab elements from slot
+   * @returns the array of Tab
+   */
+  private getTabElements () {
```
wsuwt, line 252: Please add return type.

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +154,147 @@ export class TabBar extends ResponsiveElement {
...
    * @returns {void}
    */
-  private toggleScrollButton (elementWidth: number): void {
-    const { scrollLeft, scrollWidth } = this.content;
-
-    if (this.vertical) {
+  private onSlotChange (): void {
+    this.setLevel();
+    const tabList = this.getFocusableTabs();
+    if (tabList.length < 1) {
```
wsuwt, line 173: Can we check `tabList` length before setting the level ?

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +154,147 @@ export class TabBar extends ResponsiveElement {
...
+   * Hide/Show scroll button when element is overflow.
+   * @param elementWidth width of element
    * @returns {void}
    */
-  private onSlotChange (): void {
-    this.setLevel();
+  private toggleScrollButton (elementWidth: number): void {
+    const { scrollLeft, scrollWidth } = this.content;
+
+    if (this.vertical) {
```
wsuwt, line 289: Can this be the first statement of the function ?

packages/elements/src/tab/index.ts
```diff
@@ -88,26 +104,17 @@ export class Tab extends ControlElement {
...
-  /**
-   * @param node that should be checked
-   * @returns whether node can be ignored.
-   */
-  private isIgnorable (node: Node): boolean {
-    return node.nodeType === document.COMMENT_NODE
-      || isAllWhitespaceTextNode(node);
-  }
+  @state()
+  private isSlotHasContent = true;
```
wsuwt, line 108: Why is this `true` ?

packages/elements/src/tab/index.ts
```diff
@@ -88,26 +104,17 @@ export class Tab extends ControlElement {
...
 
   /**
-   * Checks slotted children nodes and updates component to refresh label and sub-label templates.
+   * Run on default slot slotchange
```
wsuwt, line 111:
```suggestion
   * Triggers when slot detect changes
```

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +154,147 @@ export class TabBar extends ResponsiveElement {
...
+        tab.active = false;
+      }
+    });
+  }
 
-    // handle right button
-    if (Math.floor(scrollWidth - scrollLeft) > Math.round(elementWidth)) {
-      this.rightBtn.style.setProperty('display', 'flex');
+  /**
+   * If the target of the event is a Tab, then set the value of the target to the value of the Tab
```
wsuwt, line 212:
```suggestion
   * Set tab value and fires `tab-changed` event
```

packages/elements/src/tab-bar/index.ts
Sarin-Udompanish: fixed

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +155,145 @@ export class TabBar extends ResponsiveElement {
...
-   * Hide all scroll buttons
-   * @returns {void}
+   * Return true if incoming value matches one of the existing tabs
+   * @param value Value to check
+   * @returns true if incoming value matches one of the existing tabs
    */
-  private hideScrollButtons (): void {
-    this.leftBtn.style.setProperty('display', 'none');
-    this.rightBtn.style.setProperty('display', 'none');
+  private isValidValue (value: string) {
```
wsuwt, line 162:
```suggestion
  private isValidValue (value: string): boolean {
```

packages/elements/src/tab-bar/index.ts
```diff
@@ -134,61 +155,145 @@ export class TabBar extends ResponsiveElement {
...
+  private toggleScrollButton (elementWidth: number): void {
+    if (this.vertical) {
+      return;
+    }
+
+    const { scrollLeft, scrollWidth } = this.content;
+
+    const leftBtnStyle = scrollLeft > 0 ? 'flex' : 'none';
+    const rightBtnStyle = scrollWidth - scrollLeft - elementWidth > 1 ? 'flex' : 'none';
+
```
wsuwt, line 293:
```suggestion
```

wsuwt (approved):

PR #199, opened 2022-02-09
feat(collapse): add accessibility

goremikins:
I have made a draft of what I think it should be. It works fine apart of these things:
1) I cannot use ef-header because it would not overflow text correctly in this scenario
2) I cannot use spacers to correctly pad slots (as in that case clickable area would not include these spacers). Do not think this is a big issue anyway. If really required this can be calculated programmatically.

... and I have not tested that with screen readers, hope it works :)

re: @TremayneChrist

### DIFF
--- a/packages/elements/src/collapse/index.ts
+++ b/packages/elements/src/collapse/index.ts
@@ -9,8 +9,9 @@ import {
 import { customElement } from '@refinitiv-ui/core/lib/decorators/custom-element.js';
 import { property } from '@refinitiv-ui/core/lib/decorators/property.js';
 import { query } from '@refinitiv-ui/core/lib/decorators/query.js';
+import { ifDefined } from '@refinitiv-ui/core/lib/directives/if-defined.js';
+import { state } from '@refinitiv-ui/core/lib/decorators/state.js';
 import { VERSION } from '../version.js';
-import '../header/index.js';
 import '../panel/index.js';
 import '../icon/index.js';
 
@@ -48,12 +49,13 @@ export class Collapse extends BasicElement {
       :host {
         display: block;
       }
-      [part="header"] {
-        cursor: default;
+      [part="header"], [part="header-toggle"] {
+        display: flex;
+        flex: 1 1 auto;
+        min-width: 0;
       }
       [part="toggle"] {
-        display: inline-block;
-        margin: 0 5px;
+        flex: none;
       }
       [part="content"]  {
         overflow: hidden;
@@ -68,6 +70,14 @@ export class Collapse extends BasicElement {
         }
       }
     `;
+  }
+
+  /**
+   * Observes attribute change for `attributeChangedCallback`
+   */
+  static get observedAttributes (): string[] {
+    const observed = super.observedAttributes;
+    return ['aria-level'].concat(observed);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


goremikins:
Hi, I have made a draft of what I think it should be. It works fine apart of the fact that I cannot use ef-header because it would not overflow text correctly in this scenario.

re: @TremayneChrist

dtanp-rft:
@goremikins Could you please elaborate more about overflow? Do you mean ellipsis on long header text? It works fine with my PR but it may be other scenario :)

I noticed about overflow content doesn't work as well yesterday but didn't have time to look at it. Thanks for the fix and some code in this PR are great except the point that we want to divorce Collapse from Header. Personally, I don't think it's a good idea as it will create inconsistency in our Design System.

This make me remember another issue though. Already discussed with Trem that our Panel doesn't manage scrollbar style out-of-the-box. I think we should do it -- can be totally separate issue.

packages/elements/src/collapse/index.ts
```diff
@@ -173,16 +190,25 @@ export class Collapse extends BasicElement {
...
+               aria-expanded="${this.expanded}"
+               aria-controls="content"
+               @tap=${this.toggle}>
+            <ef-icon icon="right" part="toggle"></ef-icon>
+            <span part="header-label">${this.header}</span>
+          </div>
+        </div>
         <slot name="header-right"></slot>
       </div>
       <div id="content" part="content" role="region" aria-labelledby="header-toggle">
```
wsuwt, line 210: `aria-labelledby="header-toggle"` - id `header-toggle` does not exist

packages/elements/src/collapse/index.ts
```diff
@@ -106,6 +113,16 @@ export class Collapse extends BasicElement {
   @query('ef-panel', true)
   private panel!: HTMLElement;
 
+  @state()
+  private headingLevel: string | null = null;
+
+
```
wsuwt, line 119:
```suggestion
```

packages/elements/src/collapse/index.ts
goremikins: aye, copied from somewhere

goremikins:
@dtanp-rft , sorry if I was unclear, but my comment about overflow was related to this peace of code, which is semantically correct:

```
<ef-header part="header-wrapper" level="${this.level}">
  <slot name="header-left" slot="left"></slot>
  <slot name="header-right" slot="right"></slot>
  <div part="header"
       role="heading"
       aria-level="${ifDefined(this.headingLevel || undefined)}">
    <div part="header-toggle"
         role="button"
         tabindex="0"
         aria-expanded="${this.expanded}"
         aria-controls="content"
         @tap=${this.toggle}>
      <ef-icon icon="right" part="toggle"></ef-icon>
      ${this.header}
    </div>
  </div>
</ef-header>
```

I do not like your solution for two reasons:
1. According to specs "the button element is the only element inside the heading element. That is, if there are other visually persistent elements, they are not included inside the heading element."
2. slotted items in your scenario are still outside `ef-header` and it looks correct only because you apply background-color to the `part=header` (and btw it is broken in elemental theme).

I agree, I would rather use `ef-header` with the code snippet above, but since I was not able to make it working correctly, I did it without `ef-header`. Speaking about the design system, I use shared styles for the header (take a look on `ef-collapse.less`), so that should not be a problem.

packages/elements/src/collapse/index.ts
goremikins: Fixed

PR #200, opened 2022-02-10
feat(pill): add accessibility support

bualoy-napat:
## Description
Pill support screen reader

## Type of change

## Checklist


packages/elements/src/pill/index.ts
```diff
@@ -86,6 +104,25 @@ export class Pill extends ControlElement {
...
+  /**
+   * Aria indicating state of toggle pill
+   * @ignore
+   */
+  @property({ type: String,
+    reflect: true,
+    attribute: 'aria-pressed',
+    converter: { toAttribute: emptyStringToNull } // TODO: Remove after typescript update to allow nullable for ARIAMixin
+  })
+  public ariaPressed = '';
```
Sakchai-Refinitiv: Should not override the area attributes because it breaks AOM, you can use JavaScript set/remove attributes instead.

packages/elements/src/pill/index.ts
```diff
@@ -11,9 +11,22 @@ import { customElement } from '@refinitiv-ui/core/lib/decorators/custom-element.
...
+ * Return the attribute that converted from the property
+ * Prevent empty string that reflected to attribute
+ * @private
+ * @param value value from the property
+ * @returns string converted to attribute
+ */
+const emptyStringToNull = function (value: string): string | null {
+  return value || null;
+};
+
```
Sakchai-Refinitiv: This is simple logic in the function, I think it should not be the function.

packages/elements/src/pill/index.ts
```diff
@@ -42,6 +55,11 @@ export class Pill extends ControlElement {
     return VERSION;
   }
 
+  /**
+   * Element's role attribute for accessibility
+   */
+  protected readonly defaultRole: string | null = 'button';
```
Sakchai-Refinitiv, line 48: Why default role is `button`?

packages/phrasebook/src/locale/de/pill.ts
```diff
@@ -0,0 +1,9 @@
+import { Phrasebook } from '../../translation.js';
+
+const translations = {
+  DELETE_PILL: 'Press delete key to remove pill'
```
Sakchai-Refinitiv: I think we can use generic messages to reuse in multiple components

packages/elements/src/pill/index.ts
bualoy-napat: Pill is a small button style component, I think that we should set role on the pill as `button` by default

packages/elements/src/pill/index.ts
bualoy-napat: I have already updated on the recent commit.

packages/elements/src/pill/index.ts
bualoy-napat: I have already updated on the recent commit.

packages/elements/src/pill/index.ts
Sakchai-Refinitiv: It is incorrect when the pill has a clear button because it will be the anti-pattern nested interactive element. IMO the role should be button in the toggle mode only.

Sakchai-Refinitiv (commented):
The role of part content must be `none`

packages/elements/src/pill/index.ts
```diff
@@ -95,7 +100,37 @@ export class Pill extends ControlElement {
     this.addEventListener('tapstart', this.onStartPress);
     this.addEventListener('tapend', this.onEndPress);
     this.addEventListener('mouseleave', this.onEndPress);
+    this.addEventListener('keydown', this.onKeyDown);
+
```
wsuwt:
```suggestion
```

packages/elements/src/pill/index.ts
```diff
@@ -95,7 +100,37 @@ export class Pill extends ControlElement {
...
+    super.update(changedProperties);
+  }
 
+  /**
+   * Handles key down event
+   * @param event Key down event object
+   * @returns {void}
+   */
+  private onKeyDown (event: KeyboardEvent): void {
+    if (event.key === 'Delete' && (this.clears && !this.readonly)) {
```
wsuwt: Did you check on IE11 ?

packages/halo-theme/src/custom-elements/ef-pill.less
```diff
@@ -36,6 +36,10 @@
     }
   }
 
+  &[focused="visible"] {
+    border: 1px solid @scheme-color-primary;
+  }
```
wsuwt, line 41: Why do you need this ? I think pill style should be inherited from elemental without the need to add another line of style.

packages/elements/src/pill/index.ts
```diff
@@ -95,7 +100,37 @@ export class Pill extends ControlElement {
...
+    this.addEventListener('keydown', this.onKeyDown);
+
+  }
+
+  /**
+   * Updates the element
+   * @param changedProperties Properties that has changed
+   * @returns {void}
+   */
+  protected update (changedProperties: PropertyValues): void {
```
wsuwt, line 111: Should we use `willUpdate()` or `update()` ?

packages/elements/src/pill/index.ts
Sakchai-Refinitiv: We agree to set `button` as a default role

packages/elements/src/pill/index.ts
bualoy-napat: I think we should use `update()`

packages/halo-theme/src/custom-elements/ef-pill.less
bualoy-napat: For halo theme, we inherit the theme from elemental as a reference. So, if we define it on elemental, this style will not inherit into the halo theme.

packages/elements/src/pill/index.ts
bualoy-napat: For IE 11, I need to check for the `Del` key.
```suggestion
    if ((event.key === 'Delete' || event.key === 'Del') && (this.clears && !this.readonly)) {
```

packages/elements/src/pill/__test__/pill.test.js
```diff
@@ -282,4 +282,36 @@ describe('pill/Pill', () => {
...
+
+      setTimeout(() => el.dispatchEvent(new Event('tap')));
+      await oneEvent(el, 'tap');
+      expect(el.getAttribute('aria-pressed')).to.equal('true');
+
+      setTimeout(() => el.dispatchEvent(new Event('tap')));
+      await oneEvent(el, 'tap');
+      expect(el.getAttribute('aria-pressed')).to.equal('false');
+    });
+    it('aria-pressed should be removed when toggles attribute is removed', async () => {
```
Sakchai-Refinitiv, line 308: Should have line space between test cases for readability.

bualoy-napat: I have already updated

Sakchai-Refinitiv (approved):

PR #202, opened 2022-02-10
Accessibility/tab and tab bar

Sarin-Udompanish:
## Description

There are 2 examples in tab specs, choose one which you think that it is the best approach to implement with our tab. 

1. https://www.w3.org/TR/wai-aria-practices-1.2/examples/tabs/tabs-1/tabs.html

2. https://www.w3.org/TR/wai-aria-practices-1.2/examples/tabs/tabs-2/tabs.html

**tab**

- should have `role="tab"`
- should have `aria-selected`: `true | false` to indicate current active tab 
- should support `Delete` key

**tab-bar**

- should have `role="tablist"`
- should support `Left/Right/Up/Down Arrow` keyboard navigation across tabs
- should support `Home` and `End` which navigates to the first and last tab. 

## Type of change

## Checklist


packages/elements/src/tab-bar/__test__/tab-bar.accessibility.test.js
```diff
@@ -0,0 +1,110 @@
...
+    tabList = el.querySelectorAll('ef-tab');
+    tabList[0].focus();
+    el.dispatchEvent(keyArrowRight);
+    expect(el.value).to.equal('');
+    el.dispatchEvent(keyHome);
+    expect(el.value).to.equal('');
+    el.dispatchEvent(keyEnd);
+    expect(el.value).to.equal('');
+  });
+  it('Navigation', async () => {
```
wsuwt, line 67:
```suggestion
  it('Should be able to navigate using Arrows, Home and End key correctly', async () => {
```

packages/elements/src/tab-bar/index.ts
```diff
@@ -329,18 +333,136 @@ export class TabBar extends ResponsiveElement {
     tweenAnimate({ target: this.content, startPosition: scrollLeft, endPosition });
   }
 
+  /**
+   * focus and set active to tab
```
wsuwt, line 337:
```suggestion
   * Focus and set active to tab
```

packages/elements/src/tab-bar/index.ts
```diff
@@ -329,18 +333,136 @@ export class TabBar extends ResponsiveElement {
...
+        this.last();
+        break;
+      default:
+        return;
+    }
+    event.preventDefault();
+  }
+
+  /**
+   * Sets the tabindex to -1 for all tabs except the currently focused one.
```
wsuwt, line 430:
```suggestion
   * Sets the tabindex to -1 for all tabs except the active tab```

File: packages/elements/src/tab/index.ts [hunk 4]
```diff
@@ -52,6 +55,13 @@ export class Tab extends ControlElement {
     `;
   }
 
+  /**
+   * Aria indicating current select state
+   * @ignore
+   */
+  @property({ type: String, reflect: true, attribute: 'aria-selected' })
+  public ariaSelected = 'false';
```
wsuwt, line 63: @Sarin-Udompanish Can we use `this.setAttribute('aria-selected', ... )` instead. I think we have come to an agreement that we do prefer it that way rather than declaring a new Lit property.

packages/elements/src/tab/index.ts
```diff
@@ -107,6 +131,17 @@ export class Tab extends ControlElement {
   @state()
   private isSlotHasContent = false;
 
+  /**
+   * Called after the element’s DOM has been updated the first time.
+   * register scroll event on content element to toggle scroll button
```
wsuwt, line 136: Scroll button ?

packages/elements/src/tab/index.ts
Sarin-Udompanish: Fixed

wsuwt: So we can safely remove this line yeah ?

Sarin-Udompanish: Yes, removed.

packages/elements/src/tab/index.ts
```diff
@@ -72,6 +75,8 @@ export class Tab extends ControlElement {
 
   /**
    * Specify tab's active status
+   * @param value active value
+   * @default false
```
wsuwt, line 79:
```suggestion
```

Sarin-Udompanish: fixed

wsuwt (approved):

PR #203, opened 2022-02-10
feat(collapse): accessibility support

goremikins:
## Description
Support screen reader for collapse element.
- add `aria-level` attribute and delegate `value` to the header element for semantic meaning.
- add the `header-toggle` part which has a `role button` and wraps header text to make semantic meaning for a screen reader accessible
- hide the `arrow icon` from a screen reader because of using the button in the `header-toggle` part instead.
- add `id` and `aria attributes` to the part content to support accessibility when expanded/collapsed states changed.
- using CSS `visibility hidden` property for the content to prevent users accessing the content when collapsed by using keyboard navigation.
- update theme by disabling outline from the `button` in `header-toggle`

## Type of change

## Checklist


packages/elements/src/collapse/index.ts
```diff
@@ -48,12 +51,22 @@ export class Collapse extends BasicElement {
       :host {
         display: block;
       }
-      [part="header"] {
-        cursor: default;
+      :host(:not([expanded])) [part="content"] {
+        visibility: hidden;
```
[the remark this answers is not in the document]
goremikins: Checked in Chrome, Firefox and IE11. It does work as expected and elements are not focusable when collapsed.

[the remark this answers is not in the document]
Sakchai-Refinitiv: Sorry, I tested on the wrong branch. There are many branches for collapse it makes me confused.

packages/elements/src/collapse/index.ts
```diff
@@ -185,24 +198,33 @@ export class Collapse extends BasicElement {
...
+        <div part="heading" role="heading" aria-level="${ifDefined(this.headingLevel || undefined)}">
+          <div id="header-toggle"
+               part="header-toggle"
+               role="button"
+               tabindex="0"
+               aria-expanded="${this.expanded}"
+               aria-controls="content"
+               @tap=${this.toggle}>${this.header}</div>
+        </div>
+        <ef-icon icon="right" part="toggle" slot="left" aria-hidde="true"></ef-icon>
```
Sakchai-Refinitiv, line 222: typo `aria-hidden`. it makes the user can navigate to the icon.

goremikins: Fixed

packages/elements/src/collapse/index.ts
```diff
@@ -185,24 +198,33 @@ export class Collapse extends BasicElement {
...
+          <div id="header-toggle"
+               part="header-toggle"
+               role="button"
+               tabindex="0"
+               aria-expanded="${this.expanded}"
+               aria-controls="content"
+               @tap=${this.toggle}>${this.header}</div>
+        </div>
+        <ef-icon icon="right" part="toggle" slot="left" aria-hidden="true"></ef-icon>
+        <slot name="header-left" slot="left"></slot>
```
Sakchai-Refinitiv, line 223: I have discussed with Bangkok Team, the header-left slot should place on the left most of the header because it would be good to custom style for developers. The result should look like this image below, you can try it on the demo page.
Screen Shot 2022-02-11 at 14 42 31

goremikins:
@Sakchai-Refinitiv , I have done testing and changed the order of slots. If you are happy, it can be merged.

Sakchai-Refinitiv:
> @Sakchai-Refinitiv , I have done testing and changed the order of slots. If you are happy, it can be merged.

Hi @goremikins, let me check it if not found any issue I will merge this PR.

packages/elements/src/collapse/index.ts
TremayneChrist: I have concerns about this as it's an anti-pattern. If you compare the visual design to tree, the slots are in the wrong order - the expand toggle should be on the left.

image

Sakchai-Refinitiv: Hi Trem, in the use case that you mention the developers can place the checkbox, icon, and text in the main slot instead.

goremikins:
@Sakchai-Refinitiv , please check the code again and approve if you are happy

Sakchai-Refinitiv (approved):

PR #226, opened 2022-02-21
Add iterate and locale parsing date utils

goremikins:
## Description

1. Add ability to parse localised dates.
2. Add ability to add or subtract units.
3. Add ability to iterate over date units.

## Type of change

## Checklist


packages/utils/src/date/timestamps.ts
```diff
@@ -1,5 +1,11 @@
...
-export const MILLISECONDS_IN_DAY = 86400000;
-export const MILLISECONDS_IN_HOUR = 3600000;
-export const MILLISECONDS_IN_MINUTE = 60000;
+export const YEARS_IN_CENTURY = 100;
+export const MONTHS_IN_YEAR = 12;
+export const HOURS_IN_DAY = 24;
+export const MINUTES_IN_HOUR = 60;
+export const SECONDS_IN_MINUTE = 60;
 export const MILLISECONDS_IN_SECOND = 1000;
+export const MILLISECONDS_IN_MINUTE = SECONDS_IN_MINUTE * MILLISECONDS_IN_SECOND;
```
TremayneChrist:
```suggestion
export const MILLISECONDS_IN_MINUTE = MILLISECONDS_IN_SECOND * SECONDS_IN_MINUTE;
```
Keeping consistent with others

packages/utils/src/date/shared.ts
```diff
@@ -376,30 +384,60 @@ const addMonths = (value: string, amount: number): string => {
...
+    case 'minute':
+      date.setUTCMinutes(date.getUTCMinutes() + amount);
+      break;
+    case 'second':
+      date.setUTCSeconds(date.getUTCSeconds() + amount);
+      break;
+    case 'millisecond':
+      date.setUTCMilliseconds(date.getUTCMilliseconds() + amount);
+      break;
+    // no default
```
TremayneChrist: Throw error if wrong unit is set

packages/utils/src/date/shared.ts
```diff
@@ -456,7 +494,63 @@ const addOffset = (value: string, amount: number): string => isTime(value) ? add
...
+    case 'minute':
+      date.setUTCMinutes((date.getUTCMinutes() + (MINUTES_IN_HOUR + amount) % MINUTES_IN_HOUR) % MINUTES_IN_HOUR);
+      break;
+    case 'second':
+      date.setUTCSeconds((date.getUTCSeconds() + (SECONDS_IN_MINUTE + amount) % SECONDS_IN_MINUTE) % SECONDS_IN_MINUTE);
+      break;
+    case 'millisecond':
+      date.setUTCMilliseconds((date.getUTCMilliseconds() + (MILLISECONDS_IN_SECOND + amount) % MILLISECONDS_IN_SECOND) % MILLISECONDS_IN_SECOND);
+      break;
+    // no default
```
TremayneChrist: Throw if wrong unit

packages/utils/src/date/shared.ts
```diff
@@ -456,7 +494,63 @@ const addOffset = (value: string, amount: number): string => isTime(value) ? add
  */
 const subOffset = (value: string, amount: number): string => addOffset(value, -amount);
 
+/**
+ * Iterate over the date by specified number of units
```
TremayneChrist: This description could be better. I had to read the code to see what is actually going on.

TremayneChrist:
```suggestion
 * Cycles through the unit by a specified amount, not affecting any other part of the date.
```

packages/utils/src/date/Locale.ts
```diff
@@ -0,0 +1,556 @@
...
+  private readonly hourCycle;
+  private readonly fractionalSecondDigits;
+
+  /**
+   * Create the Locale object
+   * @param formatter Date time format
+   */
+  constructor (formatter: Intl.DateTimeFormat) {
+    this.formatter = formatter;
+    this.options = formatter.resolvedOptions();
```
TremayneChrist, line 133: Check if formatter is valid. Throw if not.

packages/utils/src/date/Locale.ts
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  /**
+   * Get weekday names for the specified formatter
+   * @returns names Weekday names
+   */
+  private formatWeekdayNames (): FormattedNames[] {
+    const weekdayNames: FormattedNames[] = [];
+    for (let value = 0; value < DAYS_IN_WEEK; value += 1) {
+      const date = Date.UTC(2018, 0, value); // Sunday - Saturday
```
TremayneChrist, line 195: Move base year into a reusable const and use this across all formatters.

packages/utils/src/date/Locale.ts
```diff
@@ -0,0 +1,556 @@
...
+      // time only formats
+      this._resolvedFormat = millisecond ? TimeFormat.HHmmssSSS : second ? TimeFormat.HHmmss : TimeFormat.HHmm;
+    }
+    else if (year && !hour) {
+      // date only
+      this._resolvedFormat = day ? DateFormat.yyyyMMdd : month ? DateFormat.yyyyMM : DateFormat.yyyy;
+    }
+    else {
+      // datetime or ambiguous
+      this._resolvedFormat = millisecond ? DateTimeFormat.yyyMMddTHHmmssSSS : second ? DateTimeFormat.yyyMMddTHHmmss : DateTimeFormat.yyyMMddTHHmm;
```
TremayneChrist, line 464: What if `minute === false`?

packages/utils/src/date/Locale.ts
```diff
@@ -0,0 +1,556 @@
...
+  public parse (value: string, referenceDate: string | number | Date = 0): string {
+    const refDate = typeof referenceDate === 'string' ? utcParse(referenceDate) : new Date(referenceDate);
+
+    // format reference date to exclude excessive information
+    referenceDate = utcParse(utcFormat(refDate, this.isoFormat));
+
+    value = value.trim(); // weak formatting
+    const regExp = this.regExp;
+
+    regExp.lastIndex = 0; // 0 index?
```
TremayneChrist: Reset index in `regExp` getter if `_regExp` is already defined.

packages/utils/src/date/Locale.ts
```diff
@@ -0,0 +1,556 @@
...
+      year,
+      month,
+      day,
+      hours,
+      minutes,
+      seconds,
+      milliseconds
+    }, this.isoFormat);
+
+    // Check for leap years and number of days per month
```
TremayneChrist:
```suggestion
    // Make sure the generated date is valid.
```

packages/utils/src/date/Locale.ts
```diff
@@ -0,0 +1,556 @@
...
+          break;
+        case 'literal':
+          str += `(${regExpEscape(part.value.toLowerCase())})`;
+          break;
+        case 'era':
+          str += this.escapeList(this.eraNames);
+          break;
+        case 'timeZoneName':
+          throw new Error('Parsing timezones is ambiguous. Make sure that `timeZoneName` option is not provided');
+          break;
```
TremayneChrist: Unreachable code

packages/utils/src/date/timestamps.ts
goremikins: done

packages/utils/src/date/Locale.ts
goremikins: I think TypeScrip validation should be sufficient in this case, and even if something else is provided it will throw an error on the next line

packages/utils/src/date/Locale.ts
goremikins: While I use everywhere 2018, it does not need to be; and the year needs to be considered in context of a formatter itself

packages/utils/src/date/Locale.ts
goremikins: minutes do not matter if seconds are provided. There is no valid ISO value format, which includes seconds, but does not include minutes/hours, similarly there is no valid format that include hours only. So essentially the result will always be one of:

- yyyy
- yyyyMM
- yyyyMMdd
- HHmm
- HHmmss
- HHmmss.SSS
- yyyyMMddTHHmm
- yyyyMMddTHHmmss
- yyyyMMddTHHmmss.SSS

However when string is parsed, the missing bits are populated from the reference date.

packages/utils/src/date/Locale.ts
TremayneChrist: Well I did have another question, why is 2018 so special? :)

packages/utils/src/date/Locale.ts
TremayneChrist: I see.

packages/utils/src/date/Locale.ts
TremayneChrist: You shouldn't rely on TypeScript, it compiles into JS in the end. All public facing APIs should perform checks and report issues.

packages/utils/src/date/Locale.ts
goremikins: I just picked random date, where I know 1st of January is Sunday :)

packages/utils/src/date/Locale.ts
```diff
@@ -209,90 +231,143 @@ class Locale {
...
-    return `(${[...new Set(list.map(item => item.name))].map(item => regExpEscape(item)).join('|')})`;
-  }
-
   private _regExp: RegExp | null = null;
   /**
    * Get regular expression for the format
    */
   private get regExp (): RegExp {
     if (this._regExp) {
+      this._regExp.lastIndex = 0; // 0 index?
```
TremayneChrist: Better comment please. I have no idea what this means :)

goremikins: Comment removed

TremayneChrist (approved):

PR #227, opened 2022-02-22
fix(tree-select): selected button should support disabled item

Theeraphat-Sorasetsakul:
## Description
'Selected' button is not working if all Multi Select selection are disabled

Fixes # (ELF-1713) **jira**: https://jira.refinitiv.com/browse/ELF-1713

**Step to reproduce**: try to press selected button when all selected items are disabled
1. el.data = [{ selected: true, label: '1', value: '1', disabled: true }, { selected: true, label: '2', value: '2', readonly: true }, { label: '3', value: '3' }]
2. try to press Selected button at top right of tree-select

**Expected**: Should be able to press Selected button to filter

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/tree-select/__test__/tree-select.filter.test.js
+++ b/packages/elements/src/tree-select/__test__/tree-select.filter.test.js
@@ -268,5 +268,13 @@ describe('tree-select/Filter', () => {
       expect(el.treeManager.visibleItems.length).to.equal(selectableCount + 2, 'All items are visible');
     });
 
+    it('Should selection filter on disabled selected items when press Selected button', async () => {
+      const el = await fixture('<ef-tree-select lang="en-gb"></ef-tree-select>');
+      const data = [{ selected: true, label: '1', value: '1', disabled: true }, { selected: true, label: '2', value: '2', readonly: true }, { label: '3', value: '3' }];
+      el.data = data;
+      el.selectedClickHandler();
+      const selectedData = data.filter(item => item.selected === true);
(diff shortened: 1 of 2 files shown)


packages/elements/src/tree-select/__test__/tree-select.filter.test.js
```diff
@@ -268,5 +268,13 @@ describe('tree-select/Filter', () => {
       expect(el.treeManager.visibleItems.length).to.equal(selectableCount + 2, 'All items are visible');
     });
 
+    it('Should selection filter on disabled selected items when press Selected button', async () => {
```
wsuwt:
```suggestion
    it('Should allow selected filter button when there is only selected but disabled item in tree', async () => {
```

packages/elements/src/tree-select/__test__/tree-select.filter.test.js
```diff
@@ -268,5 +268,13 @@ describe('tree-select/Filter', () => {
...
+      const el = await fixture('<ef-tree-select lang="en-gb"></ef-tree-select>');
+      const data = [{ selected: true, label: '1', value: '1', disabled: true }, { selected: true, label: '2', value: '2', readonly: true }, { label: '3', value: '3' }];
```
wsuwt: If you want to test `disabled` only use data with `disabled` and test `readonly` separately.

wsuwt (approved):

Sakchai-Refinitiv (approved):

PR #228, opened 2022-02-23
refactor(core): use semantic readonly attribute instead

Sakchai-Refinitiv:
## Description
When using semantic HTML form controls, if you set the readonly attribute, you don't need to include aria-readonly="true"

## Type of change
Refactor

## Checklist

### DIFF
--- a/packages/translate/src/translate.ts
+++ b/packages/translate/src/translate.ts
@@ -59,7 +59,7 @@ class AsyncTranslateDirective extends AsyncDirective {
         // the code may fail if polyfills are not available in IE11 or translate syntax is wrong
         /* istanbul ignore next */
         setTimeout(() => {
-          throw new Error(error);
+          throw new Error(error as string);
         });
       });
 
@@ -78,7 +78,7 @@ const translatePromise = (scope: string, locale: string, key: string, options?: 
     // the code may fail if polyfills are not available in IE11 or translate syntax is wrong
     /* istanbul ignore next */
     setTimeout(() => {
-      throw new Error(error);
+      throw new Error(error as string);
     });
     return key;
   });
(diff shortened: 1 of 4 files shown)


Sakchai-Refinitiv:
Hi @goremikins @TremayneChrist, could you please review this PR for A11y.

packages/translate/src/translate.ts
```diff
@@ -59,7 +59,7 @@ class AsyncTranslateDirective extends AsyncDirective {
         // the code may fail if polyfills are not available in IE11 or translate syntax is wrong
         /* istanbul ignore next */
         setTimeout(() => {
-          throw new Error(error);
+          throw new Error(error as string);
```
goremikins: This is not valid casting. If you know that error is a string, than:
`.catch((error: string) => { .... }`

or if you know that error is an error, than:
`.catch((error: Error) => { .... }`

but in this case I think it can be both. From EF side we always do `throw new Error('')`, but this block can be reached externally.

Sakchai-Refinitiv: In my view, It seems weird if throw the error with the empty string. I think we should check the type of that error and get a message from it, this would be better.
```js
if (typeof error === 'string') {
  throw new Error(error);
}
else if (error instanceof Error) {
  throw error;
}
```
 What do you think?

goremikins: Empty string is an example to indicate that from EF we always create the new error: e.g. `throw new Error('ERROR MESSAGE')`. However, JavaScript allows to throw absolutely anything (e.g. `throw 12345`, or `throw 'STRING'`), that is why lint complains.

The correct syntax is `throw error instanceof Error ? error : new Error(String(error));`

Sakchai-Refinitiv: I see, if there are multiple types of errors in JavaScript, your suggestion is better to handle all of them. Thanks

Sakchai-Refinitiv: Resolved

goremikins (approved):

PR #232, opened 2022-03-02
feat(tooltip): overflow improvements

goremikins:
## Description
Add ability to add tooltip for element, where overflow target is different to overflow condition.

Expectation of this change is title overflowCondition and customRenderer should take priority over overflow tooltip.
Also to provide `unregisterOverflowTooltip` as a pair of `registerOverflowTooltip`, for developers can undo the register overflow tooltip.

This should be applied to V4 as well as V5.

## Type of change

## Checklist
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have checked my code and corrected any misspellings


packages/elements/src/select/index.ts
```diff
@@ -357,6 +365,8 @@ export class Select extends ControlElement implements MultiValue {
   protected firstUpdated (changedProperties: PropertyValues): void {
     super.firstUpdated(changedProperties);
     this.addEventListener('keydown', this.onKeyDown); /* keydown when select is closed */
+
+    registerOverflowTooltip(this, () => this.labelRef.value ? this.labelRef.value.textContent || '' : '');
```
Sakchai-Refinitiv: The `TooltipCondition` is missing cause the tooltip does not work.

goremikins: Thanks, it is fixed

Sakchai-Refinitiv (approved):

PR #234, opened 2022-03-02
Datetime field element

goremikins:
## Description
Add new element datetime field

## Type of change

## Checklist
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] New and existing unit tests pass locally with my changes


packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+export class DatetimeField extends TextField {
+  /**
+   * Element version number
+   * @returns version number
+   */
+  static get version (): string {
+    return VERSION;
+  }
+
+  protected readonly defaultTabIndex = null;
```
wsuwt: Can you add a comment to explain reason why you have set `defaultTabIndex` to null ?

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+  }
+
+  /**
+   * Check if input value should be synchronised with component value
+   * @param changedProperties Properties that has changed
+   * @returns True if input should be synchronised
+   */
+  protected override shouldSyncInputValue (changedProperties: PropertyValues): boolean {
+    // Note: changing any of these properties override the input value
+    // On blur, if the value is correct makes ure strict format is used
```
wsuwt:
```suggestion
    // On blur, if the value is correct makes sure strict format is used
```

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+  /**
+   * Synchronise input value with value.
+   * Override the method if value and input value are not the same
+   * @param changedProperties Properties that has changed
+   * @returns {void}
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected override syncInputValue (changedProperties: PropertyValues): void;
+  protected syncInputValue (): void;
+  protected syncInputValue (changedProperties?: PropertyValues): void {
```
wsuwt, line 358: My eslint warning only go away when the lines are like this. 
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  protected syncInputValue (changedProperties?: PropertyValues): void {
```

wsuwt: I also curious why there are multiple method declarations with the same name, why can't we just override it only once, can you elaborate ?

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+   * change state of `error` property according pattern validation
+   * @returns {void}
+   */
+  protected override validateInput (): void {
+    this.reportValidity();
+  }
+
+  /**
+   * Overwrite validation method for value
+   *
```
wsuwt:
```suggestion
```

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+   * Validate input according `pattern`, `minLength` and `maxLength` properties
+   * change state of `error` property according pattern validation
+   * @returns {void}
+   */
+  protected override validateInput (): void {
+    this.reportValidity();
+  }
+
+  /**
+   * Overwrite validation method for value
```
wsuwt:
```suggestion
   * Override validation method for value
```

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+   *
+   * @param value value
+   * @returns {boolean} result
```
wsuwt:
```suggestion
   * @returns Value validity
```

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+    return getFormat(value) === this.locale.isoFormat;
+  }
+
+  /**
+   * Used to show a warning when the value does not pass the validation
+   * @param value that is invalid
+   * @returns {void}
+   */
+  protected override warnInvalidValue (value: string): void {
+    new WarningNotice(`The specified value "${value}" does not conform to the required format. The format is '${this.locale.isoFormat}'.`).show();
```
wsuwt: Do we want to include `this.localName` as part of the warning ? I think it would enhance ability to debug when there are multiple `ef` elements on a page.

packages/elements/src/datetime-field/types.ts
```diff
@@ -0,0 +1,14 @@
+export type NavigationKeys = 'ArrowLeft' | 'ArrowRight' | 'ArrowUp' | 'ArrowDown' | 'Escape';
```
wsuwt, line 1: When there are selectable parts or segments like this field, do you think we should also add `Home` and `End` support ?

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+   * @param event Key down event object
+   * @returns {void}
+   */
+  protected onInputKeyDown (event: KeyboardEvent): void {
+    if (this.readonly || this.disabled || event.defaultPrevented) {
+      return;
+    }
+
+    // Ignore special keys
+    if (event.shiftKey || event.ctrlKey || event.altKey || event.metaKey) {
```
wsuwt, line 670: Why do we need to handle these keys ?

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+
+  /**
+   * Decorate `<input>` element with common properties.
+   * Override placeholder
+   * @returns template map
+   */
+  protected override get decorateInputMap (): TemplateMap {
+    return {
+      ...super.decorateInputMap,
+      placeholder: this.focused && !this.inputValue ? this.toInputValue(this.startDate) : this.placeholder || undefined,
```
wsuwt: I not sure of replacing placeholder set by the user with start date value is a good idea. Placeholder is something that would provide some kind of a hint or information about the field which we should not override.

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+  /**
+   * Used to announce part and value statuses
+   */
+  protected get screenReaderTemplate (): TemplateResult | null {
+    if (!this.focused) {
+      return null;
+    }
+    return html`
+      ${this.partLabel && this.value ? html`<div
+        aria-label="${this.t(`PICK_${this.partLabel.toUpperCase()}`)}"
```
wsuwt, line 720: My editor says 

`You are binding a non-primitive type 'DirectiveResult<{ prototype: any; new(partInfo: ChildPartInfo | AttributePartInfo | ElementPartInfo) => AsyncTranslateDirective; }>'. This could result in binding the string "[object Object]"`

I know we shouldn't cast `DirectiveResult` to `string` but can we handle this somehow ?

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,708 @@
...
+    this.setSelectionRange(selectionStart, selectionEnd);
+  }
+
+  /**
+   * Change value or selection based on keyboard navigation
+   * @param key Navigation key
+   * @param event Keyboard event
+   * @returns {void}
+   */
+  protected onNavigation (key: NavigationKeys, event: KeyboardEvent): void {
```
wsuwt, line 552: Can we split this method into smaller ones, they are relatable and basically does the similar things but I feel nested switch case is a bit hard to read.

packages/utils/src/date/Locale.ts
```diff
@@ -0,0 +1,556 @@
...
+          break;
+        case 'literal':
+          str += `(${regExpEscape(part.value.toLowerCase())})`;
+          break;
+        case 'era':
+          str += this.escapeList(this.eraNames);
+          break;
+        case 'timeZoneName':
+          throw new Error('Parsing timezones is ambiguous. Make sure that `timeZoneName` option is not provided');
+          break;
```
wsuwt: This `break` is unreachable. Ref https://sonarcloud.io/project/issues?resolved=false&types=BUG&pullRequest=234&id=refinitiv_refinitiv-ui&open=AX-2U44Mz09b4xyDNThu

packages/elements/src/datetime-field/index.ts
goremikins: There are two name declarations, one from super class (with changed properties) and another one without. But I agree this can be done better.

packages/elements/src/datetime-field/index.ts
goremikins: tbh, I think all input controls should have it as null, as input itself is always focusable

packages/elements/src/datetime-field/index.ts
goremikins: OK

packages/elements/src/datetime-field/types.ts
goremikins: we should, I will add

packages/elements/src/datetime-field/index.ts
goremikins: We need to ignore them, e.g. Shift+Left changes the selection, it should not navigate over the parts. In similar way I excluded other keyboard combinations

packages/elements/src/datetime-field/index.ts
goremikins: Valid point, therefore I replace it on focus only. Otherwise it would be impossible for the user to type the correct date.

goremikins: And from accessibility point of view, placeholders should not be used to describe the field, instead labels should be used

packages/elements/src/datetime-field/index.ts
goremikins: Interesting why does it say this. The syntax is valid and we use it in many places, perhaps something somewhere has changed....

packages/elements/src/datetime-field/index.ts
goremikins: This is probably what SonarClud thinks to which I disagree. Splitting a long, but single-purpose function into multiple makes the code less readable. The developer needs to consider multiple flows and constantly jump between lines; and essentially these functions would exist only within the context of a long function and change with it. The function can be split if it has potential to be either re-used or one may need to override parts of that.

packages/elements/src/datetime-field/index.ts
wsuwt: Would it be better to move this to `FormFieldElement` ?

packages/elements/src/datetime-field/index.ts
wsuwt: FYI, warning is part of `lit-plugin` in VSCode.

packages/elements/src/datetime-field/index.ts
wsuwt: I still think placeholder set by user should be read when element is in focus. Perhaps we should discuss this in the meeting.

packages/elements/src/datetime-field/index.ts
goremikins: too big decision :)

packages/elements/src/datetime-field/index.ts
goremikins: maybe a plugin has a bug

packages/elements/src/datetime-field/index.ts
goremikins: OK, we can discuss

packages/elements/src/datetime-field/index.ts
wsuwt: I guess we can leave it as it is.

packages/elements/src/datetime-field/index.ts
goremikins: as discussed keep default

packages/elements/src/datetime-field/index.ts
goremikins: I moved the comment to a different line, but otherwise I think it is fine as it is, as the purpose of the function is to syncInputValue :)

packages/elements/src/datetime-field/index.ts
goremikins: as discussed, placeholder is removed

packages/elements/src/datetime-field/types.ts
goremikins: 'Home' and 'End' do not fully work. The issue is when the value does not fit the input, so we might end-up with the situation when selected part is outside scrollable area. And I really do not want to calculate scrollLeft based on CSS font size.

wsuwt: Yeah I thought so too. Date time parts are dynamic and it would not benefit much anyway given that maximum number of parts is acceptable. Let's leave it as it is.

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,742 @@
...
+} from './utils.js';
+import { VERSION } from '../version.js';
+import '@refinitiv-ui/phrasebook/locale/en/datetime-field.js';
+
+/**
+ * A form control element for datetime input.
+ *
+ * @fires value-changed - Dispatched when value changes
+ * @fires error-changed - Dispatched when error state changes
+ * @fires icon-click - Dispatched when icon is clicked
```
wsuwt, line 48:
```suggestion
```

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,742 @@
...
+    }
+  }
+  public get timepicker (): boolean {
+    return this._timepicker;
+  }
+
+  private _showSeconds = false;
+  /**
+   * Flag to show seconds time segment in display.
+   * Seconds are automatically shown when `hh:mm:ss` time format is provided as a value.
```
wsuwt: This comment is not true. 

```html
<ef-datetime-field value="06:49:14" lang="en-GB"></ef-datetime-field>
```
will not display value.

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,742 @@
...
+  /**
+   * Returns true if the datetime field has am-pm
+   * @returns hasAmPm
+   */
+  protected get hasAmPm (): boolean {
+    return this.amPm || this.hasAttribute('am-pm');
+  }
+
+  /**
+   * Get the start date for when the value is not set
```
wsuwt:
```suggestion
   * Get today's date as the start date for when the value is not set
```

packages/elements/src/datetime-field/index.ts
```diff
@@ -0,0 +1,742 @@
...
+   * Set maximum date.
+   * This value must follow the `format` and be greater
+   * than or equal to the value of the `min` attribute
+   */
+  @property({ type: String, reflect: true })
+  public max: string | null = null;
+
+  private _timepicker = false;
+  /**
+   * Toggle to display the time picker
```
wsuwt, line 116: Should we also mention that, 
```suggestion
   * Toggle to display the time picker.
   * It is required in order to display time.
```
?

goremikins: I do not think so, the name of attribute is self descriptive without any comments

packages/elements/src/datetime-field/utils.ts
```diff
@@ -0,0 +1,87 @@
+import { DateTimeFormatPart, DateTimeFormatPartTypes, InputSelection } from './types';
+import { Direction } from './constants.js';
+
+const IterablePartTypes: DateTimeFormatPartTypes[] = ['year', 'day', 'month', 'hour', 'minute', 'second', 'weekday', 'dayPeriod', 'fractionalSecond'];
+export const isAllowedPart = (part: DateTimeFormatPart): boolean => IterablePartTypes.includes(part.type);
```
wsuwt: At the moment, this is only used internally on `packages/elements/src/datetime-field/utils.ts` so this might not even need an export.

goremikins: sorry, I did not get you?

wsuwt: I meant that we don't have to export `isAllowedPart` if we do not used it anywhere else.

goremikins: got you, done

packages/elemental-theme/src/custom-elements/ef-datetime-field.less
```diff
@@ -0,0 +1,4 @@
+@import 'element:ef-text-field';
+
+:host {
```
wsuwt: Remove empty rulesets.
```suggestion
```

goremikins: Done

wsuwt (approved):

PR #240, opened 2022-03-15
A11Y pagination

Sakchai-Refinitiv:
## Description
- Pagination has a navigation role by default
- Screen reader can pronounce when the input value changed or submitted
- Hide all buttons from the screen reader
- Support keyboard navigation when pressing key Arrow Up/Down, Home, End
- Change to native input because `FormFieldElement` does not support overriding a role in subclass

## Type of change

## Checklist


Theeraphat-Sorasetsakul:
It seems to be going well. Could you @wsuwt  skim to double check before merge?

Theeraphat-Sorasetsakul (approved):

packages/elements/src/pagination/index.ts
```diff
@@ -315,15 +314,22 @@ export class Pagination extends BasicElement {
    * Getter for display text in the input
    * @returns input text
    */
-  protected get inputText (): string {
+  protected get inputValue (): string {
```
wsuwt, line 317: Is it really input value or just display text or both ? This could be more readable and please update function description.

packages/elements/src/pagination/index.ts
```diff
@@ -574,6 +579,86 @@ export class Pagination extends BasicElement {
     this.notifyValueChange();
   }
 
+  protected hasNextPage (page: number): boolean {
```
wsuwt, line 588: Function comments ? and I think it is unnecessary to extract this as a new function.

packages/elements/src/pagination/index.ts
```diff
@@ -358,11 +364,31 @@ export class Pagination extends BasicElement {
...
+
   /**
    * @override
    */
   protected updated (changedProperties: PropertyValues): void {
     super.updated(changedProperties);
+
+    if (changedProperties.has('value')) {
+      this.input.setAttribute('aria-valuenow', this.internalValue.toString());
+    }
```
wsuwt: I'm a bit confused on `value` is check on `updated` and `max` is check on `update`, is there any difference at all ?

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: It seems a simple function but it is useful for refactoring and remove a duplicate code the pagination in the next sprint

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Updated

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: It is not different; I moved all of them to the `updated` method.

wsuwt:
Looks good to me.

wsuwt (approved):

PR #247, opened 2022-03-21
refactor(label): remove unused variable

wsuwt:
## Description
- remove unused variable and add comments to it.

reference to SonarCloud recommendation https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=BUG&id=refinitiv_refinitiv-ui&open=AX-qmzpaqePuut67JdEz

### DIFF
--- a/packages/elements/src/label/index.ts
+++ b/packages/elements/src/label/index.ts
@@ -175,10 +175,9 @@ export class Label extends BasicElement {
 
   /**
    * Handles any modifications to the internal HTML
-   * @param [mutation=false] is the request from a mutation event?
+   * @param [mutation=false] is the request from a mutation event ? ( reserved for future used )
    * @returns {void}
    */
-  // mutation is reserved for future used
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   protected recalculate (mutation = false): void {
     const oldValue = this.text;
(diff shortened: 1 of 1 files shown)


packages/elements/src/label/index.ts
```diff
@@ -178,8 +178,9 @@ export class Label extends BasicElement {
    * @param [mutation=false] is the request from a mutation event?
    * @returns {void}
    */
+  // mutation is reserved for future used
```
bualoy-napat: Why we need to reserve for the future?

packages/elements/src/label/index.ts
```diff
@@ -178,8 +178,9 @@ export class Label extends BasicElement {
    * @param [mutation=false] is the request from a mutation event?
```
wsuwt:
```suggestion
   * @param [mutation=false] is the request from a mutation event ? ( reserved for future used )
```

packages/elements/src/label/index.ts
wsuwt:
```suggestion
```

bualoy-napat: I have already updated.

packages/elements/src/label/index.ts
bualoy-napat: I have already updated.

bualoy-napat (approved):

PR #256, opened 2022-03-23
Pagination refactoring

Sakchai-Refinitiv:
## Description

- Pagination must not have a role.
- Update method names and comments to make them clear and easy to understand.
- Remove `keypress` event and use `keydown` event instead
- Setting aria attributes should move to the template instead.
- Create type `enum` for direction increment and decrement.
- Fix focusing input does not work when using pagination in the shadow root
- Fix incorrect return type `TranslateDirectiveResult`.

## Type of change
- [x] Refactor
- [x] Bug fixing

## Checklist


packages/elements/src/pagination/index.ts
```diff
@@ -637,27 +610,24 @@ export class Pagination extends BasicElement {
...
 
-    const newInputValue = Number(this.input.value);
-
+    // Handle keyboard shortcuts
     switch (event.key) {
       case 'Enter':
-        this.updatePageInput();
-        this.input.blur();
-        break;
       case 'Tab':
```
goremikins: Why do you need Tab here at all? When you tab away from input, blur event will fire and the page is set anyway?

packages/elements/src/pagination/index.ts
```diff
@@ -666,12 +636,18 @@ export class Pagination extends BasicElement {
...
-        return;
     }
 
-    if (event.key !== 'Tab') {
-      event.preventDefault();
+    /**
+     * Allow input keys (navigation, submit, number).
+     * Do not prevent default for `Tab` key, it should focus on next tabindex as the same native behaviour.
+     */
+    const allowKeys = ['ArrowLeft', 'ArrowRight', 'Backspace', 'Tab'];
```
goremikins: I can see at least two problems with this:
1) It does not support IE11, which has 'Left' and 'Right'
2) You can still enter invalid value, e.g. by Copy/Paste

I would implement beforeinput event to strip down all non-numeric characters. But, for IE11 (which does not support beforeinput) you can either ignore (as technically entering letters instead of numbers is a user error) or on input check the value and strip off all non numeric characters.

packages/translate/src/translate.ts
```diff
@@ -32,7 +32,8 @@ type DecoratorOptions = {
   mode?: 'directive' | 'promise';
 };
 type TranslateFunction = (prototype: BasicElement, name: PropertyKey) => void;
-type TranslateDirective = (key: string, options?: TranslateOptions, translateParams?: TranslateParams) => DirectiveResult<typeof AsyncTranslateDirective>;
+type TranslateDirectiveResult = DirectiveResult<typeof AsyncTranslateDirective>;
+type TranslateDirective = (key: string, options?: TranslateOptions, translateParams?: TranslateParams) => TranslateDirectiveResult
```
goremikins: missing `;` at the end of line

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Yes, your are correct. We do not need `Tab` here. Thanks

packages/translate/src/translate.ts
Sakchai-Refinitiv: Fixed

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: 1. Fixed
2. We can use `onpaste` event to solve this issue, itis very simple.

packages/elements/src/pagination/index.ts
```diff
@@ -20,6 +21,12 @@ import '../text-field/index.js';
 
 import '@refinitiv-ui/phrasebook/locale/en/pagination.js';
 import { translate, Translate } from '@refinitiv-ui/translate';
+import type { TranslateDirectiveResult } from '@refinitiv-ui/translate';
```
goremikins: We have discussed when the use of import type is appropriate. Move it to the main import

packages/elements/src/pagination/index.ts
```diff
@@ -20,6 +21,12 @@ import '../text-field/index.js';
...
+
+export enum Direction {
```
goremikins: Why do you export it?

packages/elements/src/pagination/index.ts
```diff
@@ -367,21 +371,10 @@ export class Pagination extends BasicElement {
...
-        this.input.setAttribute('aria-valuemax', this.max);
-      }
-      else {
-        this.input.removeAttribute('aria-valuemax');
-      }
-    }
+  protected firstUpdated (changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    // Prevent copy/paste to the text input
+    this.input.addEventListener('paste', () => false);
```
goremikins: Why do you prevent it? I can copy/paste numbers

packages/elements/src/pagination/index.ts
```diff
@@ -444,56 +447,54 @@ export class Pagination extends BasicElement {
...
-  }
-
   /**
    * Updates page value depending on direction
    * @param direction page value direction
    * @param event whether the event page-changed should fire
    * @returns {void}
    */
-  private updatePage (direction: 'increment' | 'decrement', event = false): void {
+  private updatePage (direction: Direction, event = false): void {
```
goremikins: event usually describes the Event object. In this particular case it makes sense to use `withEvent` instead

packages/elements/src/pagination/index.ts
```diff
@@ -444,56 +447,54 @@ export class Pagination extends BasicElement {
...
+    }
 
-    if (!limit) {
-      this.value = direction === 'increment' ? (page + 1).toString() : (page - 1).toString();
-      if (event) {
-        this.notifyValueChange();
-      }
+    // Update page and fire event
+    if (newPage !== page) {
+      this.value = String(newPage);
```
goremikins, line 319: `newPage` is a number, no need to cast

packages/elements/src/pagination/index.ts
```diff
@@ -444,56 +447,54 @@ export class Pagination extends BasicElement {
...
+      event && this.notifyValueChange();
     }
   }
 
   /**
-   * Update input value
+   * Update input value,
+   * but the value does not apply to the element until `Enter` key is pressed or `Blur` event is fired
```
goremikins: Update input value.  Do not update pagination actual value
until `Enter` key is pressed or `blur` event is fired

packages/elements/src/pagination/index.ts
```diff
@@ -606,28 +607,15 @@ export class Pagination extends BasicElement {
...
   /**
    * Handles action when input focused change
    * @param event focus change event
    * @returns {void}
    */
-  private onFocusedChanged (): void {
-    this.inputFocused = document.activeElement === this;
-
+  private onFocusedChanged (event: FocusEvent): void {
+    this.inputFocused = event.type === 'focus' && event.target === this.input;
```
goremikins: Is `event.target` always the input itself?

packages/elements/src/pagination/index.ts
```diff
@@ -637,27 +625,23 @@ export class Pagination extends BasicElement {
...
-
+    // Handle keyboard shortcuts
     switch (event.key) {
       case 'Enter':
-        this.updatePageInput();
-        this.input.blur();
-        break;
-      case 'Tab':
-        this.updatePageInput();
+        this.setPage(this.input.value);
```
goremikins: Create a getter for `this.input.value` as well as use https://lit.dev/docs/templates/directives/#ref `ref` directive instead of query.

packages/elements/src/pagination/index.ts
```diff
@@ -713,12 +703,13 @@ export class Pagination extends BasicElement {
           part="input"
```
goremikins, line 561: As discussed, change `status` role as per my previous comments.

goremikins (changes requested):

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: OK, that is an object, not a type.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: I have already checked, no need to export

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: type of `this.value` is a string, it needs to convert before assigning.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Do you mean to change the `spinbutton` role to status `role` on the part input?

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Updated

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Updated

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Fixed

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Yes, it should remove the condition.

packages/elements/src/pagination/index.ts
```diff
@@ -367,21 +370,15 @@ export class Pagination extends BasicElement {
...
-      this.input.setAttribute('aria-valuenow', this.internalValue.toString());
-    }
-
-    if (changedProperties.has('max')) {
-      if (!this.infinitePaginate) {
-        this.input.setAttribute('aria-valuemax', this.max);
+  protected firstUpdated (changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    // Prevent copy/paste no to the text input
+    this.input.addEventListener('paste', (event: ClipboardEvent) => {
```
goremikins: This code probably does not cover text drag&drop?

But anyway, you already do two types of input validation: `onKeyDown` and `paste`. Why not to simply stripe invalid characters as follows:

```
@input=${this.onInputInput}

protected stripeInvalidCharacters (value: string): string {
  return value.replace(/[^\d]/g, '');
}

/**
   * Runs on input element `input` event
   * @param event `input` event
   * @returns {void}
   */
  protected onInputInput (event: InputEvent): void {
    const currentInput = this.inputValue;
    const inputValue = this.stripeInvalidCharacters;
    if (currentInput !== inputValue) {
      this.inputValue = inputValue;
    }
  }

```

packages/elements/src/pagination/index.ts
goremikins: no, re:

- To start with you need either aria-labelledby which links with the label, and yes, the screen reader will pronounce label if used with aria-label on Tab/Enter (at least mine does).
- Second option you do not use aria-labelledby and use role=status with aria-live instead. In that scenario the label will be pronounced as soon as aria-label changes. To make it working correctly you may need to bind it with the focused state of input.
- But when looking at your scenario, you can just use aria-description and do not add status element at all.
- another option is to use aria-valuetext on the input itself, but I do not think it works with all screen readers

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Added

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: **Let me separate this discussion into two A11Y issues.**
1. How to set the `label` for the `input` correctly.
2. How to make screen readers pronounce when submitting page value by pressing `Tab/Enter`.

**Solutions**
1. I found a new solution for the first issue by using a `label` element that has `for` attribute same as native. I have already updated the code for review.
2. For the second issue, I think we must have the `status` element to solve that, if we do not have it that is not good for the screen reader users because it should have a confirmation message after pressing `Tab/Enter` keys, without this, users might be confused.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: I will try, it could be better if group it in one place.

Sakchai-Refinitiv: Implemented, I add the condition to handle the invalid number starting with `0` as well

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: I have already updated code, please review again.

packages/elements/src/pagination/index.ts
```diff
@@ -299,10 +147,17 @@ export class Pagination extends BasicElement {
...
-  @query('#input')
-  private input!: HTMLInputElement;
+  protected inputRef: Ref<HTMLInputElement> = createRef();
+
+  /**
+   * Getter for input element
+   */
+  protected get input (): HTMLInputElement {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return this.inputRef.value!;
```
goremikins: It is not always defined (e.g. if you call the function before `firstUpdate`), thus it should return `HTMLInputElement | null` and check if inputRef is defined. As well, please rename the function to `inputElement` to be consistent with other elements.

packages/elements/src/pagination/index.ts
```diff
@@ -607,27 +446,32 @@ export class Pagination extends BasicElement {
...
    */
-  private onFocusedChanged (): void {
-    this.inputFocused = document.activeElement === this;
+  protected onInputInput (): void {
+    const currentInput = this.input.value;
+    const inputValue = this.input.value.replace(/[^\d]/g, ''); // stripe invalid charactors
 
-    if (!this.inputFocused) {
-      this.updatePageInput();
+    // Page value cannot start with `0`, reset it if found.
```
goremikins, line 480: Why? leading `0` is a valid digit and it does not obstruct entering pages.

packages/elements/src/pagination/index.ts
```diff
@@ -666,12 +506,18 @@ export class Pagination extends BasicElement {
...
-        return;
     }
 
-    if (event.key !== 'Tab') {
-      event.preventDefault();
+    /**
+     * Allow input keys (navigation, submit, number).
+     * Do not prevent default for `Tab` key, it should focus on next tabindex as the same native behaviour.
+     */
+    const allowKeys = ['ArrowLeft', 'Left', 'ArrowRight', 'Right', 'Backspace', 'Tab'];
```
goremikins: You do not need it as `input` event is covered

packages/elements/src/pagination/index.ts
```diff
@@ -712,14 +560,16 @@ export class Pagination extends BasicElement {
           id="input"
```
goremikins, line 560: I think `id` is no longer required

packages/elements/src/pagination/index.ts
goremikins: can you please explain, what do you try to achieve with status and why by just adding `aria-label` on input you cannot satisfy the requirement?

packages/elements/src/pagination/index.ts
```diff
@@ -637,27 +638,23 @@ export class Pagination extends BasicElement {
...
   private onKeyDown (event: KeyboardEvent): void {
+
     if (event.defaultPrevented) {
       return;
     }
 
-    const newInputValue = Number(this.input.value);
-
+    // Handle keyboard shortcuts
     switch (event.key) {
```
goremikins, line 501: re: when using keyboard navigation, please reselect the value.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: The valid value of pagination must start at 1

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: The `label` element has attribute `for` which using this `id`

packages/elements/src/pagination/index.ts
goremikins: OK, than do not set the value to empty string, but just strip leading 0

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Fixed, Thanks

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Yes, the strip leading zero is better, Thanks.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Removed

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Imagine, when a blind person uses a screen reader when they open a web browser and press `Tab` to focus on the `ef-pagination`, and then the screen reader will get the `label` of `ef-pagination` and pronounce it.  

This is normal related behavior of `label` and `input element` but this is not enough for `ef-pagination` why?

When using `ef-pagination` with a screen reader, the user (blind person) can press the arrow up/down, or type to update the value of the input, but these actions do not update the `page` of `ef-pagination` until pressed the `Enter/Tab` key.

The question is what will happen with the screen reader if the user (blind person) types a `new page number` or uses the `arrow up/down` key and presses `Enter`? The answer is nothing happened.

This is unfriendly for blind users because they will not know what will happen after submitting the change and they will not know if the page has already changed or not, which makes them confused.

That is the reason why `aria-label` is not enough and we should use the `role=status` with `aria-live` to make `ef-pagination` pronounce the `page number` after the page is already changed from submitted, this is the way it should be.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Can you explain more what is the issue? I do not get what you mean.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: If you have any questions please let me know, I will call you.

packages/elements/src/pagination/index.ts
goremikins: Take a look here: https://ui.refinitiv.com/elements/time-picker, and try to ArrowUp/ArrowDown the field. The input value stays selected. If it is too difficult to implement for pagination, I am fine with as it is.

packages/elements/src/pagination/index.ts
goremikins: I disagree. I have set the call for tomorrow.

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: Done

packages/elements/src/pagination/index.ts
Sakchai-Refinitiv: image

@goremikins 
According to the meeting, we already decided to go with `Option 4` (In the image above) but I have already checked on all browsers, to make it works it must add more logic to the `ef-pagination` that might not be good because it is more complex.

@wsuwt and I think the same thing about `Option 1` it could be the best solution because it is simple and we do not add any logic to handle the `A11Y`, for the only one issue on `JAWS` (pronounce after pressing an `Enter` key) I think it is acceptable.

I have already pushed both options to the PR, please review them and let me know which one you prefer. 
Thanks, 😀

Option 4: https://github.com/Refinitiv/refinitiv-ui/pull/256/commits/79725c11617e023ce2a431ef5fcc1079a0963f2d
Option 1: https://github.com/Refinitiv/refinitiv-ui/pull/256/commits/2282fa80f240fc11da50e8e2a85a5bb1af17256d

goremikins: Let's go with option 1 than

goremikins: Please add the new re-usable mixin in elemental-theme, called ariaInvisible, which should have the following css:
```
clip: rect(1px, 1px, 1px, 1px); // IE11
clip-path: inset(50%);
height: 1px;
width: 1px;
margin: -1px;
overflow: hidden;
padding: 0;
position: absolute;
```
This block is derived from Bootstrap and was suggest by Caleb to hide content.

Sakchai-Refinitiv: @goremikins I have already added the mixing to the shared-styles directory, please review it again.

packages/elements/src/pagination/index.ts
```diff
@@ -637,40 +649,40 @@ export class Pagination extends BasicElement {
...
         this.updateInputValue(1);
+        this.inputElement?.select();
+        event.preventDefault();
         break;
       case 'End':
         this.hasLastPage() && this.updateInputValue(this.internalMax);
+        this.inputElement?.select();
+        event.preventDefault();
         break;
       default:
```
goremikins: Instead of having empty default you can just put `// no default`

goremikins (approved):

PR #257, opened 2022-03-25
ci: triggers code scan on push on any branch

wsuwt:
## Description

Triggers code scan on push on any branch


wsuwt:
cc. @dtanp push event onto any branch should be fine.

PR #270, opened 2022-04-01
refactor(heatmap): simplify get maximum text length logic

wsuwt:
## Description

Simplify get maximum text length logic.

## Type of change

- [x] Refactor

## Checklist


packages/elements/src/heatmap/helpers/text.ts
```diff
@@ -95,22 +95,12 @@ const getMaximumTextWidth = (ctx: CanvasRenderingContext2D, cells: HeatmapCell[]
...
-  for (let i = 0; i < labels.length; i++) {
-    if (maxLengthText.length >= labels[i].length) {
-      maxLengthText = maxLengthText;
-    }
-    else {
-      maxLengthText = labels[i];
-    }
-  }
-  return maxLengthText;
+  return labels.reduce((previousText, currentText) => previousText.length > currentText.length ? previousText : currentText);
```
dtanp-rft: I'd say rename var from `xxText` to `xxLabel` would be better.
And `@returns Text with maximum length` -> `@returns Longest label in the array`

Btw, method name could be better :)

wsuwt: Agreed and fixed.

packages/elements/src/heatmap/helpers/text.ts
```diff
@@ -95,22 +95,12 @@ const getMaximumTextWidth = (ctx: CanvasRenderingContext2D, cells: HeatmapCell[]
...
-      maxLengthText = labels[i];
-    }
-  }
-  return maxLengthText;
+const getMaximumLabelWidth = (labels: string[]): string => {
+  return labels.reduce((previousLabel, currentLabel) => previousLabel.length > currentLabel.length ? previousLabel : currentLabel);
 };
 
-export { getResponsiveFontSize, getMaximumTextWidth, getMaximumLabelTextWidth, MIN_FONT_SIZE, MAX_FONT_SIZE };
+export { getResponsiveFontSize, getMaximumTextWidth, getMaximumLabelWidth, MIN_FONT_SIZE, MAX_FONT_SIZE };
```
dtanp-rft: What I actually don't like about the method name is that it ends with `Width` so I would assume that this method returns width but actually, it returns the longest label.

You can see that `getMaximumTextWidth` and `getMaximumLabelWidth` have similar name while the first one is returning width (which makes sense for its name).

wsuwt: That's true. I'll go for `getLongestLabel()` instead.

wsuwt: Actually we might not even need to extract this method at all. I can just place this line onto the code itself.

dtanp-rft: I like that :) I want to push one line comment and I will approve this PR

dtanp-rft (approved):

PR #272, opened 2022-04-07
fix(list): clicking on item should focus on list itself

wsuwt:
## Description

When an item is clicked, it gets focused. This makes tabbing sequence wrong. Instead, focus should only remain at `list`.

So I propose this re-focus solution. However, this solution might be a problem when user wants to custom render and add certain event listener when an item is clicked.

## Type of change

## Checklist


packages/elements/src/tree/helpers/renderer.ts
```diff
@@ -4,7 +4,7 @@ import type { TreeDataItem } from './types';
 import { TreeManager, TreeManagerMode, CheckedState } from '../managers/tree-manager.js';
 import { Renderer } from '../../list/renderer.js';
 import { getItemId } from '../../list/helpers/item-id.js';
-import '../elements/tree-item.js';
+import { TreeItem } from '../elements/tree-item.js';
```
Sakchai-Refinitiv: use `import type` instead

packages/elements/src/tree/__test__/tree.test.js
```diff
@@ -267,7 +267,6 @@ describe('tree/Tree', () => {
       event = await oneEvent(el, 'expanded-changed');
       expect(event.detail.value, 'Group should be expanded').to.be.true;
       expect(event.detail.item, 'Item should be the same as the original').to.equal(nestedData[0]);
-      el.dispatchEvent(keyArrowDown);
```
Sakchai-Refinitiv, line 270: Why removed this event?

wsuwt: image

So if you look at the testing data, you will see that only first item is expandable. If arrow down is pressed, it will move into item that has no children so cannot expand. It was a false test case.

packages/elements/src/tree/helpers/renderer.ts
wsuwt: Updated.

packages/elements/src/tree/__test__/tree.test.js
Sakchai-Refinitiv: OK

Sakchai-Refinitiv (dismissed):

packages/elements/src/list/__test__/list.test.js
```diff
@@ -279,5 +279,21 @@ describe('list/List', () => {
...
+  it('Should have focus state remain at host when tapping in an item', async () => {
+    const el = await fixture('<ef-list></ef-list>');
+    el.data = data;
+    await elementUpdated(el);
+
+    const firstElement = el.firstElementChild;
+
+    await triggerFocusFor(el);
+
+    setTimeout(() => firstElement.dispatchEvent(new Event('tap')));
```
TremayneChrist: You should be able to use `firstElement.click()` - this should fire tap event.

packages/elements/src/tree/elements/tree-item.ts
```diff
@@ -220,7 +225,7 @@ export class TreeItem<T extends TreeDataItem = TreeDataItem> extends ControlElem
    * @param changedProperties changed properties
    * @returns {void}
    */
-  protected willUpdate (changedProperties: PropertyValues): void {
```
TremayneChrist: ?

packages/elements/src/tree/helpers/renderer.ts
```diff
@@ -24,7 +24,7 @@ export class TreeRenderer extends Renderer {
     let currentMode: TreeManagerMode;
     let currentComposer: CollectionComposer<TreeDataItem>;
 
-    super((item: TreeDataItem, composer: CollectionComposer<TreeDataItem>, element = document.createElement('ef-tree-item')): HTMLElement => {
+    super((item: TreeDataItem, composer: CollectionComposer<TreeDataItem>, element: TreeItem = document.createElement('ef-tree-item')): HTMLElement => {
```
TremayneChrist: You shouldn't need to add the type here.

packages/elements/src/tree/elements/tree-item.ts
wsuwt: My bad. Ran a clean install on it and it was fetching an older version of lit's reactive-element which still have `willUpdate` as public. Fixed.

packages/elements/src/tree/helpers/renderer.ts
wsuwt: Right because createElement should already return correct type from `HTMLElementTagNameMap`. Fixed.

packages/elements/src/list/__test__/list.test.js
wsuwt: Updated.

packages/elements/src/list/elements/list.ts
```diff
@@ -0,0 +1,687 @@
...
+   * @param item Data item to scroll to
+   * @returns {void}
+   */
+  public scrollToItem (item: T): void {
+    const element = this.elementFromItem(item);
+    if (element) {
+      const min = this.scrollTop;
+      const max = this.scrollTop + this.clientHeight - element.offsetHeight;
+      const pos = element.offsetTop;
+      pos > max ? this.scrollTop = element.offsetTop - this.clientHeight + element.offsetHeight
```
TremayneChrist: We need to set a variable here rather than updating the scrollTop value in the switch. Then use the variable to update the scrollTop position. This should fix the sonarcloud issue.

TremayneChrist:
```suggestion
      const scrollTop = pos > max ? element.offsetTop - this.clientHeight + element.offsetHeight : pos < min ? element.offsetTop : 0;
      this.scrollTop = scrollTop;
```

Something like this, or better, if you can clean up the calulations. Just needs a tidy up :)

wsuwt: Updated. I just found out that this method never really work as a standalone public method. It uses an object as a key in `elementMap` when passing a new item object to `scrollToItem`, it would be a different reference value.

TremayneChrist:
Just needs the test coverage to be fixed

dtanp-rft (approved):

PR #278, opened 2022-04-12
BrowserStack

Sakchai-Refinitiv:
### Create a new GitHub workflows for the unit testing on BrowserStack

**Changes:**
- Run test on BrowserStack on a pull request and new schedule a nightly test workflow
  - Chrome latest and previous version on Windows 11
  - Firefox latest and previous version on Windows 11
  - Edge latest and previous version on Windows 11
- Print output in minimal mode when testing on CI (also fix snapshot silent failed)
- New CLI options for the test command 
  - Option `--browserstack` using for setup and test browsers on BrowserStack
  - Option `--output` using for print output to the console
- Fix bugs & incorrect unit tests.

Example CLI commands:
```bash
npm run test button -- --browserstack chrome firefox_previous ios android

```

❗️ Please remove `chrome` and `firefox` from `GitHub Checks`  before merge this PR.
 
**Notes:**
- Testing the latest version and the previous version (latest version - 1) on the three major browsers, takes more time around 15 minutes, I think it should be in the daily/nightly test instead of the pull request.
- Found many test cases fail on Safari, which should be fixed if we want to enable unit test on Safari (https://github.com/Refinitiv/refinitiv-ui/runs/6082266893)


karma.config.js
```diff
@@ -197,6 +225,81 @@ if (argv.includeCoverage) {
   plugins.push(require.resolve('karma-coverage-istanbul-reporter'));
 }
 
+// Create BrowserStack config when browsers CLI `browsers` param has `browserstack`
+if (argv.browserstack.length && !argv.watch) {
```
Theeraphat-Sorasetsakul: Should we support to run without specific any browsers? For example exec command `npm run test -- --browserstack` then it will auto pick default browsers as same as ci being.

Sakchai-Refinitiv: Updated

Theeraphat-Sorasetsakul (dismissed):

Sarin-Udompanish (approved):

PR #283, opened 2022-04-25
fix(polyfills): add missing polyfills for IE11

goremikins:
## Description
Add missing polyfills required for IE11.

## Type of change

## Checklist


dtanp-rft (commented):
LGTM

dtanp-rft:
@goremikins I remember that you are asking the developer to try this solution, please post the feedback here. If he never come back to you, we will rely on our test result solely.

@bualoy-napat Please re-test this solution on IE11 with your fix about removing /lib/ from theme file. I will send you webpack config offline, in case you don't have a working one yet.

goremikins:
@dtanp-rft , I did my test with Vanilla JS, as well tested `slider` as per original question. Ruslan should test for Angular

dtanp-rft:
I'm sure that it works before with v5. Do you know if we need to fix in v5 as well or this is only for v6?

bualoy-napat:
@dtanp-rft I will re-test with the config that you gave me in the chat.

dtanp-rft:
Please ignore my question above. I meant v4 and Aleks confirmed that it doesn't have the problem.

goremikins:
@dtanp-rft , it should be ready for review. We do not need full set of polyfills anymore

dtanp-rft:
to @goremikins 
### Content on polyfills/readme.md

Sections before `Webpack 5` doesn't look practical to me. It doesn't help me to understand what I need to do if I don't use Webpack 5. 

We could simplify the page like below.

Explain that because EF is shipped with ES2020 syntax and use some features that only available on ES2020. This is not specific to EF but this is applied to any other dependencies that use in your application as well if they ship the package ES2020.

To support old browsers should 2 key things; Polyfills missing ES2020 features on the browsers and Transpile JavaScript syntax to ES5 or ES6. Polyfills that we recommend is core-js for polyfills and babel to transpile the syntax.

Once, the application has been polyfills and transpiled correctly, it's possible that the browsers doesn't support Web Component because Web Component was implemented in the browsers not too long ago. This will require additional polyfills, `@refinitiv-ui/polyfills`

**Webpack4**
_provide mainly instructions with very brief explanation if required_

**Webpack5**
_provide mainly instructions with very brief explanation if required_

_*** Potentially, it could be just work out-of-the-box with other modern bundler. I don't think we can provide instruction for only webpack 4 and webpack 5 which very popular in our dev community. ***_

### Section about Polyfills for format.js
This gives me some thought that we may structure our polyfills a bit by separating the file to `polyfills/webcomponent` and `polyfills/formatjs` and `polyfills/get-attribute-names`. What do you think? `@refinitiv-ui/polyfills` can include all of this.

### Remove `polyfills/minimal` is going to break backward compatible
To be fair, it would be OK for v6. Basically, from v6, we only provide polyfill for webcomponent and for core-js, developers can simply manage it in the app.

dtanp-rft:
to @bualoy-napat @Theeraphat-Sorasetsakul @iamatomath 

While I'm trying to finalize documentation for Polyfill, can you please help to test if the steps in this PR is correct? (I am thinking that it's true that we no longer need to import core-js file but it would miss step to install core-js).

- Let's use vanilla js with webpack 4 and webpack 5 to validate this. I don't think we need Angular as that will make it harder to test
- We should verify with IE11, Eikon Desktop and Workspace (because the latter two are 99% of our users)
- For Eikon Desktop + Workspace, let's verify first without polyfills, will it work? (instruction is in https://docs-restructure.refinitiv-ui.pages.dev/tutorials/vanilla-js)
- For Eikon Desktop, let's test it with latest version which is using Chromium 89
- For Workspace, let's test it with Workspace v1.14 (release on May 2021) and v1.16 (should be released to Prod soon)

Here is useful links.
- Download Workspace packages https://workspace.int.refinitiv.com/
- Detail of chromium that each Workspace uses https://confluence.refinitiv.com/display/EF/Electron+Release+Notes

We want to verify EF6 with following configuration (test app should, at least, have Button, Slider and Pagination)

|   | Webpack4  | Webpack5  | Webpack4 w/ Polyfill  | Webpack5 w/ Polyfill   |
|---|---|---|---|---|
| IE11 | N/A  | N/A  |   |   |
| Eikon Desktop 4.0.55 (CEF89)  |   |   |   |   |
| Workspace Desktop v1.14 (Chromium 87)  |   |   |   |   |
| Workspace Desktop v1.16 (Chromium 89) |   |   |   |   |

dtanp-rft (changes requested):
Request to revise content on polyfills/readmd.md

goremikins:
@dtanp-rft , I disagree that these sections do not look practical. I split the document into two parts:

### Explain what is required to make EF working
There are three main topics:

1. Support of `ES2020` (including modules) - which is done via core-js transpilation. For up-to-date `babel/preset-env` this should be resolved automatically. However, the developer may include core-js polyfills manually into the application, and I provide the list of required ones.
2. Support of Web Components - we provide a polyfill for that
3. Support of Localisation - the developer must include formatj.js polyfills

### Common bundle configurations

1. All of them require babel core and babel preset-env
2. All of them require IE11 to be included into .browerconfig

Provide common configurations:

1. WebPack 5/4 should be the same, both required babel-loader and assuming both have latest core-js and Babel used.
2. I *agree* we need more documentation on other bundlers/Angular, but this was outside my scope of knowledge.


Yeah, it breaks backwards compatibility, but I believe for v6 it is fine.

dtanp-rft (approved):

PR #286, opened 2022-04-28
refactor(slider): improve code readability and adjust code style

wsuwt:
## Description

Improve code readability and adjust code style.

## Type of change

- [x] Code refactoring

## Checklist

### DIFF
--- a/packages/elements/src/slider/index.ts
+++ b/packages/elements/src/slider/index.ts
@@ -11,6 +11,7 @@ import {
 import { customElement } from '@refinitiv-ui/core/decorators/custom-element.js';
 import { property } from '@refinitiv-ui/core/decorators/property.js';
 import { query } from '@refinitiv-ui/core/decorators/query.js';
+import { createRef, ref, Ref } from '@refinitiv-ui/core/directives/ref.js';
 import { styleMap } from '@refinitiv-ui/core/directives/style-map.js';
 import { VERSION } from '../version.js';
 import '../number-field/index.js';
@@ -130,7 +131,7 @@ export class Slider extends ControlElement {
   /**
    * An array of slider thumbs
    */
-  private thumbs: Array<HTMLDivElement> = [];
+  private thumbs: HTMLDivElement[] = [];
 
   /**
    * Whether if the thumb is being drag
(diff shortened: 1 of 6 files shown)


packages/elements/src/slider/index.ts
```diff
@@ -263,104 +334,123 @@ export class Slider extends ControlElement {
    * Getter for slider part.
    */
   @query('[part="slider"]')
```
Sakchai-Refinitiv: All `query` decorator should be changed to `ref` decorator instead.

packages/elements/src/slider/index.ts
```diff
@@ -88,12 +48,95 @@ export class Slider extends ControlElement {
...
+
+  /**
+   * Current active slider thumb
+   */
+  private activeThumb!: HTMLDivElement;
+
+  /**
+   * An array of slider thumbs
+   */
+  private thumbs: Array<HTMLDivElement> = [];
```
goremikins: private thumbs: HTMLDivElement[] = [];

packages/elements/src/slider/index.ts
```diff
@@ -88,12 +48,95 @@ export class Slider extends ControlElement {
     return VERSION;
   }
 
-  private activeThumb!: HTMLElement;
+  /**
+   * Define styles in a tagged template literal, using the css tag function.
+   * @returns CSS template
+   */
+  static get styles (): CSSResultGroup {
```
Sakchai-Refinitiv, line 56: This method should be moved to the sibling `render` method. Ref: https://confluence.refinitiv.com/display/ELF/Code+Styling+Guide

goremikins: It should not. static methods come at the top. We need to change the guidelines

packages/elements/src/slider/utils.ts
```diff
@@ -0,0 +1,60 @@
+/**
+  * Set prevent default action and stop bubbles event
+  * @param event event mouse or touch
+  * @returns {void}
+  */
+const preventDefault = function (event: Event): Event {
```
goremikins: Please use arrow functions:

`const preventDefault = (event: Event): Event => {`

goremikins:
I think for light refactor it is good enough

packages/elements/src/slider/index.ts
wsuwt: I changed all `@query` except inputs which are dynamically render and involve a lot of complexity and I think is beyond the current scope.

wsuwt: I'll save this for future refactoring.

packages/elements/src/slider/index.ts
wsuwt: I agree that static methods should be on top.

goremikins (approved):

PR #289, opened 2022-04-29
fix(core): check if ShadyCSS is defined before calling the statement

wsuwt:
## Description 

Fixes core crashing when ShadyCSS is undefined.

### DIFF
--- a/packages/core/src/registries/NativeStyleRegistry.ts
+++ b/packages/core/src/registries/NativeStyleRegistry.ts
@@ -30,9 +30,12 @@ export abstract class NativeStyleRegistry {
     style.setAttribute('scope', name);
     style.textContent = css;
     childRef ? head.insertBefore(style, childRef) : head.appendChild(style);
-    // Register style tag with ShadyCSS
-    // to support CSS variables in legacy browsers
-    ShadyCSS?.CustomStyleInterface?.addCustomStyle(style);
+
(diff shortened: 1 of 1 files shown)


packages/core/src/registries/NativeStyleRegistry.ts
```diff
@@ -30,9 +30,12 @@ export abstract class NativeStyleRegistry {
...
     style.textContent = css;
     childRef ? head.insertBefore(style, childRef) : head.appendChild(style);
-    // Register style tag with ShadyCSS
-    // to support CSS variables in legacy browsers
-    ShadyCSS?.CustomStyleInterface?.addCustomStyle(style);
+
+    if (ShadyCSS) {
+      // Register style tag with ShadyCSS
+      // to support CSS variables in legacy browsers
+      ShadyCSS?.CustomStyleInterface?.addCustomStyle(style);
```
goremikins: `ShadyCSS.CustomStyleInterface?.addCustomStyle(style); ` - u have already checked for ShadyCSS existence

wsuwt: Oops my bad.

wsuwt: Fixed.

goremikins (approved):

PR #299, opened 2022-05-10
feat(slider): add arrow key navigation support and accessible aria attributes

wsuwt:
## Description

- Add `role="slider"` to thumb
- Make thumb focusable
- Add thumb focus style based on `active` attribute
- Thumb can be used to increase or decrease the slider value
- When value changes by arrow keys, screen reader announce the current value

## Type of change

## Checklist


packages/elemental-theme/src/custom-elements/ef-slider.less
```diff
@@ -50,6 +50,11 @@
 		background: var(--thumb-color, @slider-thumb-color);
 	}
 
+  &[focused=visible] [part=thumb-container][active] [part=thumb] {
+    outline: 1px solid @scheme-color-primary;
+    outline-offset: 2px;
```
goremikins, line 55: This does not work in IE11. Does Figma have styles for that?

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,99 @@ export class Slider extends ControlElement {
...
+   * Handles key down event on thumbs
+   * @param event Keyboard event
+   * @returns {void}
    */
-  private calculatePercentage (value: number): number {
-    const percentage = Math.abs((((value || 0) - this.minNumber) / (this.maxNumber - this.minNumber)) * 100);
-    if (percentage > 100) {
-      return 100;
+  private onKeyDown (event: KeyboardEvent): void {
+    if (this.readonly || this.minNumber === this.maxNumber) {
```
goremikins: `this.minNumber >= this.maxNumber`

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,99 @@ export class Slider extends ControlElement {
...
+   * @param multiplier defaults to 100
+   * @returns percentage as a fraction of 100
+   */
+  private calculatePercentage (value: number, multiplier = 100): number {
+    const percentage = Math.abs((((value || 0) - this.minNumber) / (this.maxNumber - this.minNumber)) * multiplier);
+
+    if (percentage > multiplier) {
+      return multiplier;
     }
     else if (percentage < 0) {
```
goremikins: It cannot be `<0` as `Math.abs` is used

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,99 @@ export class Slider extends ControlElement {
...
+  }
+
+  /**
+   * Get percentage value from decimal fraction number
+   * @param value decimal fraction value
+   * @param multiplier defaults to 100
+   * @returns percentage as a fraction of 100
+   */
+  private calculatePercentage (value: number, multiplier = 100): number {
+    const percentage = Math.abs((((value || 0) - this.minNumber) / (this.maxNumber - this.minNumber)) * multiplier);
```
goremikins: `value` is always a number, no need to compare with `0`

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,99 @@ export class Slider extends ControlElement {
...
+  private onApplyStep (direction: Direction): void {
+    const currentPercentageValue = this.calculatePercentage(this.valueNumber, 1);
+    const percentageStep = this.calculatePercentage(this.minNumber + this.stepRange, 1);
+
+    const percentageValue = direction === Direction.Up ? currentPercentageValue + percentageStep : currentPercentageValue - percentageStep;
+    const nearestPossibleValue = this.getNearestPossibleValue(percentageValue);
+
+    const value = this.getValueFromPercentage(nearestPossibleValue);
+
+    this.value = this.format(value);
```
goremikins: Does the value always change? If it does not, you do not need changed event. As well I do not know why it is called `dispatchDataChangedEvent`, but not `dispatchValueChangedEvent`

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,99 @@ export class Slider extends ControlElement {
...
+  }
+
+  /**
+   * Increase or decrease value depending on direction
+   * And fires value change event
+   * @param direction Up or Down
+   * @returns {void}
+   */
+  private onApplyStep (direction: Direction): void {
+    const currentPercentageValue = this.calculatePercentage(this.valueNumber, 1);
```
goremikins: These two lines need some comment, or the function `calculatePercentage` needs to be called better

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,99 @@ export class Slider extends ControlElement {
...
+    const percentageStep = this.calculatePercentage(this.minNumber + this.stepRange, 1);
+
+    const percentageValue = direction === Direction.Up ? currentPercentageValue + percentageStep : currentPercentageValue - percentageStep;
+    const nearestPossibleValue = this.getNearestPossibleValue(percentageValue);
```
goremikins: While not a big issue, I find it confusing that some code uses thumbPosition, some other code uses percentageValue. Difficult to read and understand the code

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,99 @@ export class Slider extends ControlElement {
...
-   * Input number event blur and notify property
-   * @param event event blur input number field
+   * Adds active attribute used in styling
+   * @param event focus event
    * @returns {void}
    */
-  private onBlur (event: FocusEvent): void {
+  private onThumbFocus (event: FocusEvent): void {
+    const thumb = event.target as HTMLDivElement;
+    thumb.setAttribute('active', '');
```
goremikins: Here and in other places use the `@state` instead. It improves code visibility and ensures that all internal element updates are happening at one place during `render`.

packages/elements/src/slider/index.ts
```diff
@@ -706,20 +770,18 @@ export class Slider extends ControlElement {
       else if (distanceFrom > distanceTo) {
         this.activeThumb = this.thumbs[1];
       }
-      this.thumbs.forEach((el: HTMLElement) => {
-        el.style.zIndex = '5';
+
+      this.thumbs.forEach((thumb: HTMLElement) => {
+        thumb.style.zIndex = '5';
```
goremikins: Why `5`? :)

packages/elements/src/slider/index.ts
```diff
@@ -1197,17 +1270,17 @@ export class Slider extends ControlElement {
...
     return html`
       <ef-number-field
-        @blur=${this.onBlur}
-        @keydown=${this.onKeydown}
+        @blur=${this.onNumberFieldBlur}
+        @keydown=${this.onNumberFieldKeyDown}
         part="input"
         name="${name}"
         no-spinner
+        aria-label="${name} input"
```
goremikins: `input`? Does it need to be translated?

packages/elements/src/slider/index.ts
```diff
@@ -1197,17 +1270,17 @@ export class Slider extends ControlElement {
...
         .value="${value}"
         min="${this.min}"
         max="${this.max}"
         step="${this.step}"
-        ?disabled="${this.disabled}"
```
goremikins, line 1209: Why did you remove `disabled`?

packages/elements/src/slider/utils.ts
```diff
@@ -39,22 +39,23 @@ const countDecimalPlace = (value: string | number): number => {
...
 /**
- * Check if the value is a valid number and convert to string
+ * Check if the value is a valid number
  * If not, return fallback value instead
  * @param value value for checking
  * @param fallbackValue fallback value when value is not number;
  * @returns string of valid number
  */
-const validateNumber = (value: number, fallbackValue: number | string): string => {
+const validateNumber = (value: number, fallbackValue: number): number => {
```
goremikins: based on function name, I would expect it to return `NaN` if passed value is invalid number or not a number. The more correct would be:
```
const toNumber = (value: unknown, fallbackValue = 0): number => {
   if (typeof value === 'number' && !isNaN(value)) {
     return value;
   }
   return fallbackValue;
}
```

packages/elements/src/slider/index.ts
wsuwt: It is called `dispatchDataChangedEvent` because it supports data like `value`, `from` and `to` in one go and while it is not very elegant it checks for data change inside the function.

packages/elements/src/slider/index.ts
wsuwt: Deleted it by mistake. Added back.

packages/elemental-theme/src/custom-elements/ef-slider.less
wsuwt: There is no thumb focus design in Figma. So I decided to go in consistency with checkbox outline style.

packages/elements/src/slider/index.ts
wsuwt: Fixed.

packages/elements/src/slider/index.ts
wsuwt: Fixed.

packages/elements/src/slider/index.ts
wsuwt: Fixed.

packages/elements/src/slider/index.ts
wsuwt: Agreed. I replaced the term `percentageValue` with `thumbPosition` for consistency.

packages/elements/src/slider/index.ts
wsuwt: There must be a few `@state` to support multiple thumbs. I find this way easier to read as it is hooked on each thumb automatically.

packages/elements/src/slider/index.ts
wsuwt: From looking at the code, I think it is just to make thumb elevated above the `part=track-fill` any number more than `2` should be fine. I'll change to `3` for `thumbs` and `4` for `activeThumb`.

packages/elements/src/slider/index.ts
wsuwt: Yes maybe both `{from, to, value} input`, but as a temporal solution to labelling inputs and thumbs I'm not sure if it is worth. Or otherwise we shouldn't label them at all. What do you think ?

packages/elements/src/slider/utils.ts
wsuwt: Actually we may not need this function at all. All the getters already provided fallback values. Please see https://github.com/Refinitiv/refinitiv-ui/pull/299/commits/6633de30c30c4c28f47b171454346b40864a8450

packages/elemental-theme/src/custom-elements/ef-slider.less
goremikins: @Mavis-Alaneme , could you please create a ticket for UX/Design for this?

packages/elements/src/slider/__test__/slider.event.test.js
```diff
@@ -0,0 +1,1465 @@
...
+  oneEvent,
+} from '@refinitiv-ui/test-helpers';
+
+import '@refinitiv-ui/elements/slider';
+import '@refinitiv-ui/elemental-theme/light/ef-slider';
+
+import { tabSliderPosition, calculateValue } from './utils';
+
+describe('slider/Events', () => {
+  describe('Events', () => {
```
goremikins: I did not find any other `describe` in this file. You can either split other test and give describe a reasonable name or remove it.

packages/elements/src/slider/__test__/slider.event.test.js
```diff
@@ -0,0 +1,1465 @@
...
+    let el;
+
+    beforeEach(async () => {
+      el = await fixture('<ef-slider></ef-slider>');
+    });
+
+    it('Drag thumb slider on desktop', async () => {
+      await elementUpdated(el);
+      setTimeout(() =>
+        el.sliderRef.value.dispatchEvent(new MouseEvent('mousedown'))
```
goremikins: I can see that you use a lot of `el.sliderRef.value` private accessors. I understand this is unavoidable, but it would be better to create a function, e.g.
```
const getSliderElement = el => el.sliderRef.value;
```
So if the private implementation of the code changes you do not need to change tests. I would apply this pattern for *all* private accessors in tests.

packages/elements/src/slider/__test__/slider.keyboard.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+    describe('Default', () => {
+      const value = 5;
+      const step = 1;
+
+      beforeEach(async () => {
+        el = await fixture(
+          `<ef-slider value="${value}"  step="${step}"></ef-slider>`
+        );
+        thumb = el.shadowRoot.querySelector('[part=thumb-container]');
```
goremikins: As per above container, create a getter function

packages/elements/src/slider/__test__/slider.number-field.test.js
```diff
@@ -0,0 +1,113 @@
...
+    beforeEach(async () => {
+      el = await fixture('<ef-slider></ef-slider>');
+    });
+
+    it('input number field has set value 40 it should be slider value has correct ', async () => {
+      el.showInputField = '';
+      await elementUpdated(el);
+      expect(el.hasAttribute('show-input-field')).to.equal(true);
+
+      let input = el.shadowRoot.querySelector('ef-number-field[name=value]');
```
goremikins: As per above comment, create a named function

packages/elements/src/slider/index.ts
goremikins: Confusing, but OK :)

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +581,97 @@ export class Slider extends ControlElement {
...
+
+  /**
+   * Increase or decrease value depending on direction
+   * Then fires value change event
+   * @param direction Up or Down
+   * @returns {void}
+   */
+  private onApplyStep (direction: Direction): void {
+    // Get current thumb position and step in percentage format
+    const thumbPosition = this.calculatePercentage(this.valueNumber, 1);
```
goremikins: I think it makes more sense to call this function as `calculatePosition`. At least my assumption of calculating percents is `value / total * 100`, but the function clearly does more than that :)

packages/elements/src/slider/index.ts
goremikins: I know you need multiple `@state`, which is fine. We should try to follow Lit guidelines: https://lit.dev/docs/components/rendering/, which recommends to "avoid making DOM updates outside of render()" and "Lit batches updates to maximize performance and efficiency. Setting multiple properties at once triggers only one update, performed asynchronously at microtask timing.".

packages/elements/src/slider/index.ts
```diff
@@ -706,20 +767,19 @@ export class Slider extends ControlElement {
...
 
-      // Set focus to input when drag.
-      if (this.isShowInputField) {
-        this.toggleFocusField(this.activeThumb.getAttribute('name') || '', true);
-      }
+      // Reset z-index of the other thumb and set activeThumb to the top
+      this.thumbs.forEach((thumb: HTMLElement) => {
+        thumb.style.zIndex = '3';
+      });
+      this.activeThumb.style.zIndex = '4';
```
goremikins: I think this should be applied in render function

packages/elements/src/slider/index.ts
goremikins: I am fine if we do not label them. This is really odd to have these inputs anyway. I think they were added in v1 to satisfy someone.

wsuwt: Yeah I agree, it is a strange design choice. I'm going to remove label for now.

packages/elements/src/slider/__test__/slider.number-field.test.js
wsuwt: Done.

packages/elements/src/slider/__test__/slider.keyboard.test.js
wsuwt: Done.

packages/elements/src/slider/__test__/slider.event.test.js
wsuwt: Done.

packages/elements/src/slider/__test__/slider.event.test.js
wsuwt: Done.

packages/elements/src/slider/index.ts
wsuwt: Thanks. I like the name. Updated.

packages/elements/src/slider/index.ts
wsuwt: Updated method name to `calculatePosition` instead.

bualoy-napat:
I already tested with JAWS and Chrome, Edge, Firefox. The outcome was correct as we expected.

packages/elements/src/slider/index.ts
```diff
@@ -359,6 +353,24 @@ export class Slider extends ControlElement {
...
+   * Array of slider thumbs
+   */
+  @queryAll('[part~=thumb-container]')
+  private thumbs!: HTMLDivElement[];
+
+  /**
+   * Current focused thumb
+   */
+  @state()
+  private activeThumb!: HTMLDivElement | null;
```
goremikins: You do not need `!` here

packages/elements/src/slider/index.ts
```diff
@@ -575,29 +584,95 @@ export class Slider extends ControlElement {
...
+   */
+  private onApplyStep (direction: Direction): void {
+    // Get current thumb position and step in percentage format
+    const thumbPosition = this.calculatePosition(this.valueNumber, 1);
+    const step = this.calculatePosition(this.minNumber + this.stepRange, 1);
+
+    const possibleValue = direction === Direction.Up ? thumbPosition + step : thumbPosition - step;
+    const nearestPossibleValue = this.getNearestPossibleValue(possibleValue);
+
+    const value = this.getValueFromPercentage(nearestPossibleValue);
```
goremikins: `getValueFromPosition`

goremikins (dismissed):
Please resolve latest minor formatting issues as per my comments and this is ready to go

packages/elements/src/slider/index.ts
wsuwt: Done.

packages/elements/src/slider/index.ts
wsuwt: Done.

Sarin-Udompanish (dismissed):

Sarin-Udompanish (approved):

PR #301, opened 2022-05-11
feat(card): support accessibility

Sakchai-Refinitiv:
## Description
Add A11Y to card element

## Type of change

## Checklist


packages/elements/src/card/index.ts
```diff
@@ -42,31 +45,6 @@ export class Card extends BasicElement {
     return VERSION;
   }
 
-  /**
-   * A `CSSResultGroup` that will be used
-   * to style the host, slotted children
-   * and the internal template of the element.
-   * @returns CSS template
-   */
-  static get styles (): CSSResultGroup {
```
goremikins: Please move it back where it was. I have updated the styling guide: https://confluence.refinitiv.com/display/ELF/Code+Styling+Guide

packages/elements/src/card/__demo__/dev.html
```diff
@@ -0,0 +1,298 @@
+<!DOCTYPE html>
+<html lang="en-GB">
```
goremikins: Why this file is here? Please remove

packages/phrasebook/src/locale/en/card.ts
```diff
@@ -0,0 +1,10 @@
+import { Phrasebook } from '../../translation.js';
+import './shared.js';
```
goremikins: It does not need shared

packages/phrasebook/src/locale/en/card.ts
```diff
@@ -0,0 +1,10 @@
...
+
+const translations = {
+  MENU: 'menu'
```
goremikins: Name is as OPEN_MENU: 'Open Menu'

packages/elements/src/card/index.ts
```diff
@@ -170,6 +158,15 @@ export class Card extends BasicElement {
     this.footerHasContent = isSlotEmpty(event.target as HTMLSlotElement);
   }
 
+  /**
+   * Run on overlay menu open changed
+   * @param event overlay menu opened changed event
+   * @returns {void}
+   */
+  private onMenuOpenChanged (event: OpenedChangedEvent): void {
+    this.openMenuElement?.setAttribute('aria-expanded', String(event.detail.value));
```
goremikins: Do not use `setAttribute` on internal elements. Use `render` lifecycle

packages/elements/src/card/index.ts
```diff
@@ -184,6 +181,27 @@ export class Card extends BasicElement {
...
+   * Run when observed attributes change value
+   * @param name attribute name
+   * @param oldValue old attribute value
+   * @param newValue new attribute value
+   * @returns {void}
+   */
+  public attributeChangedCallback (name: string, oldValue: string | null, newValue: string | null): void {
+    super.attributeChangedCallback(name, oldValue, newValue);
+    if (name === 'aria-level') {
+      this.headingLevel = newValue;
```
goremikins: Why do you set aria-level here? The app developer sets the header manually, so they have to define it:
```html
<ef-card>
        <span slot="header"><h5>Header Slot</h5></span>
        <p>Content</p>
      </ef-card>
```

packages/elements/src/card/index.ts
```diff
@@ -132,13 +112,20 @@ export class Card extends BasicElement {
...
+  private headingLevel: string | null = null;
+
   /**
    * Open menu
    * @returns {void}
    */
   private openMenu (): void {
     if (this.menuElement && !(this.menuElement.fullyOpened || this.menuElement.transitioning)) {
       this.menuElement.opened = true;
+      this.openMenuElement?.setAttribute('aria-expanded', 'true');
```
goremikins: Please do not use `setAttribute` on internal elements. Use `render` lifecycle

goremikins: Here and in all other places

goremikins: If you need to store menu opened state, you can do like this:

```
@state()
private menuOpened = false;
```

packages/phrasebook/src/locale/en/card.ts
goremikins: re: do not forget to update https://jira.refinitiv.com/browse/ELF-1645 with the new translation text

packages/elements/src/card/index.ts
Sakchai-Refinitiv: Can you give me a reason why you want to move it back to the top of the file? I think styles should be a sibling with the template because it is related and easy to visible no need to scroll up and down.

packages/phrasebook/src/locale/en/card.ts
Sakchai-Refinitiv: I think `menu` is more generic than `Open Menu`

packages/phrasebook/src/locale/en/card.ts
Sakchai-Refinitiv: Removed

packages/elements/src/card/__demo__/dev.html
Sakchai-Refinitiv: Removed

packages/elements/src/card/index.ts
Sakchai-Refinitiv: Why does it need to re-render everything again? just set the attribute it might be better

packages/elements/src/card/index.ts
Sakchai-Refinitiv: Why does it need to re-render everything again? just set the attribute it might be better

packages/elements/src/card/index.ts
Sakchai-Refinitiv: The menu should be inside the heading level, not a sibling. what do you think?

packages/elements/src/card/index.ts
goremikins: They were always at the top, but anyway, the reason is that static methods belong to the class instance and it is general convention across many programming languages to keep them at the top of the class.

packages/phrasebook/src/locale/en/card.ts
goremikins: It is more generic, but the button opens menu, is not it? :)

packages/elements/src/card/index.ts
goremikins: It will not re-render, `lit` engine checks if it needs to re-render and checks how to re-render, in this particular case it would update the attribute. It is easy to track the flow of updates if all updates are happening at one place, in this case in `render` method. And another reason, that it is more performant to make all updates in batched update. Please read here: https://lit.dev/docs/components/lifecycle/#reactive-update-cycle on why this is important.

packages/elements/src/card/index.ts
goremikins: Read my comment above....

packages/elements/src/card/index.ts
goremikins: Not sure, menu belongs to the card, not to the heading.

packages/elements/src/card/index.ts
Sakchai-Refinitiv: That gets that point, but I think it should be a sibling with a template that makes more benefits for developers. I will move it back for now and we will discuss it with everyone about coding style later in the questions channel on MS Team.  😀

packages/elements/src/card/index.ts
Sakchai-Refinitiv: I have already discussed it with the Bangkok team, that it is better if we place the menu outside the header because in the VoiceOver users do not need to press an extra key to go inside the header to access the menu.

Sakchai-Refinitiv: The next question is, do we need to support `aria-level` attribute for `header` attribute on `ef-card`? because developers can use slot instead that is the same you mention above example.

packages/phrasebook/src/locale/en/card.ts
Sakchai-Refinitiv: It is OK for both. Action Menu is a new option which one do you prefer?😀

packages/elements/src/card/index.ts
goremikins: We do not need to support, as per example developers can either provide heading element, or provide aria-level themselves.

packages/phrasebook/src/locale/en/card.ts
goremikins: I still prefer `Open Menu` :)

packages/elements/src/card/index.ts
```diff
@@ -147,8 +159,10 @@ export class Card extends BasicElement {
    * @returns {void}
    */
   private closeMenu (): void {
-    if (this.menuElement) {
+    if (this.menuElement && this.menuElement.opened) {
       this.menuElement.opened = false;
+      this.openMenuElement?.setAttribute('aria-expanded', 'false');
+      this.openMenuElement?.focus(); // Need refocus on button because A11Y JAWS lost focus from the element
```
goremikins: Interesting, this is the default behaviour of `overlay` to re-focus back on the element that has opened the menu. Perhaps a bug with the `overlay`, I will double check why it does not do this is this particular case.

packages/elements/src/card/index.ts
Sakchai-Refinitiv: I get it, this is better.

packages/elements/src/card/index.ts
Sakchai-Refinitiv: Fixed

packages/elements/src/card/index.ts
Sakchai-Refinitiv: Currently, I must handle re-focusing in `ef-card` if this is a bug I will remove that code, Can you please check re-focusing when close the menu by pressing key `esc` as well because re-focusing does not work the same.

goremikins: It is a bug, I have created PR: https://github.com/Refinitiv/refinitiv-ui/pull/306

packages/elements/src/card/index.ts
Sakchai-Refinitiv: I think so too, I have already removed the code for aria-level and I will mention this in docs.

packages/elements/src/card/index.ts
Sakchai-Refinitiv: OK, I will remove my code for re-focusing

packages/phrasebook/src/locale/en/card.ts
Sakchai-Refinitiv: Updated

bualoy-napat:
I have already tested the JAWS screen reader. The behavior was correct except the re-focus back to element after press `esc` key.

packages/elements/src/card/index.ts
Sakchai-Refinitiv: I have already removed

Sakchai-Refinitiv:
> I have already tested the JAWS screen reader. The behavior was correct except the re-focus back to element after press `esc` key.

Hi @bualoy-napat it is a bug from Overlay Aleks already create PR for that one https://github.com/Refinitiv/refinitiv-ui/pull/306

goremikins:
The code looks good to me.

@bualoy-napat , could you please check with screen readers?

bualoy-napat:
> The code looks good to me.
> 
> @bualoy-napat , could you please check with screen readers?

Look perfect after I re-checked with the JAWS screen reader.

bualoy-napat (approved):

goremikins (approved):

PR #302, opened 2022-05-11
chore(*): add global element name type definitions

goremikins:
## Description
Add global element name type definitions to ensure that `document.createElement('ef-element-name')` brings the correct type and does not require casting.

## Type of change

## Checklist


packages/elements/src/tooltip/index.ts
```diff
@@ -563,6 +563,12 @@ class Tooltip extends BasicElement {
   }
 }
 
+declare global {
+  interface HTMLElementTagNameMap {
+    'ef-tooltip': Tooltip;
+  }
+}
+
```
TremayneChrist, line 571: Can we move this to the bottom like the rest?

TremayneChrist: It's fine. It's after the class is defined.

TremayneChrist (approved):

PR #308, opened 2022-05-17
feat(clock): improve clock accessibility

Theeraphat-Sorasetsakul:
## Description
Add accessibility for clock

Key requirements: 
- Focusable only when interactive
- Focus point is only at the host 
- Left/Right arrow keys to navigate segments
- Remove Tab focus on segments
- aria-live announces value when focused or changes, e.g. `Time: 12:40PM` 

Fixes # ([ELF-1817](https://jira.refinitiv.com/browse/ELF-1817))

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


Theeraphat-Sorasetsakul:
My solution is
• Default role is group.
• When interactive, set tabindex=0 to host and set role=spinbutton
• Doesn't use aria-live but use aria-valuetext instead.

However, It seem to be problem to VO and custom keydown event for UP/DOWN arrow. Up/Down arrow, Ctrl+Option+Up/Down arrow (VO navigation) don't work to increase and decrease value. But Up/Down arrow with other key such as Alt is work. This problem is also happened on w3: https://www.w3.org/TR/wai-aria-practices/examples/spinbutton/datepicker-spinbuttons.html.

packages/elemental-theme/src/custom-elements/ef-clock.less
```diff
@@ -74,10 +74,11 @@
...
 
-  [part~='hours']:focus, [part~='minutes']:focus, [part~='seconds']:focus {
-    color: @scheme-color-primary;
-
-    [part='increment-button'], [part='decrement-button'] {
+  &:focus {
+    [part~='hours'], [part~='minutes'], [part~='seconds'] {
+      color: @scheme-color-primary;
+    }
+    [active]{
```
goremikins: Wrong formatting, space is missing after ].

Why do you need it for focus? Can active be always set to `block`?

packages/elements/src/clock/index.ts
```diff
@@ -130,6 +132,12 @@ export class Clock extends ResponsiveElement {
   @state()
   private tickTimestamp = 0;
 
+  /**
+   * Current active state for interactive
+   */
+  @state()
+  private active = 'hours';
```
goremikins: 1. Create `enum` to describe possible state
2. `active` is too common name, rename it to e.g. `activeSegment`

packages/elements/src/clock/index.ts
```diff
@@ -482,28 +507,68 @@ export class Clock extends ResponsiveElement {
...
+    this.renderRoot.querySelector(`[part~="${this.active}"]`)?.querySelector('[part="decrement-button"]')?.setAttribute('active', '');
+  }
+
+  /**
+  * Updates aria-value-text to have the same value as screen
+  * @returns {void}
+  */
+  private updateAriaValueText () {
+    const value = `Time: ${padNumber(this.displayHours, 2)}:${padNumber(this.displayMinutes, 2)}${this.showSeconds ? ':' + padNumber(this.displaySeconds, 2) : ''}${this.amPm ? ' ' + this.displayAmPm : ''}`;
+    this.ariaValueText = value;
```
goremikins: Do not use AOM model, it is still not supported by Firefox and not yet a standard

packages/elements/src/clock/index.ts
```diff
@@ -482,28 +507,68 @@ export class Clock extends ResponsiveElement {
...
+  }
+
+  /**
+  * Updates active attribute for increment and decrement button
+  * by new segment
+  * @param segment new active segment
+  * @returns {void}
+  */
+  private changeActiveSegment (segment: string) {
+    this.renderRoot.querySelectorAll('[active]')?.forEach((el) => el.removeAttribute('active'));
```
goremikins: Can you do all these as part of `render` cycle?

packages/elements/src/clock/index.ts
goremikins: In addition, ariaValueText should be used with ariaValueNow. I cannot find the latest.

packages/elements/src/clock/index.ts
```diff
@@ -601,10 +666,27 @@ export class Clock extends ResponsiveElement {
...
+    this.addEventListener('keydown', (event) => this.onKeydown(event));
     this.renderRoot.addEventListener('tapstart', (event) => this.onTapStart(event as TapEvent));
   }
 
+  protected update (changedProperties: PropertyValues): void {
+    super.update(changedProperties);
+
+    if (changedProperties.has('interactive')) {
+      if (this.interactive) {
+        this.tabIndex = 0;
```
goremikins: Why not to announce values in the same way as time-picker using `role=status`?

packages/elements/src/clock/index.ts
```diff
@@ -482,28 +507,68 @@ export class Clock extends ResponsiveElement {
...
+  * @param segment active segment
   * @returns {TemplateResult} template result
   */
-  private generateButtonsTemplate (): TemplateResult {
+  private generateButtonsTemplate (segment: string): TemplateResult {
+    const isFocused = segment === this.active;
     return html`
-      <div part="increment-button" role="button" data-key="${UP}"></div>
-      <div part="decrement-button" role="button" data-key="${DOWN}"></div>
+      <div part="increment-button" role="button" data-key="${UP}" aria-hidden="true"
```
goremikins: I am not sure who did and approved the use of `data-key`, but can you please change it to use references instead: https://lit.dev/docs/templates/directives/#ref.

packages/elemental-theme/src/custom-elements/ef-clock.less
Theeraphat-Sorasetsakul: Updated to add space.
I think it's been designed that way since v3. To make it always show, may need to talk with World clock team.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated to create enum and changed name.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated by remove this function because it already done by `active="${ifDefined(isFocused ? '' : undefined)}"`.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated to not use AOM and add aria-valuenow. But I found aria-valuenow isn't necessary if we use aria-valuetext. Because screen reader like JAW announces `aria-valuetext` by replace `aria-valuenow`. Note sure why doc like MDN said it is required.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: It prefers to use `aria-valuetext` instead of having an `aria-live` or `role=status` element. Because having `aria-live` or `role=status` element inside `role=spinbutton` isn't work for screen reader. I have proven with Goh that it doesn't work.
To make it work, it should place in the same level of dom.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated to use ref instead of dataset. I create upButtonRef and downButtonRef, and change logic on `onTapStart` for ref solution.

packages/elements/src/clock/index.ts
```diff
@@ -601,10 +673,31 @@ export class Clock extends ResponsiveElement {
    */
   protected firstUpdated (changedProperties: PropertyValues): void {
     super.firstUpdated(changedProperties);
-    this.renderRoot.addEventListener('keydown', (event) => this.onKeydown(event as KeyboardEvent));
+    this.addEventListener('keydown', (event) => this.onKeydown(event));
     this.renderRoot.addEventListener('tapstart', (event) => this.onTapStart(event as TapEvent));
   }
 
+  protected update (changedProperties: PropertyValues): void {
```
wsuwt: Should we use `willUpdate` instead ?

packages/elements/src/clock/index.ts
```diff
@@ -482,28 +526,55 @@ export class Clock extends ResponsiveElement {
...
+  private handleLeftRightKey (): void {
+    this.activeSegment = this.activeSegment === Segment.HOURS ? Segment.MINUTES : Segment.HOURS;
+  }
+
+  /**
+  * Updates aria-value-text to have the same value as screen
+  * @returns {void}
+  */
+  private updateAriaValue () {
+    const value = `Time: ${padNumber(this.displayHours, 2)}:${padNumber(this.displayMinutes, 2)}${this.showSeconds ? ':' + padNumber(this.displaySeconds, 2) : ''}${this.amPm ? ' ' + this.displayAmPm : ''}`;
```
wsuwt: Can this be broken down into smaller statements ?

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated to use `willUpdate`.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated by break the line.

packages/elements/src/clock/index.ts
```diff
@@ -482,28 +526,58 @@ export class Clock extends ResponsiveElement {
...
+    this.activeSegment = this.activeSegment === Segment.HOURS ? Segment.MINUTES : Segment.HOURS;
+  }
+
+  /**
+  * Updates aria-value-text to have the same value as screen
+  * @returns {void}
+  */
+  private updateAriaValue () {
+    const value = `Time: ${padNumber(this.displayHours, 2)}`
+      + `:${padNumber(this.displayMinutes, 2)}`
```
wsuwt: Can we extract `:` separator like we did in a lot of elements ? Would that make it easier to read ?

packages/elements/src/clock/index.ts
```diff
@@ -482,28 +526,58 @@ export class Clock extends ResponsiveElement {
...
+  private handleLeftRightKey (): void {
+    this.activeSegment = this.activeSegment === Segment.HOURS ? Segment.MINUTES : Segment.HOURS;
+  }
+
+  /**
+  * Updates aria-value-text to have the same value as screen
+  * @returns {void}
+  */
+  private updateAriaValue () {
+    const value = `Time: ${padNumber(this.displayHours, 2)}`
```
wsuwt: Should the word `Time` be translated ?

packages/elements/src/clock/index.ts
wsuwt: Also other than breaking the lines, extracting hours, minute etc. into variables is an alternative.

documents/src/pages/elements/clock.md
```diff
@@ -234,3 +234,12 @@ ef-clock {
 <ef-clock analogue tick show-seconds am-pm></ef-clock>
 ```
 
+
+## Accessibility
+::a11y-intro::
+
+`ef-clock` is a set of number segments that contain hours, minutes, seconds, and time convention periods. It is assigned with `role="group"` and can include `aria-label` to describe the context of time.
```
goremikins: I do not think we need any of this text, as the clock works out of the box and does not require any additional work from application developers. @dtanp-rft , do you agree?

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -18,12 +18,24 @@ describe('clock/Interactive', () => {
     }
 
     const InputKey = {
+      arrowLeft: {
+        ieKey: 'Left',
+        key: 'ArrowLeft',
+        which: 37,
```
goremikins: `which` and `keyCode` are deprecated APIs. You can safely remove those

packages/elements/src/clock/__test__/clock.test.js
```diff
@@ -111,5 +111,14 @@ describe('clock/Clock', () => {
       expect(valueEvent.detail.value, 'event\'s value should be 00:00:01').to.be.equal('00:00:01');
     });
   });
+
+  describe('Tick', () => {
+    beforeEach(async () => {
```
goremikins: Why do you need `beforeEach` here?

packages/elements/src/clock/index.ts
```diff
@@ -62,6 +68,8 @@ export class Clock extends ResponsiveElement {
     return VERSION;
   }
 
+  protected readonly defaultRole: string | null = 'group';
```
goremikins: There is no defaultRole for the component as the role is based on `interactive` attribute. Keep is as default (`null`)

packages/elements/src/clock/index.ts
```diff
@@ -482,28 +526,58 @@ export class Clock extends ResponsiveElement {
   * @param event Event Object
   * @returns {void}
   */
-  private handleUpKey (event: KeyboardEvent): void {
-    this.shift(UP, this.getShiftAmountFromTarget(event.target));
+  private handleUpKey (): void {
+    const selection = this.renderRoot.querySelector(`[part~=${this.activeSegment}]`);
+    this.shift(Direction.UP, this.getShiftAmountFromTarget(selection));
```
goremikins: This code needs to be refactored to remove querySelector and `getShiftAmountFromTarget` function. You already know what is the `activeSegment` and can calculate `shiftAmount` from the active segment.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Resolved by Alek PR

documents/src/pages/elements/clock.md
Theeraphat-Sorasetsakul: Resolved by Alek PR.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Resolved by Alek PR

documents/src/pages/elements/clock.md
dtanp-rft: Yes the statement is wrong. I've revised this section and already push the changes.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated by no role by default

packages/elements/src/clock/index.ts
```diff
@@ -130,6 +142,18 @@ export class Clock extends ResponsiveElement {
   @state()
   private tickTimestamp = 0;
 
+  /**
+   * Current active state for interactive
```
wsuwt:
```suggestion
   * Current active segment for interactive mode
```

packages/elements/src/clock/index.ts
```diff
@@ -259,11 +298,21 @@ export class Clock extends ResponsiveElement {
   @query('[part~=seconds]', true)
   private secondsPart!: HTMLDivElement;
 
+  /**
+   * A reference for active up button
```
wsuwt:
```suggestion
   * A reference for up button
```

packages/elements/src/clock/index.ts
```diff
@@ -259,11 +298,21 @@ export class Clock extends ResponsiveElement {
...
+   */
+  private upButtonRef: Ref<HTMLElement> = createRef();
+
+  /**
+   * A reference for active down button
```
wsuwt:
```suggestion
   * A reference for down button
```

packages/elements/src/clock/index.ts
```diff
@@ -449,8 +511,22 @@ export class Clock extends ResponsiveElement {
    * @returns {void}
    */
   private onTapStart (event: TapEvent): void {
-    if (event.target instanceof HTMLElement && event.target.dataset.key) {
-      this.shift(event.target.dataset.key as UpOrDown, this.getShiftAmountFromTarget(event.target));
+    // do nothing if not interactive
```
wsuwt:
```suggestion
```
It is already self-explanatory.

packages/elements/src/clock/index.ts
```diff
@@ -479,46 +568,69 @@ export class Clock extends ResponsiveElement {
...
   * @returns {void}
   */
-  private handleDownKey (event: KeyboardEvent): void {
-    this.shift(DOWN, this.getShiftAmountFromTarget(event.target));
+  private handleDownKey (): void {
+    this.shift(Direction.DOWN, this.getShiftAmountFromSegment(this.activeSegment));
+  }
+
+  /**
+  * Updates aria-value-text to have the same value as screen
```
wsuwt:
```suggestion
  * Set aria-valuenow to display value and aria-valuetext to translated format
```

packages/elements/src/clock/index.ts
```diff
@@ -479,46 +568,69 @@ export class Clock extends ResponsiveElement {
...
+  * @param segment active segment
   * @returns {TemplateResult} template result
   */
-  private generateButtonsTemplate (): TemplateResult {
+  private generateButtonsTemplate (segment: Segment): TemplateResult {
+    const isActive = this.focused && segment === this.activeSegment;
     return html`
-      <div part="increment-button" role="button" data-key="${UP}"></div>
-      <div part="decrement-button" role="button" data-key="${DOWN}"></div>
+      <div ${isActive ? ref(this.upButtonRef) : undefined}
```
wsuwt: Why do you need to add and remove ref when active changes ?

packages/elements/src/clock/index.ts
```diff
@@ -601,10 +713,29 @@ export class Clock extends ResponsiveElement {
    */
   protected firstUpdated (changedProperties: PropertyValues): void {
     super.firstUpdated(changedProperties);
-    this.renderRoot.addEventListener('keydown', (event) => this.onKeydown(event as KeyboardEvent));
+    this.addEventListener('keydown', (event) => this.onKeydown(event));
```
wsuwt, line 698: Why one is added to `renderRoot` and the other one is not ?

packages/elements/src/clock/index.ts
```diff
@@ -479,46 +568,69 @@ export class Clock extends ResponsiveElement {
...
-      <div part="increment-button" role="button" data-key="${UP}"></div>
-      <div part="decrement-button" role="button" data-key="${DOWN}"></div>
+      <div ${isActive ? ref(this.upButtonRef) : undefined}
+           part="increment-button"
+           role="button"
+           aria-hidden="true"
+           ?active=${isActive}></div>
+      <div ${isActive ? ref(this.downButtonRef) : undefined}
+           part="decrement-button"
+           role="button"
```
wsuwt: What's the point of even having role at all when it is `aria-hidden` ? I assume there is no difference at all.

packages/elements/src/clock/index.ts
```diff
@@ -479,46 +568,69 @@ export class Clock extends ResponsiveElement {
...
 
   /**
   * Template for increment and decrement button
   * if interactive mode is enabled.
+  * @param segment active segment
   * @returns {TemplateResult} template result
   */
-  private generateButtonsTemplate (): TemplateResult {
+  private generateButtonsTemplate (segment: Segment): TemplateResult {
+    const isActive = this.focused && segment === this.activeSegment;
```
wsuwt: Shouldn't this be the same as line 628 https://github.com/Refinitiv/refinitiv-ui/pull/308/files#diff-9683b0f1fc8b119d1e15fc2156fa96112e9615d2d14a6aa657a6f0598c5faf48R628 ?

wsuwt: Would it be better if it becomes a common function ?

packages/elements/src/clock/__test__/clock.test.js
Theeraphat-Sorasetsakul: Updated to remove this because it use only once times.

packages/elements/src/clock/__test__/clock.interactive.test.js
Theeraphat-Sorasetsakul: Updated to remove

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Because the ref refers to active buttons.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated by remove the condition at `generateButtonTemplate`. Because `generateSegmentTemplate` is already checked

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated by remove the role

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated by remove it

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Resolved by your PR

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -65,49 +69,49 @@ describe('clock/Interactive', () => {
...
 
       expect(el.offset, 'offset should be 60').to.be.equal(60);
       expect(el.displayMinutes, 'minutes should be 1').to.be.equal(1);
     });
     it('Decreases minute when decrease button is clicked', async () => {
-      const decrementBtn = minutesSegment.querySelector('[part=decrement-button]');
-
       await onTapstart(minutesSegment, el);
+      
+      const decrementBtn = minutesSegment.querySelector('[part=decrement-button]');
```
wsuwt: Extract these internal DOM references to a single variable. So that it is easier to update when part name changes.

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -132,11 +136,86 @@ describe('clock/Interactive', () => {
...
+      });
+      it('Switch activeSegment to hours when arrow Right', async () => {
+        await onTapstart(hoursSegment, el);
+        createKeyboardEvent(el, InputKey.arrowRight);
+        await elementUpdated(el);
+
+        const hoursIncrementBtn = hoursSegment.querySelector('[part=increment-button]');
+        const hoursDecrementBtn = hoursSegment.querySelector('[part=decrement-button]');
+        const minuteIncrementBtn = minutesSegment.querySelector('[part=increment-button]');
+        const minuteDecrementBtn = minutesSegment.querySelector('[part=decrement-button]');
```
wsuwt: Here as well.

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -132,11 +136,86 @@ describe('clock/Interactive', () => {
...
+      it('Should update aria-valuetext after arrow Up on increase minute', async () => {
+        await onTapstart(minutesSegment, el);
+        createKeyboardEvent(el, InputKey.arrowUp);
+        await elementUpdated(el);
+        await nextFrame();
+  
+        expect(el.getAttribute('aria-valuetext')).to.be.equal('Time: 00:01');
+        expect(el.getAttribute('aria-valuenow')).to.be.equal(el.displayTime.toString());
+      });
+      it('Should update aria-valuetext after arrow Down on decrease minute', async () => {
```
wsuwt:
```suggestion
      it('Should update aria-valuetext when Arrow Down is pressed on minute segment', async () => {
```

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -132,11 +136,86 @@ describe('clock/Interactive', () => {
...
+        await onTapstart(minutesSegment, el);
+        createKeyboardEvent(el, InputKey.arrowUp);
+        await elementUpdated(el);
+        await nextFrame();
+  
+        expect(el.getAttribute('aria-valuetext')).to.be.equal('Time: 00:01');
+        expect(el.getAttribute('aria-valuenow')).to.be.equal(el.displayTime.toString());
+      });
+      it('Switch activeSegment to hours when arrow Right', async () => {
```
wsuwt:
```suggestion
      it('Switches segment to hours when Arrow Right is pressed', async () => {
```

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -132,11 +136,86 @@ describe('clock/Interactive', () => {
...
       expect(offsetChangedCount, 'offset-changed count should be 1').to.be.equal(1);
       expect(offsetEvent.detail.value, '  should be 60').to.be.equal(60);
     });
+    
+    describe('Accessibility', () => {
+      it('Should have role="spinbutton" and focusable', async () => {
+        expect(el.getAttribute('role')).to.be.equal('spinbutton');
+        expect(el.getAttribute('tabindex')).to.be.equal('0');
+      });
+      it('Should update aria-valuetext after arrow Up on increase hour', async () => {
```
wsuwt:
```suggestion
      it('Should increase hour value and update aria-valuetext when Arrow Up is pressed on hour segment', async () => {
```

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -18,17 +18,21 @@ describe('clock/Interactive', () => {
     }
 
     const InputKey = {
```
wsuwt, line 24: `InputKey.ArrowLeft` is so much easier to read.

packages/elements/src/clock/index.ts
```diff
@@ -502,23 +609,30 @@ export class Clock extends ResponsiveElement {
...
-  private generateSegmentTemplate (name: string, value: number): TemplateResult {
+  private generateSegmentTemplate (segment: Segment, value: number): TemplateResult {
+    const isActive = this.interactive && (segment === this.activeSegment);
     return html`
-      <div part="segment ${name}${ifDefined(this.isSegmentShifted(name) ? ' shifted' : '')}" tabindex="${ifDefined(this.interactive ? '0' : undefined)}">
+      <div part="segment ${segment}${this.isSegmentShifted(segment) ? ' shifted' : ''}"
+           ?active=${isActive}>
         ${padNumber(value, 2)}
-        ${this.interactive ? this.generateButtonsTemplate() : undefined}
+        ${isActive ? this.generateButtonsTemplate() : undefined}
```
wsuwt: This is still forcing the re-generation of the whole template. Instead of button always have `active` why don't you utilise active attribute and use CSS to toggle buttons ?

```suggestion
         <div ${ref(this.upButtonRef)}
            part="increment-button"
            aria-hidden="true"
            active=${ifDefined(isActive || undefined)}></div>
          <div ${ref(this.downButtonRef)}
            part="decrement-button"
            aria-hidden="true"
            active=${ifDefined(isActive || undefined)}></div>
```

packages/elements/src/clock/index.ts
```diff
@@ -220,27 +243,50 @@ export class Clock extends ResponsiveElement {
...
+      this._interactive = newValue;
+      if (this._interactive) {
+        this.tabIndex = 0;
+        this.setAttribute('role', 'spinbutton');
+      }
+      else {
+        this.tabIndex = -1;
+        this.removeAttribute('role');
+      }
+      this.requestUpdate('tick', oldValue);
```
wsuwt: Tick ?

packages/elements/src/clock/index.ts
```diff
@@ -220,27 +243,50 @@ export class Clock extends ResponsiveElement {
...
+  /**
+   * Enable interactive mode. Allowing the user to offset the value.
+   * Set tabIndex >= 0 to make the clock interactive
+   * Set/remove tabIndex to move clock to non interactive mode
+   * @param value Set interactive mode
+   */
+  @property({ type: Boolean })
+  public set interactive (value: boolean) {
+    const newValue = !!value;
+    const oldValue = this.tick;
```
goremikins: `tick`?

packages/elements/src/clock/__test__/clock.interactive.test.js
Theeraphat-Sorasetsakul: Updated to create reference variable.

packages/elements/src/clock/__test__/clock.interactive.test.js
Theeraphat-Sorasetsakul: Updated to create reference variable.

packages/elements/src/clock/__test__/clock.interactive.test.js
Theeraphat-Sorasetsakul: Updated to rename test case

packages/elements/src/clock/__test__/clock.interactive.test.js
Theeraphat-Sorasetsakul: Updated to rename test case

packages/elements/src/clock/__test__/clock.interactive.test.js
Theeraphat-Sorasetsakul: Updated to rename key name

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: My mistake. It should be interactive. Updated to interactive.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: My mistake. It should be interactive. Updated to interactive.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: If we do this way, default clock will have hidden button. Do we really want this?

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: I put condition because ref should be referred to active buttons. The ref identifies only one element. If it doesn't check, the ref will refers to the last rendered element.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: For keydown event, it listens on host because current focus point is changed to host.
For tapstart event, it listens on renderRoot because the event need to identify which target is pressed. If the event listens on host, event.target will get only host.

packages/elements/src/clock/index.ts
wsuwt: What is the issue with hidden button ? I only see that it should be more performant.

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -39,75 +47,79 @@ describe('clock/Interactive', () => {
...
       minutesSegment = el.shadowRoot.querySelector('[part~=minutes]');
+      updateBtnSelector();
     });
 
     it('Increases hour when increase button is clicked', async () => {
-      const incrementBtn = hoursSegment.querySelector('[part=increment-button]');
-
       await onTapstart(hoursSegment, el);
-      await onTapstart(incrementBtn, el);
+      updateBtnSelector();
```
wsuwt: Why do you need to call this function again when you already called it in `forEach` ?

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -39,75 +47,79 @@ describe('clock/Interactive', () => {
       elem.dispatchEvent(keyboardEvent('keyup', keyOption));
     };
 
+    const updateBtnSelector = () => {
+      incrementBtnInHours = hoursSegment.querySelector('[part=increment-button]');
+      decrementBtnInHours = hoursSegment.querySelector('[part=decrement-button]');
+      incrementBtnInMinutes = minutesSegment.querySelector('[part=increment-button]');
+      decrementBtnInMinutes = minutesSegment.querySelector('[part=decrement-button]');
```
wsuwt: A function referencing variables `hoursSegment` outside is bad. Please pass `el` into this function.

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -132,11 +144,78 @@ describe('clock/Interactive', () => {
...
+      });
+      it('Switches segment to hours when Arrow Right is pressed', async () => {
+        await onTapstart(hoursSegment, el);
+        createKeyboardEvent(el, InputKey.ArrowRight);
+        await elementUpdated(el);
+        updateBtnSelector();
+
+        expect(el.activeSegment).to.be.equal('minutes');
+        expect(incrementBtnInHours).to.be.equal(null);
+        expect(decrementBtnInHours).to.be.equal(null);
```
wsuwt: If you want to test that segment switches, only test for line 201. If you want to test that buttons should not show then do it in separate test case.

packages/elements/src/clock/index.ts
```diff
@@ -220,27 +243,50 @@ export class Clock extends ResponsiveElement {
...
+
+  /**
+   * Enable interactive mode. Allowing the user to offset the value.
+   * Set tabIndex >= 0 to make the clock interactive
+   * Set/remove tabIndex to move clock to non interactive mode
+   * @param value Set interactive mode
+   */
+  @property({ type: Boolean })
+  public set interactive (value: boolean) {
+    const newValue = !!value;
```
wsuwt: Why ? `type: Boolean` should already convert it to boolean.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: It's ok if you agree to have buttons in default clock (You can see what changes in snapshot). I've updated to remove `generateButtonTemplate` and do your suggestion. But I add more condition by `${isActive ? ref(this.upButtonRef) : undefined}`.

packages/elements/src/clock/__test__/clock.interactive.test.js
```diff
@@ -18,17 +22,21 @@ describe('clock/Interactive', () => {
     }
 
     const InputKey = {
-      arrowUp: {
+      ArrowLeft: {
+        ieKey: 'Left',
```
wsuwt, line 26: When do you use `ieKey` ?

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated to remove `!!`. Also removed at `public set tick`.

packages/elements/src/clock/index.ts
```diff
@@ -207,7 +231,7 @@ export class Clock extends ResponsiveElement {
     return this._tick;
   }
   public set tick (value: boolean) {
-    const newValue = !!value;
+    const newValue = value;
```
wsuwt: You don't need to declare `newValue`

packages/elements/src/clock/__test__/clock.interactive.test.js
Theeraphat-Sorasetsakul: Updated to use iekey. https://github.com/Refinitiv/refinitiv-ui/pull/308/commits/174dfccb62ceaa7662e9633281a50888009d85cf

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated to  remove newValue and use coming value instead.

packages/elements/src/clock/index.ts
wsuwt: Please take a look at this PR https://github.com/Refinitiv/refinitiv-ui/pull/329, this is an alternative solution without using ref. Choose either this or multiple `ref` but not adding and removing `ref` on active change.

packages/elements/src/clock/index.ts
```diff
@@ -220,27 +241,49 @@ export class Clock extends ResponsiveElement {
...
   */
   @property({ type: Boolean, reflect: true })
   public analogue = false;
 
+  private _interactive = false;
+
+  /**
+   * Enable interactive mode. Allowing the user to offset the value.
+   * Set tabIndex >= 0 to make the clock interactive
+   * Set/remove tabIndex to move clock to non interactive mode
```
wsuwt: Please update comments according to what the code actually does.

packages/elements/src/clock/index.ts
```diff
@@ -220,27 +241,49 @@ export class Clock extends ResponsiveElement {
...
    */
-  @property({ type: Boolean, attribute: 'am-pm' })
+  @property({ type: Boolean, attribute: 'am-pm', reflect: true })
   public amPm = false;
 
   /**
    * Display the seconds segment.
    */
-  @property({ type: Boolean, attribute: 'show-seconds' })
+  @property({ type: Boolean, attribute: 'show-seconds', reflect: true })
```
wsuwt, line 250: Why reflect ?

packages/elements/src/clock/index.ts
```diff
@@ -220,27 +241,49 @@ export class Clock extends ResponsiveElement {
   /**
    * Display the digital clock in 12hr format.
    */
-  @property({ type: Boolean, attribute: 'am-pm' })
+  @property({ type: Boolean, attribute: 'am-pm', reflect: true })
```
wsuwt, line 244: Why reflect ?

Theeraphat-Sorasetsakul: It actually is an idea from Alek. It should be reflect because it effects to UI.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: It actually is an idea from Alek. It should be reflect because it effects to UI.

goremikins: to be inline with the time-picker. Potentially these properties can have UI differences.

packages/elements/src/clock/index.ts
Theeraphat-Sorasetsakul: Updated the comments.

packages/elements/src/clock/index.ts
wsuwt: Got it. Thanks.

wsuwt (approved):

PR #311, opened 2022-05-19
Refactor code

goremikins:
@wsuwt , @Theeraphat-Sorasetsakul I have refactored the code to what I think it should look like.

Please ensure that screen readers still work and fix tests.


wsuwt:
@goremikins I decided that we should keep `role` as unset. `JAWS` will skip the text if role is set to `spinbutton` without focusable `tabIndex`. In case when label is needed in non-interactive mode, developers can set `role="group"` and label themselves.

Theeraphat-Sorasetsakul:
I'm going to merge this and edit it to no role by default in [PR - feat(clock): improve clock accessibility](https://github.com/Refinitiv/refinitiv-ui/pull/308)

Theeraphat-Sorasetsakul (approved):

PR #328, opened 2022-05-24
chore: translate aria message for Chinese

dtanp-rft:
Transation of aria message for Traditional & Simplified Chinese.


Sakchai-Refinitiv (approved):
Look good

PR #329, opened 2022-05-24
Clock refactor

wsuwt:
## Description
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/clock/__demo__/index.html
+++ b/packages/elements/src/clock/__demo__/index.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <title>Clock</title>
     <link rel="stylesheet" href="/node_modules/@refinitiv-ui/demo-block/demo.css">
   </head>
   <body>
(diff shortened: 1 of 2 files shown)


Theeraphat-Sorasetsakul:
This will goes back to the old way that Clock uses `data-key` for button elements. https://github.com/Refinitiv/refinitiv-ui/pull/308#pullrequestreview-975461938. Alek suggests to use ref instead of `data-key` that attach as attribute.

Theeraphat-Sorasetsakul:
If we try to check direction on `name` attribute. I think the old way, `data-key`, will be suited and more readable that attribute. How do you think?

Theeraphat-Sorasetsakul:
> If we try to check direction on `name` attribute. I think the old way, `data-key`, will be suited and more readable that attribute. How do you think?

After look at other element that use `data-*`, it look like old style code that `data-*` appears only on `interactive-chart`, `color-dialog`, and `clock`. `data-*` doesn't matched for our code style. So I agree to use `name` attribute with Direction.

Theeraphat-Sorasetsakul (approved):

PR #331, opened 2022-05-25
fix(tree): render wrong state when data is invalid

Nantawat-Poothong:
## Description
In this case is fixed about tree element render wrong checked/unchecked state when data is invalid. 

It seems that tree-manager recursive logic checkItem/uncheckItem when call item descendants is always sent false to `manageRelationships` parameter and not work perfectly when item level 1 called item level 2 with descendants.

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-1835

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/tree/managers/tree-manager.ts
+++ b/packages/elements/src/tree/managers/tree-manager.ts
@@ -388,7 +388,7 @@ export class TreeManager<T extends TreeDataItem> {
       this.composer.setItemPropertyValue(item, 'selected', true);
       if (manageRelationships) {
         this.forceUpdateOnPath(item);
-        this.getItemDescendants(item).forEach(descendant => this._checkItem(descendant, false));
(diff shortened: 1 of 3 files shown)


packages/elements/src/tree/managers/tree-manager.ts
```diff
@@ -388,7 +388,7 @@ export class TreeManager<T extends TreeDataItem> {
       this.composer.setItemPropertyValue(item, 'selected', true);
       if (manageRelationships) {
         this.forceUpdateOnPath(item);
-        this.getItemDescendants(item).forEach(descendant => this._checkItem(descendant, false));
+        this.getItemDescendants(item).forEach(descendant => this._checkItem(descendant, this.isItemParent(descendant)));
```
goremikins, line 391: I am unsure how much the performance is affected by this. @TremayneChrist ?

PR #337, opened 2022-05-27
Accessibility/rating

Sarin-Udompanish:
## Description
`ef-rating` does not assign any roles as it's not an actionable component.

However, in the interactive mode, rating will be assigned `role="slider"`. Users can use arrow keys to update value. Current value of the rating will be updated to `aria-valuenow`. You can assign `aria-label` and `aria-valuetext` to `ef-rating` to give the component an assistive name for screen reader.

### Keyboard Support

Right / Up Arrow - Increases rating one star.
Left / Down Arrow - Increases rating one star.
Home - Sets rating to its minimum value, one stars.
End - Sets rating to its maximum value, five stars.

## Type of change

## Checklist


packages/elements/src/rating/index.ts
```diff
@@ -8,7 +8,6 @@ import {
 } from '@refinitiv-ui/core';
 import { customElement } from '@refinitiv-ui/core/decorators/custom-element.js';
 import { property } from '@refinitiv-ui/core/decorators/property.js';
-import { queryAll } from '@refinitiv-ui/core/decorators/query-all.js';
 import { repeat } from '@refinitiv-ui/core/directives/repeat.js';
 import { VERSION } from '../version.js';
 interface ItemType {
```
wsuwt: Can we remove `Type` from this interface and improve its naming ?

wsuwt: I also doubt it really need to be an object.

packages/elements/src/rating/index.ts
```diff
@@ -86,6 +96,56 @@ export class Rating extends BasicElement {
...
+  /**
+   * Invoked whenever the element properties are updated
+   * @param changedProperties changed properties
+   * @returns {void}
+   */
+  protected updated (changedProperties: PropertyValues): void {
+    super.updated(changedProperties);
+    changedProperties.forEach((_, propName) => {
+      if (propName === 'value') {
+        this.value = this.valueNumber.toString();
```
wsuwt: Can we do this in setter or `willUpdate` instead ?

packages/elements/src/rating/index.ts
```diff
@@ -86,6 +96,56 @@ export class Rating extends BasicElement {
...
+        this.computeRating(this.maxNumber, this.valueNumber);
```
wsuwt: Will this trigger another update ? Can this be move to `willUpdate`

packages/elements/src/rating/index.ts
```diff
@@ -32,9 +31,25 @@ export class Rating extends BasicElement {
...
+   */
+  static get styles (): CSSResultGroup {
+    return css`
+      :host {
+        display: inline-block;
+      }
+    `;
+  }
+
   private items: ItemType[] = [];
```
wsuwt: IMO, `items` is such a bad naming. Can we replace this with something else ?

packages/elements/src/rating/index.ts
```diff
@@ -111,54 +171,118 @@ export class Rating extends BasicElement {
...
    */
-  private handleTap (index: number): void {
-    if (!this.interactive) {
+  private interactiveChanged (): void {
+    if (this.interactive) {
+      this.setAttribute('role', 'slider');
+      this.setAttribute('aria-valuemin', `${this.interactiveMinValue}`);
+      this.setAttribute('aria-valuenow', this.value);
+      this.setAttribute('aria-valuemax', this.max);
+      this.tabIndex = 0;
```
wsuwt: You can use `setAttribute` for consistency

packages/elements/src/rating/index.ts
```diff
@@ -32,9 +31,25 @@ export class Rating extends BasicElement {
...
+      :host {
+        display: inline-block;
+      }
+    `;
+  }
+
   private items: ItemType[] = [];
   private valuePrevious = 0;
 
+  private interactiveMinValue = 1; // min value of interactive mode
```
wsuwt: I would prefer this constant in all caps. Something like `MIN_VALUE`

packages/elements/src/rating/index.ts
```diff
@@ -111,54 +171,118 @@ export class Rating extends BasicElement {
...
+   * @param {number} index index of star
+   * @returns {void}
+   */
+  private handleTap (index: number): void {
+    if (!this.interactive) {
+      return;
+    }
+
+    const valueIndex = this.maxNumber - index;
+    if (this.valuePrevious !== valueIndex) {
```
wsuwt: Why not `this.value` ?

packages/elements/src/rating/index.ts
```diff
@@ -32,9 +31,25 @@ export class Rating extends BasicElement {
...
+  static get styles (): CSSResultGroup {
+    return css`
+      :host {
+        display: inline-block;
+      }
+    `;
+  }
+
   private items: ItemType[] = [];
   private valuePrevious = 0;
```
wsuwt: I don't think we need this at all.

packages/elements/src/rating/index.ts
```diff
@@ -111,54 +171,118 @@ export class Rating extends BasicElement {
...
+
+    event.preventDefault();
+  }
+
+  /**
+   * Update value and fired value-changed event
+   * @param value value to updated
+   * @returns {void}
+   */
+  private updateValue (value: number | string): void {
```
wsuwt: It would be good to make this function to serve one purpose, that is update and notify value not converting anything at all. I would prefer this value to only accept `string`

packages/elements/src/rating/index.ts
```diff
@@ -111,54 +171,118 @@ export class Rating extends BasicElement {
...
-        this.value = this.valueNumber.toString();
-        this.computeRating(this.maxNumber, this.valueNumber);
-      }
-      else if (propName === 'max') {
-        this.max = this.maxNumber.toString();
-        this.computeRating(this.maxNumber, this.valueNumber);
-      }
-    });
+  private stepUp (): void {
+    if (this.valueNumber + 1 > this.maxNumber) {
```
wsuwt: **4.5 + 1 is more than 5** so in that case it won't update.

packages/elements/src/rating/index.ts
```diff
@@ -111,54 +171,118 @@ export class Rating extends BasicElement {
...
+   * @returns {void}
    */
-  static get styles (): CSSResultGroup {
-    return css`
-      :host {
-        display: inline-block;
-      }
-    `;
+  private stepDown (): void {
+    if (this.valueNumber - 1 < this.interactiveMinValue) {
```
wsuwt: Also in this case, when value is `1.5` stepDown will not be able to update the value.

packages/elements/src/rating/index.ts
```diff
@@ -54,11 +69,6 @@ export class Rating extends BasicElement {
   @property({ type: String, reflect: true })
```
wsuwt: I'm not sure why we need to reflect.

packages/elements/src/rating/__demo__/index.html
```diff
@@ -42,14 +42,23 @@
 
   <demo-block layout="normal" header="Interactive" tags="interactive,value">
     <div>
-      <ef-rating interactive value="5"></ef-rating>
+      <ef-rating interactive value="-5"></ef-rating>
```
wsuwt, line 68: I think we should show default case first in the first example of demo-block.

documents/src/pages/elements/rating.md
```diff
@@ -140,3 +140,19 @@ customRating.addEventListener('value-changed', (event) =>  {
...
 ```
+
+## Accessibility
+::a11y-intro::
+
+`ef-rating` does not assign any roles as it's not an actionable component.
+
+However, in the interactive mode, rating will be assigned `role="slider"`. Users can use arrow keys to update value. Current value of the rating will be updated to `aria-valuenow`. You can assign `aria-label` and `aria-valuetext` to `ef-rating` to give the component an assistive name for screen reader.
+
+If you need non-interactive rating to be focusable, you can set `tabindex="0"` and assigning `aria-label` value.
```
wsuwt: This does not make sense. Non-interactive should not be focusable.

packages/elements/src/rating/index.ts
```diff
@@ -42,123 +51,248 @@ export class Rating extends BasicElement {
...
    */
-  static get styles (): CSSResultGroup {
-    return css`
-      :host {
-        display: inline-block;
-      }
-    `;
+  private get StarsTemplate (): TemplateResult[] {
+    const stars = [];
+    for (let i = 0; i < this.maxNumber; i += 1) {
```
wsuwt:
```suggestion
    for (let index = 0; index < this.maxNumber; index += 1) {
```

packages/elements/src/rating/index.ts
```diff
@@ -42,123 +51,248 @@ export class Rating extends BasicElement {
...
+   * Therefore `flex: reverse` style is applied and the items are constructed in the reverse mode to mimic the correct behaviour.
+   * @returns stars template
    */
-  static get styles (): CSSResultGroup {
-    return css`
-      :host {
-        display: inline-block;
-      }
-    `;
+  private get StarsTemplate (): TemplateResult[] {
```
wsuwt:
```suggestion
  private get starsTemplate (): TemplateResult[] {
```

packages/elements/src/rating/index.ts
```diff
@@ -42,123 +51,248 @@ export class Rating extends BasicElement {
...
+    const stars = [];
+    for (let i = 0; i < this.maxNumber; i += 1) {
+      const reverseIdx = this.valueNumber - (this.maxNumber - i) + 1;
```
wsuwt:
```suggestion
      const reverseIndex = this.valueNumber - (this.maxNumber - i) + 1;
```

packages/elements/src/rating/index.ts
```diff
@@ -42,123 +51,248 @@ export class Rating extends BasicElement {
...
+      const v = reverseIdx > 0 ? Math.min(1, reverseIdx) : 0;
```
wsuwt: What is `v` ?

packages/elements/src/rating/index.ts
```diff
@@ -42,123 +51,248 @@ export class Rating extends BasicElement {
...
   /**
-   * Process click event to set the new value
-   * @param {number} index index of star
+   * Update value and fired value-changed event
+   * @param value value to updated
    * @returns {void}
    */
-  private handleTap (index: number): void {
-    if (!this.interactive) {
+  private updateValue (value: string): void {
```
wsuwt:
```suggestion
  private notifyValueChange (value: string): void {
```
Just to be consistent with the other component.

Sarin-Udompanish: updated

packages/elements/src/rating/index.ts
Sarin-Udompanish: updated

packages/elements/src/rating/index.ts
Sarin-Udompanish: updated

packages/elements/src/rating/index.ts
Sarin-Udompanish: updated

packages/elements/src/rating/index.ts
Sarin-Udompanish: updated

packages/elements/src/rating/index.ts
```diff
@@ -42,123 +51,248 @@ export class Rating extends BasicElement {
...
+   */
+  protected willUpdate (changedProperties: PropertyValues): void {
+    super.willUpdate(changedProperties);
+
+    if (changedProperties.has('interactive')) {
+      this.interactiveChanged();
     }
+    if (this.interactive) {
+      if (changedProperties.has('value')) {
+        this.setAttribute('aria-valuenow', this.value);
```
wsuwt: Can you do this in `value` setter ?

Sarin-Udompanish: updated

packages/elements/src/rating/index.ts
```diff
@@ -42,123 +51,246 @@ export class Rating extends BasicElement {
...
+
+  /**
+   * Set number of total stars
+   * @param max max value
+   * @default '5'
+   */
   @property({ type: String })
-  public max = '5';
+  public set max (max: string) {
+    const newMax = max && this.isValidValue(max) ? Math.round(Number(max)).toString() : '5';
```
wsuwt: Would 
```suggestion
    const newMax = max && this.isValidValue(max) ? Math.round(Number(max)).toString() : this.MAX_VALUE;
```
be better ?

documents/src/pages/elements/rating.md
Sarin-Udompanish: updated

packages/elements/src/rating/index.ts
Sarin-Udompanish: updated

wsuwt (dismissed):

wsuwt (approved):

PR #338, opened 2022-05-27
feat(autosuggest): accessibility support

goremikins:
## Description
Adds accessibility support. Add event types.

## Type of change

## Checklist


packages/elements/src/autosuggest/__demo__/index.html
```diff
@@ -652,5 +652,45 @@
...
+            });
+          });
+          autoSuggest.addEventListener('opened-changed', (event) => {
+            const opened = event.detail.value;
+            input.setAttribute('aria-expanded', `${opened}`);
+            if (!opened) {
+              input.removeAttribute('aria-activedescendant');
+            }
+          });
+          autoSuggest.addEventListener('item-select', (event) => {
```
wsuwt: I thought this is suppose to be `item-highlight` event, but I'm surprise `item-select` also works similarly.

goremikins: Autosuggest is not really consistent with the rest of framework, so that is one of the reasons why it needs refactoring.

goremikins: You are right, it should be highlight

goremikins: well, no, it needs to be item-select to work correctly

documents/src/pages/elements/autosuggest.md
```diff
@@ -1319,6 +1319,48 @@ Construct the `ef-item` element from the suggestion object using one-to-one mapp
...
+
+```javascript
+const inputElement = document.getElementById('input');
+const autoSuggest = document.getElementById('autosuggest');
+
+// Set `aria-expanded` based on autosuggest `opened` state
+autoSuggest.addEventListener('opened-changed', (event) => {
+  const opened = event.detail.value;
+  inputElement.setAttribute('aria-expanded', `${opened}`);
+  input.setAttribute('aria-activedescendant');
```
wsuwt:
```suggestion
```
`aria-activedescendant` without value do not have any meaning.

documents/src/pages/elements/autosuggest.md
```diff
@@ -1319,6 +1319,48 @@ Construct the `ef-item` element from the suggestion object using one-to-one mapp
...
+});
+
+// Set `aria-activedescendant` based on selected item
+autoSuggest.addEventListener('item-select', (event) => {
+  const target = event.detail.target;
+  const targetId = target ? (target.id || '') : '';
+  input.setAttribute('aria-activedescendant', targetId);
```
wsuwt: Shouldn't it be 
```suggestion
  inputElement.setAttribute('aria-activedescendant', targetId);
```
?

packages/elements/src/autosuggest/__demo__/index.html
```diff
@@ -652,5 +652,39 @@
...
+          const input = document.getElementById('accessibility-input');
+          autoSuggest.addEventListener('suggestions-fetch-requested', ({ target, detail: { query } }) => {
+            window.filterData(query).then(({ data }) => {
+              target.suggestions = data;
+            });
+          });
+          autoSuggest.addEventListener('opened-changed', (event) => {
+            const opened = event.detail.value;
+            input.setAttribute('aria-expanded', `${opened}`);
+            input.setAttribute('aria-activedescendant');
```
wsuwt: Why set empty `aria-activedescendant` ?

packages/elements/src/autosuggest/__demo__/index.html
wsuwt: What's the difference ? I tried switching to `item-highlight` and it still update id correctly.

packages/elements/src/autosuggest/__demo__/index.html
wsuwt: On closing, removing `aria-activedescendant` would be useful.

packages/elements/src/autosuggest/helpers/utils.ts
```diff
@@ -63,37 +64,45 @@ export const updateElementContent = (el: Item, query: string, label: string, val
  * @returns item
  */
 export const itemRenderer: AutosuggestRenderer = (suggestion: AutosuggestItem, query: AutosuggestQuery | null): HTMLElement => {
```
wsuwt: Would it be a good idea to separate this into `renderer.ts` like the other components ? I see that this is like a combination of both utils and renderer.

packages/elements/src/autosuggest/__demo__/index.html
goremikins: OK, will do

packages/elements/src/autosuggest/helpers/utils.ts
goremikins: With `item-highlight` it did not work correctly with VoiceOver. I think the reason is that the autosuggest text is populated on `item-select`, not `item-highlight` and the reader get confused. But as I have mention in the docs, the actual implementation might be different to our sample

documents/src/pages/elements/autosuggest.md
goremikins: done

documents/src/pages/elements/autosuggest.md
goremikins: done

packages/elements/src/autosuggest/__snapshots__/Autosuggest.md
```diff
@@ -0,0 +1,56 @@
...
+    aria-selected="false"
+    id="moreResults"
+    part="more-results"
+    tabindex="0"
+  >
+    <span part="more-results-text">
+      More results for
+      <mark>
+      </mark>
+    </span>
```
Sakchai-Refinitiv, line 27: I think we should support A111Y for `more-results-text` item, currently, it does not work on Voice Over.

wsuwt: I think probably adding `id` and localised `aria-label` would do.

Sakchai-Refinitiv: I found one more issue on `more-result` item when pressing `shift` + `enter` keys, the highlighted item will be removed. I think it should keep highlighted item and ready to go on the first item of page 2 by arrow key.

Sakchai-Refinitiv (changes requested):
Hi @goremikins, please check the items navigation on the JAWS. I found a problem when put some query e.g. `as`, `j` and then using arrow down navigate to the bottom, the focusing will be moved out of the autosuggest element.  You can reproduce it on the demo page in `default` section

packages/elements/src/autosuggest/__demo__/index.html
```diff
@@ -652,5 +652,38 @@
         })();
       </script>
     </demo-block>
+    <demo-block layout="normal" header="Accessibility" tags="accessibility">
```
Sakchai-Refinitiv: Import incorrect renderer in the compound Input demo

packages/elements/src/autosuggest/__snapshots__/Autosuggest.md
goremikins: We cannot, as currently `More result text` can be changed by the app developer via public static attribute and we are not changing any APIs in this PR. Saying so, that can be added as a separate ticket, and the correct change should be announced.

goremikins:
Hi, @Sakchai-Refinitiv . `default` section is not accessible. Please test only `Accessibility` section and I cannot reproduce the issue there.

packages/elements/src/autosuggest/__demo__/index.html
goremikins: fixed

packages/elements/src/autosuggest/__snapshots__/Autosuggest.md
goremikins: Or perhaps I can try to do something

goremikins: No, unfortunately I cannot translate this string. The issue is that `translate` does not sanitise chunks and the code may be vulnerable to XSS attack. I can use smth like `sanitize-html` library, but that it outside the current scope. Furthermore, more-search item is not pronounced anyway, as active-descendant cannot go outside the DOM scope.

packages/elements/src/autosuggest/__demo__/index.html
Sakchai-Refinitiv: I see line:446 in demo/index.html still import incorrect `import { itemRenderer } from './../../../lib/autosuggest/helpers/utils';`

Sakchai-Refinitiv: you can see an error in the console on the demo page.

goremikins: Not yet pushed, but fixed :)

packages/elements/src/autosuggest/__snapshots__/Autosuggest.md
goremikins: OK, I will use old XML solution to sanitise strings :)

packages/elements/src/autosuggest/__demo__/index.html
Sakchai-Refinitiv: OK

packages/elements/src/autosuggest/__snapshots__/Autosuggest.md
Sakchai-Refinitiv: Does it work?

goremikins: done

packages/elements/src/autosuggest/__demo__/index.html
goremikins: done

Sakchai-Refinitiv:
> Hi, @Sakchai-Refinitiv . `default` section is not accessible. Please test only `Accessibility` section and I cannot reproduce the issue there.

@goremikins Yes, I cannot reproduce the same, but I wonder why it is different?

goremikins:
> > Hi, @Sakchai-Refinitiv . `default` section is not accessible. Please test only `Accessibility` section and I cannot reproduce the issue there.
> 
> @goremikins Yes, I cannot reproduce the same, but I wonder why it is different?

Perhaps nothing says to JAWS that the opened autosuggest is a popup, as the default role of it is listbox. In accessible example, the list explicitly says that the autosuggest is a popup.

Sakchai-Refinitiv:
> > > Hi, @Sakchai-Refinitiv . `default` section is not accessible. Please test only `Accessibility` section and I cannot reproduce the issue there.
> > 
> > 
> > @goremikins Yes, I cannot reproduce the same, but I wonder why it is different?
> 
> Perhaps nothing says to JAWS that the opened autosuggest is a popup, as the default role of it is listbox. In accessible example, the list explicitly says that the autosuggest is a popup.

Yes, that is correct. I have already tested it by putting the aria attributes it works well. I think separating the A11Y example on the demo page is a good idea, we should do this pattern in other elements. 🙂

packages/elements/src/autosuggest/__snapshots__/Autosuggest.md
Sakchai-Refinitiv: > I found one more issue on `more-result` item when pressing `shift` + `enter` keys, the highlighted item will be removed. I think it should keep highlighted item and ready to go on the first item of page 2 by arrow key.

OK, this issue we will improve later.

Sakchai-Refinitiv (approved):

PR #339, opened 2022-05-30
fix(clock): a11y, tick, show-seconds should not pronounce all the time

Theeraphat-Sorasetsakul:
## Description
When clock is tick, Screen reader will announce every seconds. It causes user's action will be interrupted.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/clock/index.ts
+++ b/packages/elements/src/clock/index.ts
@@ -728,7 +728,7 @@ export class Clock extends ResponsiveElement {
       void this.updateAriaValue(true);
     }
 
-    if (changedProperties.has('sessionTicks')) {
+    if (this.interactive && changedProperties.has('sessionTicks')) {
       // Avoid announce every second that could interrupt the screen reader when the user takes an action.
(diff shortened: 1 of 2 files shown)


packages/elements/src/clock/index.ts
```diff
@@ -699,6 +689,24 @@ export class Clock extends ResponsiveElement {
...
   }
 
+  /**
+   * Handles interactive by update role, tabindex, and aria attribute
+   * @returns {void}
+   */
+  private interactiveChanged (): void {
+    if (this.interactive) {
+      this.setAttribute('role', 'spinbutton');
+      this.setAttribute('tabindex', '0');
```
Sakchai-Refinitiv: Should not force set tabindex to `0` because user can set it more than `0`. You must check it is already exist first.

Theeraphat-Sorasetsakul: Updated by check tabindex before assign

Sakchai-Refinitiv (approved):

PR #342, opened 2022-05-31
fix(utils): improve svg security

TremayneChrist:
Adds patch to strip event attributes from any SVG body


packages/utils/src/loader/cdn-loader.ts
```diff
@@ -24,6 +25,22 @@ export class CdnLoader {
...
+   * Strips any event attributes which could be used to
+   * maliciously hijack the application.
+   * @param element Element to check
+   * @returns {void}
+   */
+  private stripUnsafeAttributes (element: SVGElement): void {
+    const attributes = element.getAttributeNames();
+    for (const attribute of attributes) {
+      // Remove event attributes e.g., `onclick`
+      if (attribute.startsWith('on')) {
```
wsuwt: `IE11` does not support `String.prototype.startsWith()`.

goremikins: It should come as part of the default polyfill and automatically added by babel

wsuwt: But I think babel preset should be able to handle this in most scenarios.

TremayneChrist:
I've removed the previous code duplication and shifted logic into a separate `SVGLoader`, keeping the separate Flag/Icon loaders for preload logic.

Using @goremikins suggestion of `responseXML`, fixes the final known security flaw.

goremikins (dismissed):

wsuwt (approved):

dtanp-rft:
Bangkok team has already tried to hack the code but nobody is succeeding. I think we can merge this.

PR #347, opened 2022-06-01
feat(utils): add local storage cache

TremayneChrist:
- Adds new utility to cache data in local storage
- Uses new utility to cache svg icons and flags in local storage for use across multiple sessions


TremayneChrist:
@dtanp @wsuwt This is one option for additional caching of SVGs and could give us a quick win for speeding up multiple charts across multiple windows/sessions. Next phase would be to shift loading into a service worker to speed up network requests across multiple sessions.

packages/utils/src/cache/local-cache.ts
```diff
@@ -0,0 +1,112 @@
...
+   * Reusable task for saving database to local storage
+   * @returns {void}
+   */
+  private saveTask = () => save(this.key, this.cache);
+
+  /**
+   * Caches a value against a key to use until expired
+   * @param key Cache key
+   * @param value Data to store in cache
+   * @param [expires=604800] Cache expiry in seconds. Defaults to 5 days.
```
wsuwt:
```suggestion
   * @param [expires=432000] Cache expiry in seconds. Defaults to 5 days.
```

wsuwt:
Before https://chimerical-fairy-b7b6d1.netlify.app/, after https://regal-centaur-0619bd.netlify.app/, showing positive results on reducing network requests and improving render time with 8 iframes.

dtanp-rft:
Here is more details on the result from those two links (8 iFrames, ~500 icons on each iFrame)

### Key Observation
- localStorage cache adds a little bit slowness on cold start (not significant)
- localStorage does significant boost up the renderer/loading when icon is already cached
- localStorage cache consumes ~1KB per icons, given Chart App for example, it'll approximately consume ~150KB
- Currently, localStorage on Workspace already consuming ~3.5MB by other things. It's still fine to add up 150KB. MyFile only consume 3MB
- On cold start, known issue on Chrome is observed [link](https://stackoverflow.com/questions/24122506/neterr-insufficient-resources-error-when-adding-numerous-img-elements-to-dom). Not observable on FireFox. I remember that Chart team use to report this issue but we didn't know about this known issue on Chrome before.

### COLD
* Checked Disabled Cache
* Delete localStorage

||# Requests|Transferred|Finish|Load|
|------|----------|-----------|------|----|
|BEFORE|4500      |4MB/5MB    |10s   |6s  |
|AFTER |4500      |3MB/2.3MB  |11s   |11s |

### CACHED
* Unchecked Disabled Cache
* Icon is available in localStorage

||# Requests|Transferred|Finish|Load|
|------|----------|-----------|------|----|
|BEFORE|4500      |800B/5MB   |7s    |6s  |
|AFTER |25        |1KB/1.5MB  |3.75s |4s  |

TremayneChrist:
If this speeds up cold starts we can look to create another cache using indexdb. This will remove the worry about storage limits and we can query it more often. Then we can shift SVG caching to the indexdb version of the util.

Also, this local storage version is being quite reserved when it comes to interfacing with local storage. We can probably make the util interface more frequently and speed up the cache further.

dtanp-rft:
Sounds good to me @TremayneChrist. Let's finalize this plan during IP sprint.

TremayneChrist:
New POC for Distributed Cache #439

PR #348, opened 2022-06-02
fix(autosuggest): add default min-width to autosuggest when loading

Nantawat-Poothong:
## Description

In case autosuggest asynchronous with mask, If we didn't set default min-width for autosuggest, the result when opening autosuggest is the mask didn't show, because autosuggest width doesn't set.

- Add css min-width to autosuggest when loading
- Fix backdrop opacity

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-1845


packages/elemental-theme/src/custom-elements/ef-autosuggest.less
```diff
@@ -52,7 +53,7 @@
 
   [part=backdrop] {
     background-color: @overlay-background-color;
-    opacity: @overlay-opacity;
```
goremikins: Why did you override this?

dtanp-rft: The style looks wrong to me. Not sure if reuse overlay-opacity here is a right choice. I mean we may keep the same opacity but change background color instead if it's better solution. 

Screen Shot 2565-06-06 at 17 32 12

goremikins: It is wrong all together: being the loader too large and the background too grey. Perhaps we need a proper solution from Design/UX on how to cover this case (and all other cases with asynchronous popup loading). This code was just migrated from v3 where it made more sense.

goremikins: maybe UX can say that entire screen needs to be greyed out, so we can deprecated autosuggest loading :)

dtanp-rft: I'll go with this small improvement. It's better than today anyway. You know we would need to wait very long until they take an action. It means that users won't see any improvement until then :)

Nantawat-Poothong: Updated!, I have revert override backdrop opacity :)

Sarin-Udompanish (approved):

PR #358, opened 2022-06-10
fix(tree-select): show unselected item on selection filter

Sarin-Udompanish:
## Description
There are 2 issues with no-relation mode.
1. Un-Selected Values are also shown in selected tab in no-relation mode.
( when parent item is checked but child item is unchecked it will shown all child of that parent item. )
2. Selection filter button is disabled when checked only on parent item.

Fixes # (issue)

## Checklist


packages/elements/src/tree-select/index.ts
```diff
@@ -423,7 +423,7 @@ export class TreeSelect extends ComboBox<TreeSelectDataItem> {
     };
     this.queryItems((item, composer): boolean => {
       const hasChildren = composer.getItemChildren(item);
-      if (hasChildren.length) {
+      if (hasChildren.length && this.mode === TreeManagerMode.RELATIONAL) {
```
Sakchai-Refinitiv: This change makes the values of `expanded` and `expandable` in the `memo` incorrect, I think you can fix this bug by separating `else if` to the new `if` statement.

Sarin-Udompanish: updated

packages/elements/src/tree-select/__test__/tree-select.filter.test.js
```diff
@@ -197,6 +198,22 @@ describe('tree-select/Filter', () => {
...
+    });    
+    
+    it('Should not show unselected items when clicking on selected button in no-relation mode', async () => {
+      const el = await fixture('<ef-tree-select opened no-relation></ef-tree-select>');
+      el.data = noRelationData;
+      el.selectedClickHandler();
+      expect(el.treeManager.visibleItems.length).to.equal(el.treeManager.checkedItems.length);
+    });
+
+    it('Should has correct selected amount when checking on parent item in no-relation mode', async () => {
```
wsuwt:
```suggestion
    it('Should have correct selected amount when checking on parent item in no-relation mode', async () => {
```

Sakchai-Refinitiv (approved):

PR #361, opened 2022-06-13
fix(tree): tree query selection issues

Nantawat-Poothong:
## Description
Tree fire incorrect item in `value-changed` event, when using tree `query` property.

The problem is when filtering items, the exclude item will be locked and can't perform any action.

This PR made 2 changes
1. Only lock the item in single selection mode.
2. Re-filter items when selection mode change.

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-1851

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/tree/__test__/tree.test.js
+++ b/packages/elements/src/tree/__test__/tree.test.js
@@ -490,7 +490,7 @@ describe('tree/Tree', () => {
       descendants.forEach(item => expect(el.manager.isItemHidden(item)).to.equal(false, 'Descendants of matched items must be included'));
     });
 
-    it('Text filter applied, single selection', async () => {
+    it('Should be able to select value after filter is applied', async () => {
       const el = await fixture('<ef-tree></ef-tree>');
       el.data = flatData;
       await elementUpdated(el);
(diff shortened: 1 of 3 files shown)


packages/elements/src/tree/__test__/tree.test.js
```diff
@@ -490,6 +490,50 @@ describe('tree/Tree', () => {
       descendants.forEach(item => expect(el.manager.isItemHidden(item)).to.equal(false, 'Descendants of matched items must be included'));
     });
 
+    it('Text filter applied, single selection', async () => {
```
wsuwt: I would update the test case name into something like 
```suggestion
    it('Should be able to select value after filter is applied', async () => {
```

packages/elements/src/tree/managers/tree-manager.ts
```diff
@@ -239,11 +239,12 @@ export class TreeManager<T extends TreeDataItem> {
   /**
    * Excludes an item as part of the tree.
    * @param item Item to exclude
+   * @param lock Item lock
```
wsuwt: It would be great to explain why we need this extra parameter.

Nantawat-Poothong: Thank you, updated!

goremikins (changes requested):
This fix does not look right to me. It needs to be discussed in Tech session.

PR #375, opened 2022-06-22
fix(core): fix notice to support type

Theeraphat-Sorasetsakul:
## Description
**Steps to reproduce** :
Use WarningNotice helper in an element.
**Actual result** :
Header of notice shows "Information notice: message"
**Expected result** :
Header of notice should be "Warning notice: message"

Fixes # ([ELF-1842](https://jira.refinitiv.com/browse/ELF-1842))

## Type of change

## Checklist


Nantawat-Poothong:
I think Deprecation and Warning notice syntax should be the same.

Theeraphat-Sorasetsakul:
> I think Deprecation and Warning notice syntax should be the same.

Agree. I updated both to be same.

packages/core/src/notices/Notice.ts
```diff
@@ -28,9 +23,10 @@ export class Notice {
    * Create a warning notice to show in the console.
    * @param message Warning message to show in the console
    * @param supportURL Support URL to show additional information
+   * @param type Type of Notice to show at top of message
    */
-  constructor (message: string, supportURL?: string) {
-    this.message = generateMessage(this.type, message, supportURL);
+  constructor (message: string, supportURL?: string, type = 'Information') {
```
wsuwt: I think it would be better if message types were enum, so that it can also be re-use e.g.,

`
export enum MESSAGE_TYPE {
  NOTICE = 'Information',
  DEPRECATION = 'Deprecation',
  WARNING = 'Warning'
}
`

Theeraphat-Sorasetsakul: Updated

wsuwt (approved):

PR #376, opened 2022-06-23
style: add `eslint-plugin-import` and extensions to import paths

wsuwt:
## Description

Adding `eslint-plugin-import` plugin to enforce adding extension to all internal import paths.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -43,14 +43,18 @@ export { DeprecationNotice } from './notices/DeprecationNotice.js';
  * Export events
  */
 export { TapEvent } from './events/TapEvent.js';
-export type { FocusedChangedEvent } from './types/events.js';
+import type { FocusedChangedEvent } from './types/events';
+export type { FocusedChangedEvent };
 
 /**
  * Export common interfaces
  */
-export type { MultiValue } from './interfaces/MultiValue.js';
-export type { StyleMap } from './interfaces/StyleMap.js';
-export type { StyleInfo } from './interfaces/StyleInfo.js';
+import type { MultiValue } from './interfaces/MultiValue';
+export type { MultiValue };
+import type { StyleMap } from './interfaces/StyleMap';
+export type { StyleMap };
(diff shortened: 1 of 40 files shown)


wsuwt:
cc. @goremikins @TremayneChrist

packages/core/src/index.ts
```diff
@@ -43,14 +43,14 @@ export { DeprecationNotice } from './notices/DeprecationNotice.js';
...
 export { TapEvent } from './events/TapEvent.js';
-export type { FocusedChangedEvent } from './types/events';
+export type { FocusedChangedEvent } from './types/events.js';
 
 /**
  * Export common interfaces
  */
-export type { MultiValue } from './interfaces/MultiValue';
-export type { StyleMap } from './interfaces/StyleMap';
-export type { StyleInfo } from './interfaces/StyleInfo';
```
Sakchai-Refinitiv, line 53: @goremikins I remember you had mentioned the js extension for import `type` when we migrated Lit v2, and what is the problem why we should not add it. Can you tell me again?

goremikins: `js` files are consumed by browser, while types by TypeScript compiler. Therefore it does not make logical sense to include `js` extension. Another reason, consider these scenario:
```
// types.ts
type A = {};
export A;
```
when `types.ts` is complied, `types.js` would look like:
```
export {};
```

wsuwt: You are right. When you click on `types.js` import path it will still link to `types.d.ts` correctly because names are the same, but I think it might be possible to ignore `import type` specifically.

wsuwt:
Ok, `type import` should ignore `extension` rule by default, see https://github.com/import-js/eslint-plugin-import/pull/2270. However, `export type` is not ignored which is a bug and its PR in in progress https://github.com/import-js/eslint-plugin-import/pull/2436. I still think this plugin would be beneficial to our workflow. There are more rules that we can enforce on imports and exports.

Sakchai-Refinitiv (approved):

PR #377, opened 2022-06-23
fix: change Halo theme font-weight 600 from bold to semibold

dtanp-rft:
## Description
In Halo theme, font-weight 600 should be semibold. However, EF theme map 500 to semibold and 700 to bold. This means that font-weight 600 will be bold font.

We should map 600 to semibold and keep 500 to semibold as well so it won't have any impacts that using 500 with semibold.

In theme, we also want to replace anywhere that use font-weight 500 to 600 so it will be clean and match to Figma file.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/packages/halo-theme/src/fonts.less
+++ b/packages/halo-theme/src/fonts.less
@@ -10,7 +10,7 @@
 @font-face {
   font-family: 'Proxima Nova Fin';
   font-style: normal;
-  font-weight: 500;
+  font-weight: 500 600;
   font-display: swap;
   src: url('https://cdn.refinitiv.net/public/libs/elf/assets/elf-theme-halo/resources/fonts/proximanovafin/proximanovafin-semibold.woff2') format('woff2'),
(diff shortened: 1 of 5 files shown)


packages/halo-theme/src/fonts.less
```diff
@@ -10,7 +10,7 @@
 @font-face {
   font-family: 'Proxima Nova Fin';
   font-style: normal;
-  font-weight: 500;
+  font-weight: 500 600;
```
Nantawat-Poothong, line 13: I think it's a bit odd with this line.

dtanp-rft: I see that it supports multiple value. https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-weight

Nantawat-Poothong: Too weird, I didn't see this syntax before, and when I try with my browser, e.g. Edge Chrome and Safari. It shows the same warning like 'invalid property value'.

dtanp-rft: as we discussed, it's `font-weight` in @font-face. It's not font-weight that you can set to element.

Sarin-Udompanish (approved):

PR #380, opened 2022-06-27
docs: updated incorrect import path on some pages

dtanp-rft:
Incorrect import path on some pages.

### DIFF
--- a/documents/src/pages/elements/list.md
+++ b/documents/src/pages/elements/list.md
@@ -85,13 +85,13 @@ Extending the default renderer is the easiest way to display custom content, whi
 !> Renders are currently being upgraded and should only be used for testing purposes.
 
 ```javascript
-import { List, DefaultRenderer } from '@refinitiv-ui/elements/list';
+import { ListRenderer } from '@refinitiv-ui/elements/list';
 
 const el = document.querySelector('ef-list');
 const defaultRenderer = new DefaultRenderer(el);
 
 el.renderer = (item, composer, element) => {
-  const el = defaultRenderer.apply(this, arguments);
+  const el = defaultRenderer(item, composer, element);
   // do something extra
   return el;
 };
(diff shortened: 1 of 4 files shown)


documents/src/pages/elements/notification.md
```diff
@@ -32,7 +32,7 @@ The notification bar is used to show informative content when something happens
 These methods work in a similar way to those provided by `window.console`.
 
 ```javascript
-import { info, confirm, warn, error } from '@refinitiv-ui/notification/helpers';
+import { info, confirm, warn, error } from '@refinitiv-ui/elements/notification';
```
Nantawat-Poothong, line 33: I had tried to implement this import with example code for notification, and I seems not to work for me.
It always throw error "Illegal constructor"

documents/src/pages/elements/list.md
```diff
@@ -85,13 +85,13 @@ Extending the default renderer is the easiest way to display custom content, whi
 !> Renders are currently being upgraded and should only be used for testing purposes.
 
 ```javascript
-import { List, DefaultRenderer } from '@refinitiv-ui/elements/list';
+import { ListRenderer } from '@refinitiv-ui/elements/list';
 
 const el = document.querySelector('ef-list');
 const defaultRenderer = new DefaultRenderer(el);
```
dtanp-rft: @Nantawat-Poothong  this should be updated as well. Don't know if `const defaultRenderer` is a good choice for the name. It might be better with something like `const itemRenderer`.

Nantawat-Poothong: Sounds great, I found combo-box has variable name defaultRenderer. Should we change it to itemRenderer too?

dtanp-rft: I'd say yes. defaultRenderer could be confusing as this is not actually the default renderer.

Nantawat-Poothong: I have created PR; you can take a look here https://github.com/Refinitiv/refinitiv-ui/pull/381

Nantawat-Poothong (approved):

PR #394, opened 2022-07-12
feat(item): show tooltip when text ellipsis

Nantawat-Poothong:
## Description
Show tooltip when item label is ellipsis.

Solution
- register tooltip when item has created

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-1712

## Type of change

## Checklist


packages/elements/src/item/index.ts
```diff
@@ -260,6 +261,18 @@ export class Item extends ControlElement {
...
 
+  /**
+   * Called after the component is first rendered
+   * @param changedProperties Properties which have changed
+   * @returns {void}
+   */
+  protected firstUpdated (changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    if (this.labelEl) {
+      registerOverflowTooltip(this.labelEl, () => this.textContent, () => this.isTruncated);
```
goremikins: This will not work if label has been added after the element initialised

Nantawat-Poothong: Good point!, updated.

packages/elements/src/item/index.ts
```diff
@@ -187,6 +188,16 @@ export class Item extends ControlElement {
...
   }
 
+  /**
+   * Called after the component is first rendered
+   * @param changedProperties Properties which have changed
+   * @returns {void}
+   */
+  protected firstUpdated (changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    registerOverflowTooltip(this, () => this.textContent, () => this.isTruncated);
```
goremikins: And why `isTruncated` is a public method? I think we can remove the method and use `isElementOverflown`, similar to text-field

Nantawat-Poothong: `isTruncated` is a read-only property in item since v4. I can use `isElementOverflown` instead, but I'm not sure if it's ok to remove `isTruncated`?

goremikins: remove it

Nantawat-Poothong: Cool! updated! 😆

packages/elements/src/item/index.ts
```diff
@@ -187,6 +189,16 @@ export class Item extends ControlElement {
...
   }
 
+  /**
+   * Called after the component is first rendered
+   * @param changedProperties Properties which have changed
+   * @returns {void}
+   */
+  protected firstUpdated (changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    registerOverflowTooltip(this, () => this.textContent, () => this.labelEl ? isElementOverflown(this.labelEl) : false);
```
TremayneChrist: We need to check if the host element already has a tooltip and show that as priority

TremayneChrist (changes requested):

packages/elements/src/item/index.ts
goremikins: `registerOverflowTooltip` method does that by default

Nantawat-Poothong: Hi guy, thank for review. I still confused, should I check if the host element already has a tooltip, or shouldn't check because `registerOverflowTooltip` method does that by default.

goremikins: `() => this.textContent` covers only `textContent` case. The item can be populated via the `label` as well. This case is not covered.

goremikins: if the `ef-item` has `title` attribute, it will be taken as priority. @TremayneChrist , did you mean that?

Nantawat-Poothong: > `() => this.textContent` covers only `textContent` case. The item can be populated via the `label` as well. This case is not covered.

Good point @goremikins, I didn't think about it before.
Should I add more renderer conditions like this?
```suggestion
    registerOverflowTooltip(this, () => this.textContent || this.label, () => this.labelEl ? isElementOverflown(this.labelEl) : false);
```

Nantawat-Poothong: As for these labels, I doubt about sub-label text ellipsis, should I consider this too?, if that is so, I should update the overflow condition too, right?. Or should I register a tooltip for a sub-label either.

goremikins: I think it should support subLabel as well

TremayneChrist: > if the `ef-item` has `title` attribute, it will be taken as priority. @TremayneChrist , did you mean that?

Yes. Thanks for clarification 👍

Nantawat-Poothong:
Hi @goremikins , From the previously discussed about tooltip should support label and subLabel as well.
I have already updated, and still need to write a few test cases to pass the Test Coverage.
BTW, what do you think of this update?

packages/elements/src/item/index.ts
```diff
@@ -131,6 +133,9 @@ export class Item extends ControlElement {
   @query('#label')
   private labelEl?: HTMLElement;
 
+  @query('[part=sub-label]')
```
goremikins: Please use `ref` instead

packages/elements/src/item/index.ts
```diff
@@ -204,6 +219,44 @@ export class Item extends ControlElement {
...
+    }
+    else {
+      return this.textContent || '';
+    }
+  }
+
+  /**
+   * Get element overflown
+   * @param element Target element
+   * @returns {boolean} Return true if element is overflown.
```
goremikins: We do not put types into comments. Types are only relevant for pure JavaScript code. IDE can resolve TypeScript types automatically.

packages/elements/src/item/index.ts
```diff
@@ -204,6 +219,44 @@ export class Item extends ControlElement {
...
+      if (this.label) {
+        text += this.label;
+      }
+      if (this.subLabel) {
+        text += ` (${this.subLabel})`;
+      }
+      return text;
+    }
+    else {
+      return this.textContent || '';
```
goremikins: `this.textContent` includes left and right slot content. You want to include label and sublabel only

packages/elements/src/item/index.ts
```diff
@@ -204,6 +219,44 @@ export class Item extends ControlElement {
     }
   }
 
+
+  /**
+   * Get Item content
+   * @returns return item content from slot or label and sub-label
+   */
+  private getItemContent (): string {
+    if (this.isSlotEmpty) {
```
goremikins, line 234: Why not to play around `render` function to simplify the logic? For instance:

```
<div part="center" id="label">
        <div part="label" ref="....">
          ${ isSlotEmpty ? this.labelTemplate : undefined }
          <slot @slotchange="${this.checkSlotChildren}"></slot>
        </div>
        ${ isSlotEmpty ? this.subLabelTemplate : undefined }
      </div>
```

and the you just get textContent form labelRef and subLabelRef.

goremikins (changes requested):

packages/elements/src/item/__test__/item.test.js
```diff
@@ -1,8 +1,33 @@
-import { elementUpdated, expect, fixture, nextFrame } from '@refinitiv-ui/test-helpers';
```
goremikins, line 1: Do you really need to add the real test? Essentially you create tight coupling between `item` and `tooltip`. Whenever tooltip changes, this test would stop working.

Maybe it is better to test just the functionality  of `getItemContent` and `isItemOverflown` functions? You can make those protected btw.

packages/elements/src/item/index.ts
Nantawat-Poothong: Updated.

packages/elements/src/item/index.ts
Nantawat-Poothong: Updated.

packages/elements/src/item/index.ts
Nantawat-Poothong: Updated.

packages/elements/src/item/__test__/item.test.js
Nantawat-Poothong: Updated.

packages/elements/src/item/index.ts
TremayneChrist: @Nantawat-Poothong can you comment on this?

packages/elements/src/item/index.ts
```diff
@@ -272,7 +326,7 @@ export class Item extends ControlElement {
         ${this.multipleTemplate}
         <slot name="left"></slot>
       </div>
-      <div part="center" id="label">
+      <div part="center" id="label" ${ref(this.labelRef)}>
```
TremayneChrist: Remove the element id now that you are using ref

packages/elements/src/item/index.ts
```diff
@@ -225,6 +280,14 @@ export class Item extends ControlElement {
     return this.label && this.isSlotEmpty ? html`${this.label}` : undefined;
   }
 
+  /**
+   * Get label template if it is defined and no slot content present
+   */
+  private get slotContent (): string {
+    const slot = this.labelRef.value?.querySelector('[part=center] slot') as HTMLSlotElement;
```
TremayneChrist: Can you add a ref to the slot instead?

packages/elements/src/item/index.ts
Nantawat-Poothong: @TremayneChrist Sorry for the late comment, yes I think it's possible.

Nantawat-Poothong: Maybe we could simplify logic like this?
```javascript
<div part="center" id="label">
  <div part="label" ref="....">
    ${ isSlotEmpty && this.label ? this.labelTemplate : undefined }
    <slot @slotchange="${this.checkSlotChildren}"></slot>
  </div>
  ${ isSlotEmpty && this.subLabel ? this.subLabelTemplate : undefined }
</div>
````

File: packages/elements/src/item/index.ts (hunk 9 above)
Comment (Nantawat-Poothong) [reply packages/elements/src/item/index.ts]: Updated.

File: packages/elements/src/item/index.ts (hunk 10 above)
Comment (Nantawat-Poothong) [reply packages/elements/src/item/index.ts]: Updated.

File: packages/elements/src/item/index.ts (hunk 7 above)
Comment (TremayneChrist) [reply packages/elements/src/item/index.ts:234]: Yeah let's try

Comment (Nantawat-Poothong) [reply packages/elements/src/item/index.ts:234]: Updated.

File: packages/elements/src/item/index.ts [hunk 11]
```diff
@@ -126,10 +127,20 @@ export class Item extends ControlElement {
   public for: string | null = null;
 
   /**
-   * Cache label element
+   * Reference to the label element
+   */
+
+  private labelRef: Ref<HTMLDivElement> = createRef();
+
```
TremayneChrist, line 133:
```suggestion
```

packages/elements/src/item/index.ts
```diff
@@ -126,10 +127,20 @@ export class Item extends ControlElement {
...
+   * Reference to the subLabel element
    */
-  @query('#label')
-  private labelEl?: HTMLElement;
+  private subLabelRef: Ref<HTMLDivElement> = createRef();
+
+  /**
+   * Reference to the subLabel element
+   */
+  private slotRef: Ref<HTMLDivElement> = createRef();
```
TremayneChrist: Copy/Paste issue
```suggestion
  /**
   * Reference to the slot element
   */
  private slotRef: Ref<HTMLSlotElement> = createRef();
```

packages/elements/src/item/index.ts
```diff
@@ -215,14 +275,22 @@ export class Item extends ControlElement {
...
   private get labelTemplate (): TemplateResult | undefined {
-    return this.label && this.isSlotEmpty ? html`${this.label}` : undefined;
+    return html`${this.label}`;
+  }
+
+  /**
+   * Get slot content
+   */
+  private get slotContent (): string {
+    const slot = this.slotRef.value as unknown as HTMLSlotElement;
```
TremayneChrist: Remove casting after fixing type issue above

packages/elements/src/item/index.ts
Nantawat-Poothong: I'm not sure I understand this suggest, Do you mean I should remove labelRef?

packages/elements/src/item/index.ts
Nantawat-Poothong: Updated.

packages/elements/src/item/index.ts
Nantawat-Poothong: Thank you, updated.

packages/elements/src/item/index.ts
TremayneChrist: There's an empty line under the jsdoc comment

Nantawat-Poothong: Thanks, updated.

packages/elements/src/item/index.ts
```diff
@@ -215,14 +274,22 @@ export class Item extends ControlElement {
...
-    return this.label && this.isSlotEmpty ? html`${this.label}` : undefined;
+    return html`${this.label}`;
+  }
+
+  /**
+   * Get slot content
+   */
+  private get slotContent (): string {
+    const slot = this.slotRef.value as HTMLSlotElement;
+    return slot.assignedNodes().map(e => e.textContent).join(' ').trim();
```
TremayneChrist: You're still casting type here. We should make sure the slot is available. Also, make sure we're using full names for properties in source. `e` is usually used for events and shouldn't be used for nodes. Make sure we're describing what type of object we're working with.
```suggestion
    const nodes = this.slotRef.value?.assignedNodes() || [];
    return nodes.map(node => node.textContent).join(' ').trim();
```

Nantawat-Poothong: So embarrassed to me 🫣, Thanks! Trem.

Nantawat-Poothong: Updated.

TremayneChrist (approved):

PR #406, opened 2022-07-20
Add navigation grid documentation

goremikins:
## Description

Navigation grid documentation.

## Type of change

- [X] Documentation


TremayneChrist:
Would be useful to have an example showing how this can be paired up with `tabindex` and arrow keys.

goremikins:
@TremayneChrist , I would agree, but the sample is rather big. Maybe I can add it at the end of the document...

goremikins:
@TremayneChrist , I have added a sample

Nantawat-Poothong:
I think it would be great if we had sample code to show how to integrate this with our elements. Or may be it not necessary, what do you think?

goremikins:
@Nantawat-Poothong , maybe. To make it "Our Element" we would need to wrap that code inside `BasicElement`.

documents/src/pages/utils/navigation-grid.md
```diff
@@ -5,8 +5,443 @@ type: page
...
+
+## Types
+
+The list of _TypeScript_ types.
+
+| Name | Description | Value |
+| --- | --- | --- |
+| CellIndex | Cell index as `[Column Index, Row Index]`. _index_ is 0-based. | `[number, number]` |
+| NavigationGrid | A list of `NavigationRow` | `NavigationRow[]` |
+| NavigationRow | A list of _0_ and _1_. _0_ is an inactive cell, but _1_ is active. | `(0 &#124; 1)[]` |
```
Nantawat-Poothong:
```suggestion
| NavigationRow | A list of _0_ and _1_. _0_ is an inactive cell, but _1_ is active. | `(0 | 1)[]` |
```
I think we can use "|" directly, when it wraps with backticks.

goremikins: maybe, my IDE built-in editor does not like it

goremikins:
@TremayneChrist , are you happy with this one?

TremayneChrist (approved):

PR #408, opened 2022-07-21
feat(datetime-picker): replace input with datetimeField

Sarin-Udompanish:
## Description
Integrate datetime-field with datetime-picker

## Tasks
- [x] Code
- [x] Theme
- [x] Accessibility
- [x] Testing
- [x] Documentation

## Type of change

## Checklist


packages/elemental-theme/src/custom-elements/ef-datetime-picker.less
```diff
@@ -48,7 +65,7 @@
   }
 
   &[range] {
-    ef-text-field {
+    ef-datetime-field {
```
wsuwt, line 68: Would it be better if we style this with part selector ?

packages/elements/src/datetime-picker/index.ts
```diff
@@ -6,80 +6,60 @@ import {
...
-  getLocale,
-  TranslatePropertyKey
+  TranslateDirective
 } from '@refinitiv-ui/translate';
 import {
-  getDateFNSLocale
-} from './locales.js';
-import inputFormat from 'date-fns/esm/format/index.js';
-import inputParse from 'date-fns/esm/parse/index.js';
-import isValid from 'date-fns/esm/isValid/index.js';
```
wsuwt: We should be able to remove `date-fns` from `package.json`.

packages/elemental-theme/src/custom-elements/ef-datetime-picker.less
TremayneChrist: Yes, it should follow other elements and use part selector.

wattachai-lseg:
Due to the following concerns, this change would be revisit in the next major release instead.
* UX need some improvements
* There some remaining screen reader issues
* unclear direction on localization support e.g. date format
* Pontential breaking changes

Nantawat-Poothong:
Happy Anniversary!

PR #412, opened 2022-07-25
feat(utils): add local cache

Theeraphat-Sorasetsakul:


Theeraphat-Sorasetsakul:
### Test result from commit [ca17fa7](https://github.com/Refinitiv/refinitiv-ui/pull/412/commits/ca17fa7a69386a14427e4695240449f555cf0063)
After using indexedDB the result is faster than expected, it finishes in 1.58 seconds.

### Environment
Each app has 8 iframes. Each iframe contains 583 icons. 1 iframe has 586 requests that consist of html, js, and theme font. Test repo is [link](https://github.com/Theeraphat-Sorasetsakul/test-icon-idb).

### Step to test
Open:
1. Open incognito browser on Edge
2. Open inspect
3. Tick disabled cache
4. Enter URL

Refresh:
1. Untick disabled cache
2. Press f5 to refresh

#### V6
The current published version.
url: https://chimerical-zuccutto-a8bf13.netlify.app/
  | Requests | Transferred | resources | Finish | Load
-- | -- | -- | -- | -- | --
Open | 4690 | 3.7 MB | 2.4 MB | 8.46 s | 7.92 s
Refresh | 4689 | 4.3 kB | 2.5 MB | 7.52 s | 7.78 s

#### IndexedDB
Use utils from this pull request. The rest is same as v6.
url: https://melodic-boba-222258.netlify.app/
  | Requests | Transferred | resources | Finish | Load
-- | -- | -- | -- | -- | --
Open | 4690 | 3.0 MB | 2.2 MB | 8.94 s | 5.73 s
Refresh | 25 | 4.3 kB | 1.3 MB | 1.58 s | 1.68 s

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,136 @@
...
+| --- | --- |
+| String \| null | string data or `null` if the key isn't cached |
+
+### remove
+Remove cache data value based on provided key
+
+```typescript
+import { LocalCache } from '@refinitiv-ui/utils/cache.js';
+const key = 'my-item-01';
+const cache = new LocalCache();
```
Sakchai-Refinitiv: There are many code example that creates `LocalCache` without storage, I think it might be misleading when users read the code.

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,136 @@
...
+| --- | --- | --- |
+| key | String | Cache key |
+| value | unknown | Data to store in cache |
+| expires | Number | Cache expiry in seconds. Defaults to 5 days. |
+
+#### Returns
+
+| Type | Description |
+| --- | --- |
+| String | The new value with the months added |
```
Nantawat-Poothong: I think the description is incorrect.

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,136 @@
...
+```text
+cache.set(key, value, expires);
+```
+
+#### Arguments
+
+| Name | Type | Description |
+| --- | --- | --- |
+| key | String | Cache key |
+| value | unknown | Data to store in cache |
```
Nantawat-Poothong: I'm not sure an unknown type is correct. Should it be Any or something else?

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,161 @@
...
+      return;
+    }
+
+    if (!isFirefox()) {
+      const databases = await window.indexedDB.databases();
+      const found = databases.find(database => database.name === this.dbName);
+      if (found && this.version < (found.version || 0)) {
+        throw this.errorMessage(`Your version (${this.version}) is less than the existing version (${String(found.version)}).`);
+      }
+    }
```
TremayneChrist: Remove this once we change the constructer to support name only.

packages/utils/src/cache/local-cache.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Stores data in a local cache that can be specified to be stored in localstorage or indexedDB.
+ */
+export class LocalCache {
+
+  /**
+   * Cache ready to use
+   */
+  protected ready: Promise<boolean> | null = null;
```
TremayneChrist: Remove all ready states from local cache. This should just be a lightweight wrapper and does not need to manage ready states.

packages/utils/src/cache/local-cache.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Storage to store data
+   */
+  protected storage!: CacheStorage;
+
+  /**
+   * Internal cache object
+   */
+  protected cache: CacheMap | null | undefined;
```
TremayneChrist: Remove the cache and just use the storage directly. This file is just a wrapper and proxy to the storage.

packages/utils/src/cache/local-cache.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+   * Set storage to the local cache
+   * @param storage cache storage
+   * @returns {void}
+   */
+  protected use (storage: CacheStorage): void {
+    this.storage = storage;
+    this.ready = this.restore();
```
TremayneChrist: Restore the cache when the storage instance is created.

packages/utils/src/cache/local-cache.ts
```diff
@@ -0,0 +1,123 @@
...
+      this.cache = await this.storage.restoreItems() as CacheMap;
+      return true;
+    }
+    catch (e) { // Keep it work. Even if can't connect to storage
+      this.cache = new Map();
+      // eslint-disable-next-line no-console
+      console.error(e);
+      return false;
+    }
+  }
```
TremayneChrist: Remove this and restore when the storage object is created.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: This is void. It isn't return anything. I will remove it.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: It should be unknown type. Because we don't know what item will be store.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated.

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * Constructor
+   * @param info Storage config
+   */
+  constructor (info: CacheIndexedDBStorageConfig) {
+    const { dbName, version, storeName } = info;
+    this.dbName = dbName;
+    this.version = version;
+    this.storeName = storeName as never;
+  }
```
TremayneChrist: It looks like we can simplify this and just use a name, keeping it consistent with local storage.

packages/utils/src/cache/cache-localstorage.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Constructor
+   * @param prefixKey prefix key for all item
+   */
+  constructor (prefixKey?: string) {
+    if (prefixKey) {
+      this.prefixKey = prefixKey;
+    }
+  }
```
TremayneChrist: There should always be a prefix key. We don't want to clash with any other local storage objects. The constructor should always take a name - it's like a name of database like in IndexedDB storage.

packages/utils/src/cache/cache-indexeddb.ts
TremayneChrist: We should probably prefix the name so that we don't clash with any other databases. We can share the same prefix across all storage classes.

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Set a item against a key to this storage
+   * @param key item key
+   * @param value item value
+   * @returns {void}
+   */
+  public async setItem (key: string, value: CacheItem): Promise<void> {
+    !this.db && await this.open();
```
Nantawat-Poothong: I see `!this.db && await this.open();` everywhere, could you create a new method for this and call it instead?

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  /**
+   * Remove a item against a key to this storage
+   * @param key item key
+   * @returns {void}
+   */
+  public async removeItem (key: string): Promise<void> {
+    !this.db && await this.open();
+    await this.db?.delete(this.storeName, key);
```
Nantawat-Poothong: I'm not sure what will happen if call method `removeItem` when key is not exist in store.
I think it will be better if method `removeItem`  can return true or false, instead of void.
in case when user need return value to handle next, what do you think?

packages/utils/src/loader/svg-loader.ts
```diff
@@ -60,8 +67,19 @@ const isValidResponse = (response: XMLHttpRequest | undefined): response is XMLH
...
   if (isValidResponse(response) && response.responseXML) {
-    const svgDocument = response.responseXML.cloneNode(true) as Document;
-    const svg = svgDocument.firstElementChild;
+    const svgDocument = response.responseXML;
+    /**
+     * Getting the first node of response that is an element,
+     * just in case when firstChild isn't SVG.
+     * ( XMLDOM `cloneNode()` and `firstElementChild` are not supported in IE11 )
+     */
+    let svg;
```
Nantawat-Poothong: I'm not sure, can we declare svg variable like this?
`const svg = svgDocument.childNodes.findLast(element => element.nodeType === Node.ELEMENT_NODE);`

Theeraphat-Sorasetsakul: Not possible. childNodes isn't type array.

packages/utils/src/cache/types.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Cache storage item
+ */
+export interface CacheItem {
```
TremayneChrist: Split interfaces into separate files

packages/utils/src/loader/svg-loader.ts
```diff
@@ -98,8 +120,20 @@ export class SVGLoader extends CDNLoader {
...
     const src = await this.getSrc(name);
-    const response = await this.load(src);
-    return extractSafeSVG(response)?.outerHTML;
+
+    const cacheItem = await cache.get(src);
+    if (cacheItem === null) {
+      const response = await this.load(src);
+      const svgNode = extractSafeSVG(response)?.cloneNode(true);
+      const svgBody = svgNode ? this.xmlSerializer.serializeToString(svgNode) : undefined;
+      // console.log(svgBody);
```
TremayneChrist: Remove

TremayneChrist (changes requested):
As we discussed on the call, let's streamline and align the APIs of the storage objects.

- Shift storage logic out of local cache class
- Align storage class APIs
- Storage constructor to take only `name`

With these changes we can also remove the need to create a storage object first and provide it to the local cache. This would simplify the usage of local cache.

For example:
```ts
const cache = new LocalCache('svg-loader', { storage: 'indexeddb' });
```

packages/elements/package.json
```diff
@@ -334,6 +334,7 @@
     "chart.js": "~2.9.4",
     "d3-interpolate": "^3.0.1",
     "date-fns": "^2.22.1",
+    "idb": "^7.0.2",
```
goremikins: Should this be added to `utils` package.json?

packages/utils/src/cache/cache-localstorage.ts
```diff
@@ -0,0 +1,103 @@
...
+   * @returns Promise CacheMap items
+   */
+  public async restoreItems (): Promise<CacheMap> {
+    const prefixKey = this.prefixKey !== '' ? this.prefixKey + '-' : '';
+    const regex = new RegExp(`^${prefixKey}`);
+
+    const items = new Map() as CacheMap;
+    const keys = Object.keys(localStorage).filter(key => key.startsWith(prefixKey));
+
+    for (let i = 0; i < keys.length; i++) {
```
goremikins: `i += 1`

packages/utils/src/loader/svg-loader.ts
goremikins: can you just do: `const svg = svgDocument.children[svgDocument.children.length - 1]`?

packages/utils/src/loader/svg-loader.ts
```diff
@@ -1,5 +1,12 @@
+import { CacheIndexedDBStorage, LocalCache } from '../cache.js';
 import { CDNLoader } from './cdn-loader.js';
 
+// Storage
+const storage = new CacheIndexedDBStorage({ dbName: 'ef', version: 1, storeName: 'svg-loader' });
```
goremikins: I may suggest to use more unique dbName to avoid any potential clash

packages/utils/src/cache/cache-indexeddb.ts
Theeraphat-Sorasetsakul: Due to [Trem feedback](https://github.com/Refinitiv/refinitiv-ui/pull/412#pullrequestreview-1066797912), ready logic should move to each storage class. So the line will be remove and use this.ready instead.

Theeraphat-Sorasetsakul:
Note: Indexeddb in Firefox private mode isn't work. https://bugzilla.mozilla.org/show_bug.cgi?id=781982

packages/utils/src/cache/cache-localstorage.ts
Theeraphat-Sorasetsakul: Updated

packages/utils/src/loader/svg-loader.ts
Theeraphat-Sorasetsakul: Due to [Trem feedback](https://github.com/Refinitiv/refinitiv-ui/pull/412#pullrequestreview-1066797912), the storage will receive only name. That name will be combine with shared static prefix for both database name and store name. Version will be fit to 1.  So those concept will avoid redundant database and avoid indexeddb clash.

packages/elements/package.json
Theeraphat-Sorasetsakul: Updated. And also removed from @refinitiv-ui/elements.

packages/utils/src/loader/svg-loader.ts
Theeraphat-Sorasetsakul: Updated.

packages/utils/src/cache/types.ts
Theeraphat-Sorasetsakul: Updated.

packages/utils/src/loader/svg-loader.ts
Theeraphat-Sorasetsakul: Updated. And thank for an idea. responseXML api returns a represent element. So it should one svg child from the 
result.

packages/utils/src/cache/cache-indexeddb.ts
Theeraphat-Sorasetsakul: Updated by remove this code.

packages/utils/src/cache/local-cache.ts
Theeraphat-Sorasetsakul: Updated by move the ready logic to each storage.

packages/utils/src/cache/local-cache.ts
Theeraphat-Sorasetsakul: Updated by move the memory cache logic to storage.

packages/utils/src/cache/local-cache.ts
Theeraphat-Sorasetsakul: Update by move the ready logic to storage. It will start calling restore at constructor.

packages/utils/src/cache/cache-localstorage.ts
Theeraphat-Sorasetsakul: Updated. I create a static prefix at constants.ts to share and use with name that user input at constructor.

packages/utils/src/cache/cache-indexeddb.ts
Theeraphat-Sorasetsakul: Updated by constructor receive only name. Use the name with share prefix for database name and store name. Also set database version to 1 to avoid clash versioning.

packages/utils/src/cache/local-cache.ts
Theeraphat-Sorasetsakul: Update by move restore logic to storage and call it at constructor.

packages/utils/src/cache.ts
```diff
@@ -0,0 +1,7 @@
+export { CacheLocalStorage } from './cache/cache-localstorage.js';
+export { CacheIndexedDBStorage } from './cache/cache-indexeddb.js';
```
TremayneChrist: Don't export these. These are now internal.

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Stores data in indexedDB for use across multiple sessions.
+ */
+export class CacheIndexedDBStorage implements CacheStorage {
+  /**
+   * Prefix for database name, and store name
+   * to avoid database to clash with the workspace
+   */
+  protected prefixKey = '';
```
TremayneChrist: Name of this is a bit wrong. Maybe use `dbName`?

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+title: Local caching Utility
+location: ./utils/cache
+type: page
+layout: default
+-->
+
+::status-complete::
+
+# Local cache
+The utility provides provides an alternative way to store data on your local machine.
```
wsuwt:
```suggestion
The utility provides an alternative way to store data locally in browser.
```

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+There are 2 main classes that collaborate: wrapper and storage. LocalCache is high level class that wraps the selected storage to be asynchronous api.
+The second is storage class. Store data on the user's machine. The storages has a list of support as below:
+- localstorage
```
wsuwt:
```suggestion

- localstorage
```

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+  { storage: 'indexeddb' } // config with storage type
+);
+// Example using an api
+await cache.set(key, data);
+const item = await cache.get(key);
+```
+
+## LocalCache APIs
+
+### set
```
wsuwt, line 35:
```suggestion
### set

```

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+type: page
+layout: default
+-->
+
+::status-complete::
+
+# Local cache
+The utility provides provides an alternative way to store data on your local machine.
+There are 2 main classes that collaborate: wrapper and storage. LocalCache is high level class that wraps the selected storage to be asynchronous api.
+The second is storage class. Store data on the user's machine. The storages has a list of support as below:
```
wsuwt: Does developer needs to be aware that there is storage class ? I don't see any code related to the class here. I think we only need to mention what is LocalCache class, how to instantiate it and its storage options.

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+| key | String | Cache key |
+
+### clear
+Remove related items in the storage
+
+```typescript
+import { LocalCache } from '@refinitiv-ui/utils/cache.js';
+// Create cache and use the storage
+const cache = new LocalCache('my-cache', { storage: 'indexeddb' });
+// remove specific item follow by the key 'my-item-01'
```
wsuwt:
```suggestion
```

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+#### Arguments
+
+| Name | Type | Description |
+| --- | --- | --- |
+| key | String | Cache key |
+
+### clear
+Remove related items in the storage
```
wsuwt: What are related items ?

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+-->
+
+::status-complete::
+
+# Local cache
+The utility provides provides an alternative way to store data on your local machine.
+There are 2 main classes that collaborate: wrapper and storage. LocalCache is high level class that wraps the selected storage to be asynchronous api.
+The second is storage class. Store data on the user's machine. The storages has a list of support as below:
+- localstorage
+- indexeddb
```
wsuwt: We should mention suitable use cases for each storage option.

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+
+LocalCache can select only one storage to use. Here is a common use.
+
+```typescript
+import { LocalCache } from '@refinitiv-ui/utils/cache.js';
+// Create cache and use the storage
+const cache = new LocalCache(
+  // prefix key for localstorage or database name for indexeddb
+  'my-cache',
```
wsuwt:
```suggestion
  'cache-prefix',
```

or something better.

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+### set
+Caches a value against a key to use until expired
+
+```typescript
+import { LocalCache } from '@refinitiv-ui/utils/cache.js';
+const key = 'my-item-01';
+const value = { label: '01', value: 1};
+// Create cache and use the storage
+const cache = new LocalCache('my-cache', { storage: 'indexeddb' });
+// store 'my-item-01': { label: '01', value: 1} to LocalCache. Default cache expiry is 5 days
```
wsuwt:
```suggestion
// store 'my-item-01': { label: '01', value: 1} to `indexeddb` ?
```

documents/src/pages/utils/cache.md
```diff
@@ -0,0 +1,133 @@
...
+cache.set(key, value, expires);
+```
+
+#### Arguments
+
+| Name | Type | Description |
+| --- | --- | --- |
+| key | String | Cache key |
+| value | unknown | Data to store in cache |
+| expires | Number | Cache expiry in seconds. Defaults to 5 days. |
```
wsuwt: I think there are less verbose ways to document cache APIs. I personally find something like TS declaration much easier to understand. 

```ts
export declare class LocalCache {
    /**
     * Storage to store data
     */
    protected storage: CacheStorage;
    constructor(name: string, config: LocalCacheConfig);
    /**
     * Caches a value against a key to use until expired
     * @param key Cache key
     * @param value Data to store in cache
     * @param [expires=432000] Cache expiry in seconds. Defaults to 5 days.
     * @returns Promise void
     */
    set(key: string, value: string, expires?: number): Promise<void>;
    /**
     * Returns cache data value based on provided key
     * @param key Cache key
     * @returns Promise string data or `null` if nothing is cached
     */
    get(key: string): Promise<unknown | null>;
    /**
     * Remove cache data value based on provided key
     * @param key Cache key
     * @returns Promise void
     */
    remove(key: string): Promise<void>;
    /**
     * Clear all memory cache
     * @returns Promise void
     */
    clear(): Promise<void>;
}
```

wsuwt (changes requested):

packages/utils/src/cache.ts
```diff
@@ -0,0 +1,7 @@
...
+export { CacheMap } from './cache/types.js';
+export { LocalCache, LocalCacheConfig } from './cache/local-cache.js';
+
+export { CacheStorage } from './cache/interfaces/CacheStorage.js';
+export { CacheItem } from './cache/interfaces/CacheItem.js';
```
wsuwt, line 5: I think files structure can be improve. Why isn't `CacheMap` place in the same folder as `CacheStorage` and `CacheItem` ?

wsuwt: Can storage be group ? Can we remove `cache-` prefix ?

- storages
       - idb
       - localstorage

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+/**
+ * Stores data in indexedDB for use across multiple sessions.
+ */
+export class CacheIndexedDBStorage implements CacheStorage {
+  /**
+   * Prefix for database name, and store name
+   * to avoid database to clash with the workspace
```
wsuwt:
```suggestion
   * to avoid database to clash with other storages. 
```

packages/utils/src/cache/constants.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * static prefix our cache version
+ */
+const PREFIX = 'ef-cache-';
```
wsuwt:
```suggestion
**
 * Default cache storage name prefix
 */
export const PREFIX = 'ef-cache-';
```

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Clear all item in this storage
+   * @returns {void}
+   */
+  public async clear (): Promise<void> {
+    await this.ready;
+    this.cache?.clear();
+    await this.db?.clear(this.prefixKey as never);
```
wsuwt: Interesting choice of casting key to `never` here. Is there better ways to do this, from looking at `idb` dts they do offer some types.

packages/utils/src/cache/interfaces/CacheStorage.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  /**
+   * Returns all values in this storage
+   */
+  restoreItems(): Promise<unknown>;
+
+  /**
+   * Set a value against a key
+   */
+  setItem(key: string, value: unknown): Promise<void>;
```
wsuwt:
```suggestion
  setItem(key: string, value: unknown): Promise<void>;
  
```

wsuwt:
> Note: Indexeddb in Firefox private mode isn't work. https://bugzilla.mozilla.org/show_bug.cgi?id=781982

Would it be possible to at least show warning ?

documents/src/pages/utils/cache.md
TremayneChrist: Should actually document this as just the `name` of the cache. A user shouldn't have to worry about any prefixes.

documents/src/pages/utils/cache.md
TremayneChrist:
```suggestion
Clears all items from cache
```

packages/utils/src/cache/cache-indexeddb.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Returns Error message when unable to connect indexedDB
+   * @param message {String}
+   * @returns Error message
+   */
+  private errorMessage (message: string): Error {
+    return new Error(`Unable to connect to indexedDB.\nDatabase name:'${this.prefixKey}'.\nDatabase Version: ${this.version}\nStore name: ${this.prefixKey}\n ${message}`);
+  }
```
TremayneChrist: Only place class related methods in class. As this is just a helper for throwing an error message, you can move it outside of the class.

packages/utils/src/cache/interfaces/CacheStorage.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Structure for any local storage
+ */
+export interface CacheStorage {
+
+  /**
+   * Returns all values in this storage
+   */
+  restoreItems(): Promise<unknown>;
```
TremayneChrist: Aligning with `clear`, let's use `restore` instead. Also the `Promise<unknown>` doesn't look right to me. I feel this method would just restore the items to the storage class, so can return `Promise<void>`?

packages/utils/src/cache/local-cache.ts
```diff
@@ -0,0 +1,74 @@
...
+  protected storage!: CacheStorage;
+
+  constructor (name: string, config: LocalCacheConfig) {
+    if (config.storage.toLowerCase() === 'indexeddb') {
+      this.storage = new CacheIndexedDBStorage(name);
+    }
+    if (config.storage.toLowerCase() === 'localstorage') {
+      this.storage = new CacheLocalStorage(name);
+    }
+  }
```
TremayneChrist: As this is a public API, perform checks on params. Also options should be optional.
```suggestion
  constructor (name: string, config?: LocalCacheConfig) {
    if (typeof name !== 'string') {
      throw new TypeError('Expected name to be of type string');
    }
    if (name.length === 0) {
      throw new RangeError('Expected name to have a length');
    }
    const options = Object.assign({}, config);
    switch (options.storage) {
      case 'indexeddb':
        this.storage = new CacheIndexedDBStorage(name);
        break;
      case 'localstorage':
      default:
        this.storage = new CacheLocalStorage(name);
    }
    throw new RangeError('Unknown storage type');
```

packages/utils/src/cache/constants.ts
TremayneChrist: We can make the key easier to read by splitting this values differently. Prefix could be as simple as `ef`.

```
[prefix][storeName][cacheKey] -> [ef][svg-loader][star.svg]
```

packages/utils/src/cache/cache-localstorage.ts
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Stores data in `localStorage` for use across multiple sessions.
+ */
+export class CacheLocalStorage implements CacheStorage {
+
+  /**
+   * Prefix for all keys
+   * to avoid key to clash with the workspace
+   */
+  protected prefixKey = '';
```
TremayneChrist: Again, I'm not too sure on this property name. I appreciate this is used to prefix storage keys, however, in the terms of the class and the storage object, it is the name of the database.

It might be cleaner to align this with the IndexedDB storage object and use `dbName` or `storeName` and then have a helper function outside of the class to generate the prefix/key name.

```typescript
const getItemKey = (dbName, cacheKey) => `${dbName}[${cacheKey}]`;

export class CacheLocalStorage implements CacheStorage {
  ...
  constructor (name: string) {
    this.dbName = `[${PREFIX}][${name}]`;
  }
}
```

documents/src/pages/utils/cache.md
TremayneChrist: Storage classes should now be internal, so they don't need to be documented.

documents/src/pages/utils/cache.md
TremayneChrist: I think this is just aligning with Aleks' docs. We probably need to revisit how we document these APIs as a whole.

TremayneChrist (changes requested):
- User should be able to create a `LocalCache` easily, by just providing a name. Options should be optional and the storage should default to `localstorage`.
- Remove type casting, like `obj as never`.
- Make sure all Storage classes are internal only
- Simplify prefixing in storage classes and make sure they're automatic and a user doesn't need to know about them.
- Make keys more readable. `[prefix][storeName][cacheKey]`
- Add parameter checks to publis APIs
- Move helper methods out of classes

documents/src/pages/utils/cache.md
wsuwt: You are right. I got confused.

packages/utils/src/cache/interfaces/CacheStorage.ts
Theeraphat-Sorasetsakul: To change from `Promise<unknown>` to `Promise<void>`, Do you mean restoring into memory cache? It's not related to getReady function, isn't it?

packages/utils/src/cache/storages/localstorage.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Stores data in `localStorage` for use across multiple sessions.
+ */
+export class CacheLocalStorage implements CacheStorage {
+
+  /**
+   * Prefix for all keys
+   * to avoid key to clash with the workspace
+   */
+  protected prefixKey = '';
```
wsuwt: I think this should be align with indexdb storage class.

packages/utils/src/cache/cache-localstorage.ts
Theeraphat-Sorasetsakul: Updated.
I create a helper `getItemKey` for a key in localstorage. The helper will be replace logic of handle key. And change `prefixKey` to `dbName`.

packages/utils/src/cache/local-cache.ts
Theeraphat-Sorasetsakul: Updated.
I move `throw new TypeError('Unknown storage type');` to be default case from your suggest ` throw new RangeError('Unknown storage type');`. It will throws error when user input wrong storage. And also supports case `undefined` when user don't set the local cache option.

packages/utils/src/cache/cache-indexeddb.ts
Theeraphat-Sorasetsakul: Updated to move out of the class.

packages/utils/src/cache/cache-indexeddb.ts
Theeraphat-Sorasetsakul: Updated. The name is `dbName`.

packages/utils/src/cache.ts
Theeraphat-Sorasetsakul: Updated by remove those storage export.

packages/utils/src/cache/interfaces/CacheStorage.ts
Theeraphat-Sorasetsakul: Updated by following your suggestion.

packages/utils/src/cache/cache-indexeddb.ts
Theeraphat-Sorasetsakul: Updated to remove type casting.
I and Goh propose an idea to use literal type ``type StoreName = `[${StoragePrefix.DEFAULT}][${string}]` ``, and cast it into store type of idb (thrid party indexeddb).

packages/utils/src/cache/storages/localstorage.ts
Theeraphat-Sorasetsakul: Updated.
I create a helper `getItemKey` for a key in localstorage. The helper will be replace logic of handle key. And change `prefixKey` to `dbName`.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by following your suggestion.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by following your suggestion.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by following your suggestion.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by changing comment to remove all items in the cache.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by replacing Storage classes content to more detail on localcahe class.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by following your suggestion.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated. I put more info of each storages.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by changing comment to name of the cache.

documents/src/pages/utils/cache.md
Theeraphat-Sorasetsakul: Updated by following your suggestion.

packages/utils/src/cache.ts
Theeraphat-Sorasetsakul: Updated by moving file. Remove prefix `cache-` and create storage folder.

packages/utils/src/cache/cache-indexeddb.ts
Theeraphat-Sorasetsakul: Updated by following your suggestion.

packages/utils/src/cache/constants.ts
Theeraphat-Sorasetsakul: Updated by prefix value is `ef`. I and Goh propose an idea to use enum `StoragePrefix.DEFAULT`. It uses for literal type to resolve type casting issue.

packages/utils/src/cache.ts
wsuwt: Why is `CacheMap` in types and `CacheItem`, `CacheStorage` in interfaces ? They are related and should be group together.

documents/src/pages/utils/cache.md
wsuwt: I think storing in browser sounds better than machine.

documents/src/pages/utils/cache.md
wsuwt: Again here, storage classes don't need to be mentioned in the docs.

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/**
+ * Literal type for dynamic store name casting to idb
+ */
+type StoreName = `[${StoragePrefix.DEFAULT}][${string}]`;
+
+/**
+ * Returns Error message when unable to connect indexedDB
+ * Database and Store are same name. Version is 1. So don't try to log a real database config.
```
wsuwt: Not sure what this means.

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -0,0 +1,177 @@
...
+  [key: StoreName]: {
+    key: string;
+    value: CacheItem;
+  }
+}
+
+/**
+ * Literal type for dynamic store name casting to idb
+ */
+type StoreName = `[${StoragePrefix.DEFAULT}][${string}]`;
```
wsuwt: Why `StoreName` ? I think it would sound nicer if it aligns with `dbName`

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -0,0 +1,177 @@
...
+    this.cache?.clear();
+    await this.db?.clear(this.dbName);
+  }
+
+  /**
+   * Restore all values into memory cache
+   * @returns Promise void
+   */
+  public async restore (): Promise<void> {
+    const cacheItems = new Map() as CacheMap;
```
wsuwt:
```suggestion
    const cache: CacheMap = new Map();
```

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -0,0 +1,177 @@
...
+   */
+  public async clear (): Promise<void> {
+    await this.ready;
+    this.cache?.clear();
+    await this.db?.clear(this.dbName);
+  }
+
+  /**
+   * Restore all values into memory cache
+   * @returns Promise void
```
wsuwt:
```suggestion
   * @returns {void}
```

packages/utils/src/cache/storages/localstorage.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    return Promise.resolve();
+  }
+
+  /**
+   * Restore all values into memory cache
+   * @returns Promise void
+   */
+  public async restore (): Promise<void> {
+    const items = new Map<string, CacheItem>();
```
wsuwt:
```suggestion
   const cache: CacheMap = new Map();
```

packages/utils/src/cache/storages/localstorage.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+   * Retrieve cache item from localStorage
+   * @param key key to retrieve value
+   * @returns data from the key
+   */
+  private retrieve (key: string): CacheItem | null {
+    try {
+      return JSON.parse(localStorage.getItem(key) || '') as CacheItem;
```
wsuwt, line 133: There's no need to have try catch block. You can return null right away if `localStorage.getItem(key)` is null.

packages/utils/src/loader/svg-loader.ts
```diff
@@ -1,5 +1,9 @@
+import { LocalCache } from '../cache.js';
 import { CDNLoader } from './cdn-loader.js';
 
+// Create cache and select the storage
```
wsuwt:
```suggestion
```

packages/utils/src/cache/local-cache.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { CacheStorage } from './interfaces/CacheStorage';
+import { IndexedDBStorage } from './storages/indexeddb.js';
+import { LocalStorage } from './storages/localstorage.js';
+
+export type LocalCacheConfig = {
+  storage: 'localstorage' | 'indexeddb';
+};
+
+/**
+ * Stores data in a local cache that can be specified to be stored in localstorage or indexedDB.
```
wsuwt:
```suggestion
 * Cache utility that simplifies, unifies and enhances native browser storage. 
```

packages/utils/src/cache/storages/localstorage.ts
```diff
@@ -0,0 +1,125 @@
...
+import { getItemKey } from '../helpers.js';
+
+/**
+ * Stores data in `localStorage` for use across multiple sessions.
+ */
+export class LocalStorage implements CacheStorage {
+
+  /**
+   * Database for this class. Prefix for key of all items.
+   * to avoid key to clash with other.
```
wsuwt:
```suggestion
```

packages/utils/src/cache/storages/localstorage.ts
```diff
@@ -0,0 +1,125 @@
...
+import { StoragePrefix } from '../constants.js';
+import { getItemKey } from '../helpers.js';
+
+/**
+ * Stores data in `localStorage` for use across multiple sessions.
+ */
+export class LocalStorage implements CacheStorage {
+
+  /**
+   * Database for this class. Prefix for key of all items.
```
wsuwt:
```suggestion
   * Database name.
```

packages/utils/src/cache/storages/indexeddb.ts
Theeraphat-Sorasetsakul: I want to say what database, store, and version that the util try to access. Because Indexeddb allows only access latest version. To know current version, `indexeddb..databases()` is native function to know all database version isn't support in firefox.

packages/utils/src/cache/storages/indexeddb.ts
Theeraphat-Sorasetsakul: Updated to following your suggestion. And do this to other comment.

packages/utils/src/cache/storages/indexeddb.ts
Theeraphat-Sorasetsakul: Updated to following your suggestion.

packages/utils/src/cache/storages/indexeddb.ts
Theeraphat-Sorasetsakul: Updated by following your suggestion.

packages/utils/src/cache/storages/indexeddb.ts
wsuwt: Please revise. I do not get what you are trying to write.

packages/utils/src/cache/storages/localstorage.ts
Theeraphat-Sorasetsakul: I'm not sure to remove try catch block. Because most of the way to use JSON.parse() requires try catch. And we don't know what a value of localStorage.getItem(key).

packages/utils/src/cache/storages/indexeddb.ts
Theeraphat-Sorasetsakul: I want to log database, store, and version that the util try to access that will be database name equal to store name, version is fit to 1, and it has only 1 store in the database. So I think the errorMessage should log what attempt connection. It don't log current version.

To log current version, I haven't found the way that support all browsers. `window.indexeddb.databases()` is one way to get current version of all databases, but it's not support on Firefox.

wsuwt: I would rephrase to something shorter and rewrite the comment. Does something like `window.indexeddb.databases() ? window.indexeddb.databases() : 1` work ?

packages/utils/src/cache.ts
Theeraphat-Sorasetsakul: Updated by moving CacheMap to interfaces folder.

packages/utils/src/cache/storages/indexeddb.ts
Theeraphat-Sorasetsakul: Updated. As we talked, It already said in open. So the comment line is removed.

packages/utils/src/cache/constants.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * Default cache storage name prefix
+ */
+export enum StoragePrefix {
```
wsuwt:
```suggestion
export enum DatabasePrefix {
```

Theeraphat-Sorasetsakul:
I already updated all feedbacks. Could you please revisit again @wsuwt @TremayneChrist .

wsuwt (dismissed):

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -0,0 +1,169 @@
+import { openDB } from 'idb';
+import type { DBSchema, IDBPDatabase } from 'idb';
+import type { CacheMap } from '../interfaces/CacheMap';
+import type { CacheItem } from '../interfaces/CacheItem';
+import type { CacheStorage } from '../interfaces/CacheStorage';
+import { DatabasePrefix } from '../constants.js';
+
+type DbName = `[${DatabasePrefix.DEFAULT}][${string}]`;
```
TremayneChrist:
```suggestion
type DBName = `[${DatabasePrefix.DEFAULT}][${string}]`;
```

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * IDB's database instance
+   */
+  private db: IDBPDatabase<IndexedDBDatabase> | undefined;
+
+  /**
+   * Database version
+   */
+  private version = 1;
```
TremayneChrist, line 45:
```suggestion
  private readonly version = 1;
```

packages/utils/src/cache/storages/localstorage.ts
```diff
@@ -0,0 +1,139 @@
...
+  protected dbName = '';
+
+  /**
+   * Internal cache object
+   */
+  protected cache: CacheMap | null | undefined;
+
+  /**
+   * Constructor
+   * @param name prefix key for all item
```
TremayneChrist:
```suggestion
   * @param name name of the data store
```

Theeraphat-Sorasetsakul: Updated by following your suggest.

packages/utils/src/cache/storages/indexeddb.ts
Theeraphat-Sorasetsakul: Updated by following your suggest.

documents/src/navigation.md
```diff
@@ -89,7 +89,9 @@ type: nav
     - [Overview](./accessibility/overview)
   - [Internationalization (i18n)](./intl/internationalization)
   - [Theme Switching](./guides/theme-switching)
-
+- Building Components
+  - Utilities
+  * [Cache](./utils/cache)
```
TremayneChrist: Remove from nav for now

TremayneChrist: We can add it back when we continue with `efx-element` documentation

Theeraphat-Sorasetsakul: Updated by following your suggest. I remove the all changes in this file.

TremayneChrist (approved):

PR #413, opened 2022-07-25
Refactor datetime-picker

goremikins:


packages/elements/src/datetime-picker/index.ts
```diff
@@ -880,44 +694,12 @@ export class DatetimePicker extends ControlElement implements MultiValue {
...
-    }
-
-    if (path.includes(this.iconEl)) {
-      this.toggleOpened();
-    }
-    else if (!this.inputTriggerDisabled) {
-      this.setOpened(true);
-    }
+  private onButtonTap (event: TapEvent): void {
+    this.setOpened(true);
```
Sarin-Udompanish, line 667: The popup will not close when click on the button again.

packages/elements/src/datetime-picker/index.ts
```diff
@@ -880,44 +694,12 @@ export class DatetimePicker extends ControlElement implements MultiValue {
...
-      return; /* popup is managed separately */
-    }
-
-    if (path.includes(this.iconEl)) {
-      this.toggleOpened();
-    }
-    else if (!this.inputTriggerDisabled) {
-      this.setOpened(true);
-    }
+  private onButtonTap (event: TapEvent): void {
```
Sarin-Udompanish: Unused variable ( event )

packages/elements/src/datetime-picker/index.ts
```diff
@@ -546,6 +547,31 @@ export class DatetimePicker extends ControlElement implements MultiValue {
...
+  private filterInvalidViews (views: string[]): string[] {
+    const filtered = [];
+
+    // views must match in duplex mode
+    if (views.length !== (this.isDuplex() ? 2 : 1)) {
+      return [];
+    }
+
+    for (let i = 0; i < views.length && filtered.length <= 2; i += 1) {
+      const view = views[0];
```
Sarin-Udompanish: Should it be `const view = views[i]` ?

goremikins: Done

PR #415, opened 2022-07-27
docs(interactive-chart): fix incorrect sample in seasonality section

dtanp-rft:
Incorrect sample code in Seasonality section.

those options should be a configuration in `series` object.
https://tradingview.github.io/lightweight-charts/docs/api/interfaces/SeriesOptionsCommon

### DIFF
--- a/documents/src/pages/elements/interactive-chart.md
+++ b/documents/src/pages/elements/interactive-chart.md
@@ -467,7 +467,6 @@ const months = {
 }
 const chart = document.getElementById('seasonality');
 chart.config = {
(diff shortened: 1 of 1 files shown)


Nantawat-Poothong:
Interesting, I already reviewed PR and it works perfectly.
I doubt about the sample code, is it wrong in the first place or tradingview just updated their schema.

dtanp-rft:
Tbh, I have exactly the same question! :)

Nantawat-Poothong (approved):

PR #430, opened 2022-08-16
fix(datetime-picker): change focus outline styles of calendar button

Sarin-Udompanish:
## Description
Change focus outline styles of calendar button 

From

Screen Shot 2565-08-16 at 16 01 34

To

Screen Shot 2565-08-16 at 15 19 40

I use 2px outline width in Halo theme because with 1px it looks like a border more than an outline but I'm fine with 1px width as well.


goremikins:
I do not think I like it :)

Sarin-Udompanish:
Updated @goremikins 

- Add a left border to calendar button.
- Change focus style of calendar button ( can't use :focus-visible selector because it too new ).
- Change border color on readonly and disabled state.

Screen Shot 2565-08-17 at 17 47 39

Calendar button focused

Screen Shot 2565-08-17 at 17 48 15

PR #432, opened 2022-08-18
fix(list-item): incorrect style on halo theme

Nantawat-Poothong:
## Description
The `ef-list-item` element have import an incorrect style in halo-theme, it makes `ef-combo-box` and `ef-list` has an incorrect style too.

Solution
- change `ef-list-item` halo-theme to import theme from `ef-item` halo-theme instead of elemental-theme.
- and solar-theme as well.

## Type of change

### DIFF
--- a/packages/halo-theme/src/custom-elements/ef-list-item.less
+++ b/packages/halo-theme/src/custom-elements/ef-list-item.less
@@ -1,1 +1,1 @@
-@import "ef-item";
+@import 'ef-item';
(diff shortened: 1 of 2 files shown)


packages/halo-theme/src/custom-elements/ef-list-item.less
```diff
@@ -1 +1 @@
-@import '@refinitiv-ui/elemental-theme/src/custom-elements/ef-list-item';
+@import "ef-item";
```
wsuwt:
```suggestion
@import 'ef-item';
```

packages/solar-theme/src/custom-elements/ef-list-item.less
```diff
@@ -1 +1 @@
-@import '@refinitiv-ui/elemental-theme/src/custom-elements/ef-list-item';
+@import "ef-item";
```
wsuwt:
```suggestion
@import 'ef-item';
```

wsuwt (approved):

PR #433, opened 2022-08-22
v7 using fetch

Theeraphat-Sorasetsakul:
## Description
Replace XHR requests with Fetch. Because v7 doesn't support IE.

Fixes # ([jira link ELF-1888](https://jira.refinitiv.com/browse/ELF-1888))

## Type of change

## Checklist


packages/utils/src/loader/cdn-loader.ts
```diff
@@ -49,39 +49,41 @@ export class CDNLoader {
...
     catch (e) {
       // Failed response...
       this.responseCache.delete(href);
+      if (e instanceof Response) {
+        return Promise.resolve({
+          status: e.status || 0,
+          statusText: e.statusText || 'Failed to make request'
+        } as Response);
+      }
       return Promise.resolve({
```
Sakchai-Refinitiv, line 66: It's failed, why resolve?

packages/utils/src/loader/cdn-loader.ts
```diff
@@ -49,39 +49,41 @@ export class CDNLoader {
...
+  private async loadContent (href: string): Promise<Response> {
     try {
-      const response = await this.xmlRequest(href);
+      const response = await this.fetchRequest(href);
       return response;
     }
     catch (e) {
       // Failed response...
       this.responseCache.delete(href);
+      if (e instanceof Response) {
```
Sakchai-Refinitiv: Two returns they are the same thing should group it

packages/utils/src/loader/cdn-loader.ts
```diff
@@ -49,39 +49,41 @@ export class CDNLoader {
...
-        if (xhr.readyState === 4 && xhr.status >= 200 && xhr.status < 400) {
-          resolve(xhr);
-        }
-        else {
-          reject(xhr);
-        }
-      };
-      xhr.onerror = (): void => reject(xhr);
-      xhr.send();
+  private async fetchRequest (href: string): Promise<Response> {
```
Sakchai-Refinitiv: Rename it to 'fetch' might be better.

wsuwt: I think we might not even need this as a separate function. This can be easily merge into `loadContent` with a single catch block.

packages/utils/src/loader/cdn-loader.ts
Theeraphat-Sorasetsakul: Updated. Due to we called, I will resolve this conversation.

packages/utils/src/loader/cdn-loader.ts
Theeraphat-Sorasetsakul: Agreed and updated. I merges fetchRequest function into `loadContent`.

packages/utils/src/loader/cdn-loader.ts
Theeraphat-Sorasetsakul: loadContent is response handler. So it need to return as resolve error response to let other function takes over.

packages/utils/src/loader/svg-loader.ts
```diff
@@ -48,19 +48,21 @@ const stripUnsafeNodes = (...elements: Node[]): void => {
  * @param response Request response to test
  * @returns Is valid SVG
  */
-const isValidResponse = (response: XMLHttpRequest | undefined): response is XMLHttpRequest => {
-  return !!response && response.status === 200
-  && response.getResponseHeader('content-type') === 'image/svg+xml';
+const isValidResponse = (response: Response | undefined): response is Response => {
+  const isSVG = !!response?.headers.get('content-type')?.startsWith('image/svg+xml');
```
Sakchai-Refinitiv:
```suggestion
  const isSVG = Boolean(response?.headers.get('content-type')?.startsWith('image/svg+xml'));
```

Sakchai-Refinitiv: Please check `!!` in other changes as well

Theeraphat-Sorasetsakul: Updated by following your suggest.

packages/utils/src/loader/cdn-loader.ts
```diff
@@ -49,48 +49,33 @@ export class CDNLoader {
...
+      }
+      else if (e instanceof Response) {
+        errorMessage = e.statusText;
+      }
       return Promise.resolve({
-        status: 0, response: 'Failed to make request', responseText: 'Failed to make request'
-      } as XMLHttpRequest);
+        status: 0,
+        statusText: errorMessage
+      } as Response);
```
Nantawat-Poothong, line 69: Why you resolve this error when load content is fail and where this error will be used?

packages/utils/src/loader/cdn-loader.ts
```diff
@@ -49,48 +49,33 @@ export class CDNLoader {
...
-  private async loadContent (href: string): Promise<XMLHttpRequest> {
+  private async loadContent (href: string): Promise<Response> {
     try {
-      const response = await this.xmlRequest(href);
-      return response;
+      return await fetch(href);
     }
     catch (e) {
       // Failed response...
       this.responseCache.delete(href);
```
Nantawat-Poothong, line 58: Why need to delete response cache.

Theeraphat-Sorasetsakul: To delete from the cache, the cache should keep only success request. Fail requests like network problem, offline, or other issues shouldn't store because the request isn't successful. Fail response may take error or doesn't exist.

packages/elements/src/icon/__test__/helpers/helpers.js
```diff
@@ -1,24 +1,61 @@
...
-    return false;
+  }
+  return false;
 }
 
 export const generateUniqueName = name => `${name}_${iconId+=1}`;
 
 export const createMockSrc = icon => `https://mock.cdn.com/icons/${icon}.svg`;
+
+export const createFakeResponse = (body, config = responseConfigSuccess) => {
```
Nantawat-Poothong, line 27: I have no idea what this function is used for. How it related to test case?

packages/utils/src/loader/cdn-loader.ts
Theeraphat-Sorasetsakul: Agree. We should response to tell error  some where.

packages/elements/src/icon/__test__/helpers/helpers.js
Theeraphat-Sorasetsakul: In test, we don't make real request. The test must work in offline. That why I create this function to mock response to make it as fetch response.

packages/elements/src/flag/__test__/helpers/helpers.js
```diff
@@ -1,24 +1,61 @@
...
 export const createMockSrc = flag => `https://mock.cdn.com/flags/${flag}.svg`;
+
+export const createFakeResponse = (body, config = responseConfigSuccess) => {
+  const { ok, status, headers} = config;
+  const response = new window.Response(body, {
+    ok,
+    status,
+    headers,
+    clone: () => {
+      text: async () => {
```
Nantawat-Poothong, line 34: Should we wrap return object to this instead?
 ```javascript
clone: () => ({
  text: async () => {
    return await Promise.resolve(body);
  }
})
```
 Bracket from clone function make me confuse. that bracket is for function scope or just return an object 😅

Sakchai-Refinitiv (approved):

PR #439, opened 2022-08-25
POC - Distributed Cache for reducing network requests

Sakchai-Refinitiv:
## Description

Improve loading performance by reducing network requests. In the scenario, there are the same resources on multiple `IFrames`, the requests for that resources will reduce to only one and then request to the server, for other requests they will wait for data from the first request via messaging.

- Use `BroadcastChannel API` to send messages between `IFrames` to synchronize caching.
- Use `localStorage` to store a started request resources
- All states will be deleted when all resources are loaded and all messages are already sent.
- Handling the race condition when requesting and messaging.

❗️Important: This is the draft version of `POC` for a demo, The next step, must be implemented in the `LocalCache` class and merge into v7 after this PR https://github.com/Refinitiv/refinitiv-ui/pull/412

## Demo & Testing

I created this branch https://github.com/Theeraphat-Sorasetsakul/test-icon-idb/tree/local-cache-reducing-network-requests based on `LocalCache` demo  from @Theeraphat-Sorasetsakul  for testing  you can setup in the following steps

1. clone the  demo repository above
2. checkout to branch `local-cache-reducing-network-requests`
3. run command `npm i`
4. run command `npm link <path-to-your-refinitiv-ui-utils-package>` (this PR)
5. run command `npm run dev`
6. run command `npm run start:dev` (in separate terminal)

### Tracing Icon
You can trace each icon in all IFrames to see the actions of each IFrame how they interact with each icon by do the following steps
1. Open the demo pages
2. Open the developer tool
3. Go to tab `console` in developer tools
4. input filter logs for a specific icon for example, filter `3d` to find `3d-surface.svg` please see in the following image

image

Note: this console.log will be removed when this PR is ready to merge

## Type of change

## Checklist


Sakchai-Refinitiv:
@TremayneChrist @Theeraphat-Sorasetsakul, I'm about design a new `LocalCache` class  and move this implementation in there, you can early review it if you want 😀

dtanp-rft:
Do we really need to review this PR? I thought this is POC and surely code will change a lot with prod-ready code.

dtanp-rft:
My initial thought is that you may need to wrap your code into kind of class / module and design its public/protect/private API of your class. As of now, it's mixed in SVG loader. 

Then may need whiteboard review with Trem and team how this class/module will work with LocalCache and SVGLoader.

packages/utils/src/cache/constants.ts
```diff
@@ -4,3 +4,7 @@
 export enum DatabasePrefix {
   DEFAULT = 'ef'
 }
+
+export const CACHE_PREFIX = 'ef';
+export const MESSENGER_PREFIX = 'ef';
+export const MESSENGER_NO_MESSAGE_DELAY = 3000;
```
wsuwt: What does `MESSENGER_NO_MESSAGE_DELAY` mean ?

packages/utils/src/cache/core-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+   * Storage to store data
+   */
+  protected storage!: CacheStorage;
+
+  constructor (name: string, config?: CacheConfig) {
+
+    if (typeof name !== 'string') {
+      throw new TypeError('Expected name to be of type string');
+    }
+    if (name.length === 0) {
```
wsuwt: What are the chances of this happening ?

packages/utils/src/cache/core-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @param key Cache key
+   * @param value Data to store in cache
+   * @param [expires=432000] Cache expiry in seconds. Defaults to 5 days.
+   * @returns {void}
+   */
+  public async set (key: string, value: string | Promise<string>, expires = 432000): Promise<void> {
+
+    let cacheValue: string;
+    if (value instanceof Promise) {
+      cacheValue = await value.then(item => item || '');
```
wsuwt: Why do you need `.then` when it is already await.

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+logger.timeStart(window.name);
+
+/**
+ * Mutual exclusion, prevent getting requests state concurrently from multi tabs/windows
+ */
+const mutex = new Mutex();
+
+enum StorageType {
+  Requests='requests',
+  Unloaded='unloaded'
```
wsuwt:
```suggestion
  Unloaded = 'unloaded'
```

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+type StorageNames = {
+  [name in StorageType]: string
+};
+
+type Requests = {
+  [key: string]: 'true'
```
wsuwt: Why do we need this type

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+
+const logger = new Logger();
+logger.timeStart(window.name);
+
+/**
+ * Mutual exclusion, prevent getting requests state concurrently from multi tabs/windows
+ */
+const mutex = new Mutex();
+
+enum StorageType {
```
wsuwt: Why is this name storage type when it is not a storage type

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+  /**
+   * Names for manage all states temporary
+   */
+  protected storageNames: StorageNames;
+
+  /**
+   * Get resource request list, sync the new requests to cache
+   */
+  protected get requestsState (): Requests {
+    const requests = (JSON.parse(localStorage.getItem(this.storageNames.requests) as string) || {}) as Requests;
```
wsuwt: This bit looks untidy. Can the statement be broken down.

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+    this.listen();
+  }
+
+  /**
+   * Initialize listening events from messages and storage
+   * @returns {void}
+   */
+  private listen (): void {
+    // Listen storage event to clear everythings if remove requests state
+    addEventListener('storage', ({ key, newValue }) => {
```
wsuwt: What is `storage` event ?

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  /**
+   * Caches a value against a key to use until expired
+   * @param key Cache key
+   * @param value Data to store in cache
+   * @param [expires=432000] Cache expiry in seconds. Defaults to 5 days.
+   * @returns {void}
+   */
+  public async set (key: string, value: string | Promise<string | undefined>, expires = 432000): Promise<void> {
```
wsuwt:
```suggestion
  public override async set (key: string, value: string | Promise<string | undefined>, expires = 432000): Promise<void> {
```

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  /**
+   * Returns cache data value based on provided key
+   * @param key Cache key
+   * @returns Promise string data or `null` if nothing is cached
+   */
+  public async get (key: string): Promise<string | null> {
+    const iconName: string = key.split('/').pop() || '';
+    const item = await this.storage.get(key) as CacheItem;
```
wsuwt: Can we utilise generic type to set type for the storage ? 

e.g., `new CacheStorage<CacheItem>()`

and

```ts
interface CacheStorage<T> {
  get(key: string): Promise<T | null>;
}
```

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  /**
+   * Caches a value to active cache without writing to storage
+   * @param key Cache key
+   * @param value Data to store in cache
+   * @param [expires=432000] Cache expiry in seconds. Defaults to 5 days.
+   * @returns {void}
+   */
+  protected syncActive (key: string, value: string, expires = 432000): void {
```
wsuwt: What is active ?

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,300 @@
...
+
+
+  /**
+   * Clean up all temporary states
+   * @returns {void}
+   */
+  private clean (): void {
+    localStorage.removeItem(this.storageNames.requests);
+    this.requests = {};
+    this.waiting.clear();
```
wsuwt, line 329: I would keep either `clean` or `clear` not both.

packages/utils/src/cache/messenger.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Constructor
+   * @param name messenger name
+   */
+  constructor (name: string) {
+    const messengerName = `[${MESSENGER_PREFIX}][${name}]`;
+    this.broadcastChannel = new BroadcastChannel(messengerName);
+    this.storageNames = {
+      messagePosts: `${messengerName}[message-posts]`
```
wsuwt: Can we utilise TypeScript's template literal here ?

packages/utils/src/cache/messenger.ts
```diff
@@ -0,0 +1,122 @@
...
+   */
+  private increaseTotalPost (): number {
+    let messageCount = localStorage.getItem(this.storageNames.messagePosts) || 0;
+    messageCount = Number(messageCount) + 1;
+    localStorage.setItem(this.storageNames.messagePosts, messageCount.toString());
+    return messageCount;
+  }
+
+  /**
+   * Distribute cache
```
wsuwt: Or 
```suggestion
   * Distribute message
```
?

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -76,13 +76,36 @@ export class IndexedDBStorage implements CacheStorage {
...
+
   /**
    * Returns an item from cache database using provided key
    * @param key item key
    * @returns CacheItem or `null` if nothing is cached
    */
   public async get (key: string): Promise<CacheItem | null> {
     await this.ready;
+    // if (key.includes('3d')) {
+    //   console.log(window.name, '3d db is ready?', this.ready, this.cache);
```
wsuwt:
```suggestion
```

packages/utils/src/cache/storages/indexeddb.ts
```diff
@@ -159,6 +191,7 @@ export class IndexedDBStorage implements CacheStorage {
   private async getReady (): Promise<boolean> {
     try {
       await this.restore();
+      // console.log(window.name, '3d ready', Date.now().toString());
```
wsuwt:
```suggestion
```

wsuwt (dismissed):

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: I have already updated it.

packages/utils/src/cache/storages/indexeddb.ts
Sakchai-Refinitiv: Removed

packages/utils/src/cache/storages/indexeddb.ts
Sakchai-Refinitiv: Removed

packages/utils/src/cache/messenger.ts
Sakchai-Refinitiv: Fixed

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: Added

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: Refactored

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: Refactored and removed

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: Refactored and removed

packages/utils/src/cache/core-cache.ts
Sakchai-Refinitiv: Refactored and removed

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: It is used to store the request state of each cache item (icon) in localStorage to prevent the race condition problem. The value is always to be `true` because the logic in checking the requested state use key for checking only.

packages/utils/src/cache/storages/indexeddb.ts
Sakchai-Refinitiv: Removed

packages/utils/src/cache/storages/indexeddb.ts
Sakchai-Refinitiv: Removed

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,307 @@
...
+   */
+  protected storageNames: StorageNames;
+
+  /**
+   * Get resource request list, sync the new requests to cache
+   */
+  protected get requestsState (): Requests {
+    const state = localStorage.getItem(this.storageNames.requests);
+    if (!state) {
+      return {};
```
wsuwt: What does `{}` value return represent ?

packages/utils/src/cache/messenger.ts
```diff
@@ -0,0 +1,115 @@
...
+export type Message = {
+  id: number;
+  key: string;
+  value: string;
+}
+
+export type OnMessageCallback = (message: Message) => void;
+export type OnCompleteCallback = () => void;
+
+type StorageNames = { messagePosts: string };
```
wsuwt: Would this offer better type safe guard ? 
```suggestion
type StorageNames = { messagePosts: `${string}[message-posts]` };
```

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,307 @@
+import { Logger } from './helpers.js';
+import { Mutex } from 'async-mutex';
+import { CoreCache } from './core-cache.js';
+import { CacheMessenger } from './messenger.js';
+import type { CacheConfig } from './core-cache.js';
```
wsuwt:
```suggestion
import type { CacheConfig } from './core-cache';
```

packages/utils/src/cache/distributed-cache.ts
wsuwt: Maybe `Requested` or `Request` could be a better name.

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,307 @@
...
+const logger = new Logger();
+logger.timeStart(window.name);
+
+/**
+ * Mutual exclusion, prevent getting requests state concurrently from multi tabs/windows
+ */
+const mutex = new Mutex();
+
+type StorageNames = {
+  requests: string
```
wsuwt: If request is singular, I would use just
```suggestion
  request: string
```
or 
```suggestion
  requested: string
```

packages/utils/src/cache/distributed-cache.ts
wsuwt: Would it be better to use `null` to represent no request ?

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: The name should be `clean` because the action of this method is clean up the unused temporary data not clear all data, and the other reason is we cannot use `clear` because the method `clear` already exists in class which is used to clear all cache items.

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: I think the name `Requests` is still better because it stores multiple requests inside and should not be `Requested` because this type stores what is requested to the network which does not care about the states `pending` or `requested`

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: Yes, that is true, but I think it should always be the object because it makes coding easier/shorter that no need to check null every time before accessing the data inside.

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: I mention this issue in the above discussion, What do you think about it?

packages/utils/src/cache/constants.ts
Sakchai-Refinitiv: I rename it to `MESSENGER_LAST_MESSAGE_INTERVAL` it is the interval to check the last message

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: I rename it to `syncActiveCache` which using to add cache to the memory cache

packages/utils/src/cache/messenger.ts
Sakchai-Refinitiv: Yes, this is safer, I already updated it.

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: Updated, I change it to inline type import instead.

packages/utils/src/cache/messenger.ts
Sakchai-Refinitiv: Refactored

packages/utils/src/cache/constants.ts
wsuwt: I think this one has similar concept to debounce. What do you think ?

packages/utils/src/cache/distributed-cache.ts
wsuwt: Personally I would choose null checks over a few lines of code.

packages/utils/src/cache/helpers.ts
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable no-console */
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+
+import { MESSENGER_LAST_MESSAGE_INTERVAL } from './constants.js';
+
+type LoggerTime = {
+  startAt: number
+  endAt?: number
+}
+export class Logger {
```
wsuwt: This file is just a logger. I think it should be rename as `Logger.ts` or `helper/Logger.ts`.

Sakchai-Refinitiv: Please ignore it, it using for tracing only and will be removed before merge to main branch

packages/utils/src/cache/constants.ts
Sakchai-Refinitiv: Yes, it similar.

Sakchai-Refinitiv: This code has been removed.

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: This code has been removed

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: Updated

packages/utils/src/cache/distributed-cache.ts
Sakchai-Refinitiv: This code has been removed

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,360 @@
...
+    this.messenger.notify('coordinator', 'true');
+  }
+
+  /**
+   * Initialize listening events from messenger and storage
+   * @returns {void}
+   */
+  private listen (): void {
+    // Listen storage event for leader election
+    window.onstorage = ({ key, newValue }) => {
```
TremayneChrist: Don't listen in this way, because it will override the previous listener when a new instance of the cache is created.

Use `addEventListener` instead.

This listener will also have to be removed when the cache is no longer used.

packages/utils/src/cache/distributed-cache.ts
```diff
@@ -0,0 +1,360 @@
...
+    }
+  }
+
+  /**
+   * Clean up by resetting all requesting states,
+   * if found unload event that makes the current states is not correct
+   * in case users interupt a browser by refresh page
+   * @returns {void}
+   */
+  private handleUnload (): void {
```
TremayneChrist, line 181: We're not handling onload here. Shouldn't this be part of `listen()`

packages/utils/src/loader/svg-loader.ts
```diff
@@ -98,6 +93,19 @@ export class SVGLoader extends CDNLoader {
...
+   * Tries to load an SVG icon from CDN
+   * @param src name or Source location.
+   * @returns Promise which will be resolved with SVG
+   */
+  public async loader (src: string): Promise<string | undefined> {
+    return new Promise((resolve) => {
+      void this.load(src)
+        .then(response => extractSafeSVG(response))
+        .then((svg) => resolve(svg?.outerHTML));
+    });
```
TremayneChrist:
```suggestion
    const response = await this.load(src);
    const svg = extractSafeSVG(response);
    return svg?.outerHTML;
```

TremayneChrist (changes requested):
Overall, I find the distributed cache class very intertwined with other logic. We must revisit this and move all of the separate pieces into their own modules. This could be helped by refactoring to use a single host to manage all cache requests.

For now, though, let's look to fix these review comments and merge to the ~~v7~~ `distributed-chace` branch. In the new year we can iterate on top of this and look to simplify further.

packages/utils/src/cache/distributed-cache.ts
Nantawat-Poothong: My bad, I just changed Goh code from `addEventListener` to `onstorage` without aware of this.

TremayneChrist:
I've changed the base branch so that we can keep it isolated from v7 but continue to iterate on this solution in a clean PR next year.

packages/utils/src/cache/distributed-cache.ts
TremayneChrist: Let's leave this for now.

packages/utils/src/cache/distributed-cache.ts
TremayneChrist: As there is no hook for removing the listerner yet, we can skip this for now.

PR #440, opened 2022-08-25
ci: enable BrowserStack for nightly test

Sakchai-Refinitiv:
## Description

Enable `BrowserStack` for nightly test

## Type of change

Automate Testing

## Checklist


Sakchai-Refinitiv:
@dtanp-rft  We have not run the nightly test for a while, it might have test cases failed which means we should fix it.

Sakchai-Refinitiv:
> @dtanp-rft We have not run the nightly test for a while, it might have test cases failed which means we should fix it.

OK, But we should not leave it for long because the test cases might increase in Safari testing.

dtanp-rft:
We should exclude Safari test from BrowserStack as it's still failed. Then we can merge nightly and at least, it runs on FireFox & Chrome latest + latest -1.

Sakchai-Refinitiv:
> We should exclude Safari test from BrowserStack as it's still failed. Then we can merge nightly and at least, it runs on FireFox & Chrome latest + latest -1.

I have already updated it, the testing will run on three browsers Chrome, Firefox, and Edge in the version latest + latest -1.

dtanp-rft (approved):

PR #448, opened 2022-09-05
feat(color-picker): add accessibility

Nantawat-Poothong:
## Description
- add `role=button`
- add `aria-haspopup=dialog`
- use `readableColor` util
- use `aria-live` and `aria-label`
- add translation

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-1947

## Type of change

## Checklist


Theeraphat-Sorasetsakul (dismissed):

Theeraphat-Sorasetsakul (approved):

packages/elements/src/color-picker/__demo__/index.html
```diff
@@ -39,21 +39,23 @@
...
         const defaultColorOutput = document.getElementById('defaultColorOutput');
+        const readbleColor = document.getElementById('readbleColor');
         hexInputDefault.addEventListener('change', event => {
           defaultColorPicker.setAttribute('value', hexInputDefault.value);
         });
         defaultColorPicker.addEventListener('value-changed', event => {
           const value = event.detail.value;
           hexInputDefault.value = value;
           defaultColorOutput.value += 'Value Change : ' + value + '\n';
+          setTimeout(() => readbleColor.textContent = value ? defaultColorPicker._colorAriaLabel : '', 0);
```
wsuwt, line 58: Can we get label from attribute instead of accessing private property ? Why does it have to be in `settimeout` ?

Nantawat-Poothong: **Can we get label from attribute instead of accessing private property ?**
- Sound good, I will try this one.

**Why does it have to be in settimeout ?**
- Correct me if I am wrong. aria-label works with `TranslatePromise` as an async task and will not translate immediately. That means it should wait with settimeout.

PR #449, opened 2022-09-05
fix(interactive-chart): fix left position on legend

Theeraphat-Sorasetsakul:
## Description
The legend is overlap the left y-axis when the width of y-axis is too large.
Here an error link: https://codepen.io/sirirats/pen/yLKYRQd?editors=1010

Fixes # ([ELF-1931](https://jira.refinitiv.com/browse/ELF-1931))

Solution:
Add callback to calculate Legend's left position when the chart has left y-axis. The callback is subscribed to sizeChange event that attached on x-axis. The sizeChange event called when width or height is changed and called after rendered. Actually, legend relate to Y-axis but lightweight-chart doesn't have event for y-axis, 

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/interactive-chart/index.ts
+++ b/packages/elements/src/interactive-chart/index.ts
@@ -162,6 +162,8 @@ export class InteractiveChart extends ResponsiveElement {
 
   private hasDataPoint = false;
 
+  private DEFAULT_LEGEND_LEFT_POSITION = 15;
+
   /**
    * @returns return config of property component
    */
@@ -247,11 +249,11 @@ export class InteractiveChart extends ResponsiveElement {
 
   /**
   * Legend value observer
-  * @param value Legend value
+  * @param disabledLegend Legend value
   * @returns {void}
   */
-  private onLegendChange (value: boolean): void {
-    if (!value) {
+  private onLegendChange (disabledLegend: boolean): void {
+    if (!disabledLegend) {
       this.createLegend();
     }
     else {
@@ -342,9 +344,7 @@ export class InteractiveChart extends ResponsiveElement {
       this.mergeConfig(config);
       this.applyChartOptionSize(width, height);
 
-      if (!this.disabledLegend) {
-        this.createLegend();
-      }
+      this.onLegendChange(this.disabledLegend);
 
       if (this.legendStyle === 'horizontal') {
         this.legendContainer.classList.add(this.legendStyle);
@@ -394,6 +394,7 @@ export class InteractiveChart extends ResponsiveElement {
       this.legendContainer.textContent = '';
       this.chart.unsubscribeCrosshairMove(this.handleCrosshairMoved);
       this.legendInitialized = false;
+      this.chart.timeScale().unsubscribeSizeChange(this.onTimeScaleSizeChange);
     }
   }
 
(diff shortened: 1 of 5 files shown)


packages/elements/src/interactive-chart/index.ts
```diff
@@ -745,6 +752,8 @@ export class InteractiveChart extends ResponsiveElement {
         this.rowLegend = this.shadowRoot.querySelectorAll('.row');
       }
       this.chart.subscribeCrosshairMove(this.handleCrosshairMoved);
+      // Legend relates to value of each series that relate to priceScale. But the axis is no event for now. So use x-axis instead.
```
Nantawat-Poothong: How this comment explain the next line code?

packages/elements/src/interactive-chart/index.ts
```diff
@@ -162,6 +163,7 @@ export class InteractiveChart extends ResponsiveElement {
 
   private hasDataPoint = false;
 
+
```
Nantawat-Poothong:
```suggestion
```

packages/elements/src/interactive-chart/index.ts
```diff
@@ -734,6 +718,29 @@ export class InteractiveChart extends ResponsiveElement {
...
+   * @returns {void}
+   */
+  private handleLegendLeftPosition (): void {
+    const position = this.getPriceScalePosition();
+    if (position === 'left' || position === 'two-price') {
+      const leftPriceScaleWidth = this.chart?.priceScale('left').width() || 0;
+      this.legendContainer.style.left = `${leftPriceScaleWidth + this.DEFAULT_LEGEND_LEFT_POSITION}px`;
+    }
+    else {
+      this.legendContainer.style.left = '15px';
```
Nantawat-Poothong: Why don't use `this.DEFAULT_LEGEND_LEFT_POSITION` instead of `15`.
and should it possible to fix to use 
```javascript
this.legendContainer.style.left = `${leftPriceScaleWidth + this.DEFAULT_LEGEND_LEFT_POSITION}px`;
```
for any case instead of use if else condition?

packages/elements/src/interactive-chart/index.ts
Theeraphat-Sorasetsakul: I try to describe why I choose to subscribe sizeChange event. When we come back, you won't be confused why subscribe at x-axis event. The event has only at x-axis (timeScale). Currently, y-axis (priceScale) doesn't have it.

packages/elements/src/interactive-chart/index.ts
Theeraphat-Sorasetsakul: Agree. I will update it.

Theeraphat-Sorasetsakul: Updated.

packages/elements/src/interactive-chart/index.ts
Nantawat-Poothong: I mean, does `subscribeSizeChange` have anything to do with the axis in this comment?
Is it not confusing in the further?

Theeraphat-Sorasetsakul: Updated to add more description to be clear.

Nantawat-Poothong (approved):

PR #450, opened 2022-09-12
feat(create-efx): init project

Sarin-Udompanish:
## Description
We need to provide developer a tool to scaffolding efx element. It should be part of mono repo and publish to public npm (not with @refinitiv-ui scope name.

npx create-efx name
It will create a new folder, named `name` and it will add `efx-name` in `name` inside package.json.

npx create-efx .
It will not create a new folder but use the current directory. If the directory name doesn't have `efx-` prefix, it will add `efx-` prefix in `name` inside package.json.

Developer should be able to do `npx create-efx ../` too

Additional Notes

- The boilerplate shouldn't use any `elf-cli`. It should be using standard tool e.g. es-server, etc, or even to look at adopting vite
- We should take this opportunity to improve current template such as theme switching in demo page, how we import theme in demo page, etc
- Testing could be pretty much the same as it is today as changing to use new testing framework could be a huge task
- In the future, v7, developer can use this command to do scaffolding efx v7 element. `npx create-efx@v7 <name>`


packages/create-efx/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@types/fs-extra": "^9.0.13",
+    "@types/node": "^18.7.16",
+    "@types/prompts": "^2.0.14",
+    "chalk": "^4.1.2",
+    "fs-extra": "^10.1.0",
+    "minimist": "^1.2.6",
+    "prompts": "^2.4.2"
+  },
+  "dependencies": {
+    "@types/chalk": "^2.2.0"
```
wsuwt: Should this be a dev dependency instead ?

packages/create-efx/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  if (cmd.h || cmd.help) {
+    console.log('\nUsage: create-efx [options] [name]');
+    console.log('\nOptions:');
+    console.log('\n-h, --help           print create-efx command line options');
+    console.log('\n-v, --version        print create-efx version\n');
+    return;
+  }
+
+  if (cmd.version || cmd.version) {
```
wsuwt: Please recheck.

packages/create-efx/src/template/scripts/jsx/jsxdts-generator.js
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env node
+import fs from 'fs';
+import { fileURLToPath } from 'url';
+import path, { dirname } from 'path';
```
wsuwt: dirname is unused.

packages/create-efx/src/template/themes/halo-theme/efx-element.less
```diff
@@ -0,0 +1,29 @@
+/*
+INSTRUCTIONS
+‾‾‾‾‾‾‾‾‾‾‾‾
+- File name to be exactly the same name as your component
+- Add component's styles inside :host scope
+- Styles in top level of :host scope can be overriden by application by using standard CSS
```
Theeraphat-Sorasetsakul:
```suggestion
- Styles in top level of :host scope can be overridden by application by using standard CSS
```

Theeraphat-Sorasetsakul (dismissed):

Nantawat-Poothong (approved):

PR #469, opened 2022-09-27
chore: run SonarCloud coverage on every PRs

dtanp-rft:
Objective : Run SonarCloud coverage on every PRs.

Because we don't want to run test when we push to V6 so we still need to keep test_coverage.yml, just for this..
If we agree to run test as well on V6, we can remove test_coverage.yml and change to trigger test.yml workflow to run on V6 too.


Sakchai-Refinitiv (approved):
It works well

PR #485, opened 2022-10-04
fix(number-field): fix step=any to support integer and decimal number

Theeraphat-Sorasetsakul:
## Description

Step to reproduce:
1. `<coral-number-field value="123.123" step="any"></coral-number-field>`
2. Press step up or down
3. Decimal point is rounded up or down.

Expected:
Step="any" with decimal point should able be stay and round up or down by 1.

JIRA: https://jira.refinitiv.com/browse/ELF-2008

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step

Fixes # (ELF-2008)

## Type of change

## Checklist


documents/src/pages/elements/number-field.md
```diff
@@ -128,14 +128,16 @@ document.getElementById('mm').value = '-15';
 ```
 
 ## Set input step
-The step attribute specifies the interval between valid numbers. For instance, when `step="2"`, valid values Would only be even numbers e.g. 2,4,6,8....
+The step attribute specifies the interval between valid numbers. For instance, when `step="2"`, valid values Would only be even numbers e.g. 2,4,6,8... Alternatively, specify the step `any`. This means any value is allowed.
```
wsuwt:
```suggestion
The step attribute specifies the interval between valid numbers. For instance, when `step="2"`, valid values Would only be even numbers e.g. 2,4,6,8... Alternatively, specify the `step="any"` to allow any value. 
```

Sarin-Udompanish (approved):

PR #491, opened 2022-10-11
feat: upgrade `css-select` to version 5

wsuwt:
## Description

This is needed to avoid nth-check  <2.0.1 Severity: high

image

`css-select` V2 to v5 has no breaking change that is related to a function that we are using.


wsuwt:
Implementation of the function we are using remains the same. 

MicrosoftTeams-image (1)

Sarin-Udompanish (approved):

PR #496, opened 2022-10-11
fix(tree): update parent item in values setter

Sarin-Udompanish:
## Description
When set tree values via `values`, parent node doesn't show correct selected state because tree using setter from List that doesn't updating the parent node.

## Fixes # (issue)
Parent node displays wrong state when set values programmatically ELF-2020 

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/tree/elements/tree.ts
+++ b/packages/elements/src/tree/elements/tree.ts
@@ -7,7 +7,7 @@ import { property } from '@refinitiv-ui/core/decorators/property.js';
 import { VERSION } from '../../version.js';
 import { CollectionComposer } from '@refinitiv-ui/utils/collection.js';
 
-import { List } from '../../list/index.js';
+import { List, valueFormatWarning } from '../../list/index.js';
 import { TreeRenderer } from '../helpers/renderer.js';
 import { defaultFilter } from '../helpers/filter.js';
 import type { TreeData, TreeDataItem, TreeFilter } from '../helpers/types';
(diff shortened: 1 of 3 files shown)


packages/elements/src/tree/elements/tree.ts
```diff
@@ -425,8 +425,34 @@ export class Tree<T extends TreeDataItem = TreeDataItem> extends List<T> {
       return this.composer.getItemPropertyValue(item, 'value') as string || '';
     });
   }
-  public set values (value: string[]) {
-    super.values = value;
+  public set values (values: string[]) {
+    if (!Array.isArray(values)) {
```
Nantawat-Poothong, line 429: Why need to check Array.isArray?  values type only accept array string already, isn't it?

packages/elements/src/tree/elements/tree.ts
```diff
@@ -425,8 +425,34 @@ export class Tree<T extends TreeDataItem = TreeDataItem> extends List<T> {
...
+      valueFormatWarning.show();
+      this.values = [];
+    }
+    else {
+      // Clone value arrays
+      const newValue = values.slice();
```
Nantawat-Poothong: Why need to clone value array? you can use `const newComparison = values.sort().toString()` directly, Isn't it?

packages/elements/src/tree/elements/tree.ts
Sarin-Udompanish: It not related to the type definition. User may set values = 'some string' and we should notify an error message.

packages/elements/src/tree/elements/tree.ts
Sarin-Udompanish: You should not sort the real values.

Nantawat-Poothong: How about `const newComparison = [...values].sort().toString()` I'm not sure, what do you think?

Sarin-Udompanish: fixed

Nantawat-Poothong (dismissed):

Nantawat-Poothong (approved):

PR #498, opened 2022-10-11
Sync code to v7

bualoy-napat:
## Description

Sync code to v7


Sakchai-Refinitiv (dismissed):

package-lock.json
```diff
@@ -35401,7 +35413,7 @@
       }
     },
     "pandora-book": {
-      "version": "file:tools/pandora-book-3.0.0-66.tgz",
+      "version": "file:tools\\pandora-book-3.0.0-66.tgz",
```
Sakchai-Refinitiv, line 35416: Wh the directory separator changed

bualoy-napat: because of the window OS have different the path separator

bualoy-napat: However, I think this should not change. I will restore this point back the base version

bualoy-napat: @Sakchai-Refinitiv I already reverted this change back to the base version

Sakchai-Refinitiv (approved):

PR #512, opened 2022-10-31
docs(tutorial): replace vanilla page with bundling page

Nantawat-Poothong:
## Description
Replace vanilla page with bundling page and remove no bundling content

I didn't add content Angular legacy webpack config. Because I have tries on Angular 10 (assume using webpack 4), it's work whether import element path with /lib or not.
Ref: https://codesandbox.io/s/ef-angular10-u9v8ij?file=/src/app/app.component.ts


dtanp-rft:
@Nantawat-Poothong Cool. It's good then. I think we export path so it works. Thing that may not work is, I think, when they use webpack 5 but still use /lib/. That'd require additional config or simple change the import path.

"./heatmap": "./lib/heatmap/index.js",

dtanp-rft (approved):

PR #526, opened 2022-11-10
uncheckAllItems method not update checked state of parent node

Sarin-Udompanish:
## Description
`uncheckAllItems` method inside `tree-manager` call `uncheckItem` method recursively and inside this method has condition to prevent update timestamp of descendent items so some parent of checked items will not be re-render.

Live Demo:

https://codepen.io/mosmaall/pen/MWXpwRW?editors=1010

Steps to Reproduce:
1. Open Codepen and click `Uncheck All By Public Method` button.
2. Click ADF item
3. Click `Uncheck All By Public Method` or `Uncheck All By Set Empty values` button.

Expected Results:
Parent Item of `ADF` should not in indeterminate state.

Actual Results:
Parent Item of `ADF` is in indeterminate state.

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/tree/managers/tree-manager.ts
+++ b/packages/elements/src/tree/managers/tree-manager.ts
@@ -437,6 +437,12 @@ export class TreeManager<T extends TreeDataItem> {
    * @returns {void}
    */
   public uncheckAllItems (): void {
-    this.editableItems.forEach(item => this.uncheckItem(item));
+    this.items.forEach(item => {
+      if (this.canUncheckItem(item)) {
+        this.composer.setItemPropertyValue(item, 'selected', false);
+      }
+    });
+    // Force update of all visible parent items, making sure any indeterminate states are remove.
+    this.parentItems.forEach(item => this.updateItem(item));
   }
 }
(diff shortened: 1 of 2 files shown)


packages/elements/src/tree/managers/tree-manager.ts
```diff
@@ -437,6 +438,8 @@ export class TreeManager<T extends TreeDataItem> {
    * @returns {void}
    */
   public uncheckAllItems (): void {
-    this.editableItems.forEach(item => this.uncheckItem(item));
+    // uncheck items from top levels when manage relationships to avoid redundant re-renders
+    const items = this.manageRelationships ? this.composer.queryItems(() => true, 0) : this.checkedItems;
```
Nantawat-Poothong, line 442: I'm not sure that the difference items will cause the problem for checkAllItems and uncheckAllItems?
For example:
The `uncheckAllItems` method will force uncheck even if the item is locked.
While `checkAllItems` method can check only an editable item.

Sarin-Udompanish: It will not uncheck lock items because inside uncheckItem method has condition to check is item checkable.

Nantawat-Poothong (approved):

PR #544, opened 2022-12-01
chore: sync code to v7

dtanp-rft:


Theeraphat-Sorasetsakul:
Close because of wrong merging branch. We shouldn't directly merge from v7. So, I create a new PR at https://github.com/Refinitiv/refinitiv-ui/pull/545.

PR #547, opened 2022-12-06
fix: remove parallel=1 out of test:all because wrong test behavior

Theeraphat-Sorasetsakul:
Duplicate options `--parallel=1` causes Test coverage action run passed without test.

Example pipeline that run pass without test. https://github.com/Refinitiv/refinitiv-ui/actions/runs/3590385905/jobs/6043763189


Sakchai-Refinitiv (approved):
Look Good

PR #555, opened 2022-12-20
Update v7 docs use v7 next

dtanp-rft:
Make v7 document to use element v7 (prerelease, next dist-tags).


bualoy-napat (approved):
Look good to me

PR #567, opened 2023-01-10
Demo split design docs to separate folder

dtanp-rft:


wattachai-lseg:
Both design and implementation sections are in the same markdown file. Its content might be long but it's simple.

PR #574, opened 2023-01-23
fix: use Map to save key instead of plain object to prevent it from being replaced when a new element is created.

Sarin-Udompanish:
## Description
When an element using the translate directive is created, it will retrieve the key from the Phrasebook and store it in the `key` variable within the translate scope.

Screen Shot 2566-01-24 at 16 08 02

However, the translate directive is only called once, unless a new element with the same name is created. 
As a result, the `key` variable will be replaced each time, leading to a memory leak since the Phrasebook will only remove the most recent element reference.

## Type of change

## Checklist

### DIFF
--- a/packages/translate/src/translate.ts
+++ b/packages/translate/src/translate.ts
@@ -132,11 +132,9 @@ const observeTranslations = function (this: BasicElement, scope = this.localName
  * @param key Observe key from the `observeTranslations` function
  * @returns {void}
  */
-const disconnectTranslations = function (this: BasicElement, key: ObserverKey | undefined): void {
+const disconnectTranslations = function (this: BasicElement, key: ObserverKey): void {
   LangAttributeObserver.disconnect(this);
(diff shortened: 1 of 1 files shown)


packages/translate/src/translate.ts
```diff
@@ -154,18 +156,19 @@ const translate = function (options?: string | DecoratorOptions): TranslateFunct
     // Cannot use an element itself as a key.
     // Element may have multiple translate directives with different scope
     // Therefore we need a truly unique key
-    let key: ObserverKey;
+    const keys = new WeakMap<HTMLElement, ObserverKey>();
```
phetw:
```suggestion
    const keys = new WeakMap<BasicElement, ObserverKey>();
```

wsuwt (approved):

wsuwt:
Looks good to me. In later version we can revisit how this works again.

PR #583, opened 2023-02-02
Fix/combo-box-value-changed

wattachai-lseg:
## Description

* bug fix for `value-changed` event of combo-box with `free text` mode
* minor test case improvements espcially combo-box ones

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/combo-box/__test__/combo-box.value.test.js
+++ b/packages/elements/src/combo-box/__test__/combo-box.value.test.js
@@ -1,5 +1,5 @@
-import { fixture, expect, elementUpdated, nextFrame } from '@refinitiv-ui/test-helpers';
-import { getData, openedUpdated, snapshotIgnore, makeQueryRequest } from './utils';
+import { fixture, expect, elementUpdated, nextFrame, oneEvent } from '@refinitiv-ui/test-helpers';
+import { getData, openedUpdated, snapshotIgnore, makeQueryRequest, dispatchCustomEvent } from './utils';
 
 import '@refinitiv-ui/elements/combo-box';
 import '@refinitiv-ui/elemental-theme/light/ef-combo-box';
@@ -59,7 +59,7 @@ describe('combo-box/Value', () => {
       expect(el).shadowDom.to.equalSnapshot(snapshotIgnore);
     });
 
-    it('Free text. Set any value via API', async () => {
+    it('Free text: Set any value via API', async () => {
       const el = await fixture('<ef-combo-box free-text value="AF" opened lang="en"></ef-combo-box>');
       el.data = getData();
       await openedUpdated(el);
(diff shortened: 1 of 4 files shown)


packages/elements/src/combo-box/__test__/combo-box.value.test.js
```diff
@@ -76,7 +76,42 @@ describe('combo-box/Value', () => {
       expect(el.value).to.equal('Any', 'Value must be "Any" string');
     });
 
-    it('Free text. Reset value via API', async () => {
+    it('Free text: Set any value via API then select value in the list', async () => {
```
Tolomeo: In the test suite there is one nested `describe` only, with the "Selection by Value Attribute" description.
This test seems to care about a specific variation `free-text`, checking for 3 concerns: triggering a value change by selecting it directly in the opened list, triggering a value change by using an input and triggering a value change via property change.
It is not clear, to me at least, how this test description and its assertions match their context.
Do you think it would add value to revisit the organisation? Would that need to be executed in a different PR?

packages/elements/src/combo-box/__test__/combo-box.value.test.js
```diff
@@ -76,7 +76,42 @@ describe('combo-box/Value', () => {
...
+      // set value via attribute
+      const el = await fixture('<ef-combo-box free-text value="attribute" opened lang="en"></ef-combo-box>');
+      el.data = getData();
+      await openedUpdated(el);
```
Tolomeo, line 86: `openedUpdated` is an async function but its annotation says it returns undefined. Since the util lives in a .js file, it potentially triggers incorrect diagnostics like `'await' has no effect on the type of this expression`
Is it possible to fix the annotation in this PR?

packages/elements/src/combo-box/__test__/utils.js
```diff
@@ -12,7 +12,7 @@ export const snapshotIgnore = {
 export const openedUpdated = async (el) => {
   await elementUpdated(el);
   await nextFrame();
-  await nextFrame(); // IE11 needs a second iframe, otherwise resize observer is not run;
+  await nextFrame(); // IE11 needs another frame, otherwise resize observer is not run;
```
Tolomeo, line 16: 👍

packages/elements/src/combo-box/__test__/combo-box.value.test.js
wattachai-lseg: I've moved "Free Text mode" test cases into their own `describe`. It might not be perfect but revisiting test case organization is beyond the scope of this PR.

packages/elements/src/combo-box/__test__/combo-box.value.test.js
wattachai-lseg: Good catch! I've updated JSDoc of the function. The diagnostic should be correct now.

wattachai-lseg:
@Tolomeo could you check my latest commits and resolve our conversations if they address the concerns? Lastly, an approval would be nice :)

Tolomeo (approved):

Tolomeo:
💥 PR approved, Thanks for the changes
PS: I mistakenly approved before resolving conversations. Maybe that is why it doesn't let me resolve them now.

dtanp-rft (approved):

PR #603, opened 2023-02-27
refactor(calendar): remove ie11 BC year workaround

Nantawat-Poothong:
## Description
IE11 series: remove specific IE11 code from `ef-calendar`

Fixes # (issue)
https://jira.refinitiv.com/browse/STG-33

## Type of change

### DIFF
--- a/packages/elements/src/calendar/index.ts
+++ b/packages/elements/src/calendar/index.ts
@@ -749,19 +749,15 @@ export class Calendar extends ControlElement implements MultiValue {
           break;
         }
         return;
-      case 'Up': // IE11
       case 'ArrowUp':
         void this.onNavigation('ArrowUp');
         break;
-      case 'Down':
       case 'ArrowDown':
         void this.onNavigation('ArrowDown');
         break;
-      case 'Left':
       case 'ArrowLeft':
         void this.onNavigation('ArrowLeft');
         break;
-      case 'Right':
       case 'ArrowRight':
         void this.onNavigation('ArrowRight');
         break;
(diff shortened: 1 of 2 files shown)


packages/elements/src/calendar/index.ts
```diff
@@ -1030,16 +1025,9 @@ export class Calendar extends ControlElement implements MultiValue {
    */
   private viewFormattedDate (segment: DateSegment, includeMonth = false): TemplateResult {
     const year = segment.year;
-    const isBC = year <= 0;
-    const includeEra = isBC;
+    const includeEra = year <= 0;
```
wsuwt: I feel that previous code is a bit more clear when specifying `isBC`. What do you think ?

Nantawat-Poothong: I think the TranslateDirective will receive the variable name `includeEra` anyway.

how about this?
```suggestion
    const isBC = year <= 0;
```

  and return  with option `{ date, includeMonth, includeEra: isBC }`

wsuwt: I think it's better than before. Looks good.

Sarin-Udompanish:
Should we remove `formatLocaleDate` method from the code ?

wsuwt:
Are we able to remove `case: 'Esc'` as well ? Otherwise LGTM.

Nantawat-Poothong:
> Are we able to remove `case: 'Esc'` as well ? Otherwise LGTM.

Ah, I didn't see it, thanks for reminding me.

Sarin-Udompanish (approved):

PR #607, opened 2023-02-27
refactor(text-field): remove IE11 specific code

wsuwt:
## Description

Remove IE11 specific code

Fixes https://jira.refinitiv.com/browse/STG-25

## Type of change

### DIFF
--- a/packages/elements/src/text-field/index.ts
+++ b/packages/elements/src/text-field/index.ts
@@ -66,7 +66,6 @@ export class TextField extends FormFieldElement {
       :host {
         display: inline-block;
       }
-
       :host(:focus), :host input:focus {
         outline: none;
       }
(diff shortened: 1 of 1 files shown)


packages/elements/src/text-field/index.ts
```diff
@@ -231,11 +230,11 @@ export class TextField extends FormFieldElement {
 
   /**
    * @param error existing state of error
-   * @returns true if there is no error and browser is IE11 and minLength more than 0 and value exists
+   * @returns true if there is no error and minLength more than 0 and value exists
    */
   /* c8 ignore start */
   protected shouldValidateForMinLength (error: boolean): boolean {
-    return !!(!error && isIE && this.minLength && !!this.value);
+    return !!(!error && this.minLength && !!this.value);
```
Nantawat-Poothong: If I understand correctly, I think function `shouldValidateForMinLength`  can be remove.

wsuwt: Thanks. I double check and it's removable as `checkValidity()` would already validate against the constraints provided.

Nantawat-Poothong (approved):
LGTM

PR #617, opened 2023-03-07
refactor(chart): migrate theme config to plugin

Theeraphat-Sorasetsakul:
## Description

Migrate current theme applying code into plugin. To make it clean and readable.

Fixes # ([STG-198](https://jira.refinitiv.com/browse/STG-198))

## Type of change


packages/elements/src/chart/helpers/legend.ts
```diff
@@ -1,86 +1,97 @@
...
-//   /**
-//    * Force return to string type because Chart.js doesn't support typescript
-//    * and some type of @types/chart.js isn't correct
-//    */
-//   return (color && color !== 'transparent') ? color as string : undefined;
-// };
+const isValidColor = (color: Color): Color | undefined => {
+  /**
+   * Force return to string type because Chart.js doesn't support typescript
+   * and some type of @types/chart.js isn't correct
```
wattachai-lseg: This comment and logic looks outdated. Could you double check this?

Theeraphat-Sorasetsakul:
Legend logic has been changed to use generateLabels from chartjs. The doc needs to be update.
- New behavior will show transparent color. Replace to the old that if color is transparent, then use other valid color. For example borderColor='transparent' then pick 'backgroundColor' instead.
- New behavior won't use point color when set usePointStyle=true. The old behavior will use point color even if bar chart doesn't have type of point color.

packages/elements/src/chart/helpers/legend.ts
Theeraphat-Sorasetsakul: This file has been removed due to the legend logic will use from chart.js instead.

packages/elements/src/chart/index.ts
```diff
@@ -10,19 +10,30 @@ import { customElement } from '@refinitiv-ui/core/decorators/custom-element.js';
...
 import {
   merge,
   MergeObject
 } from './helpers/index.js';
 
 import {
-  Chart as ChartJS
+  Chart as ChartJS,
+  Plugin,
+  ChartType,
```
wattachai-lseg, line 26: `ChartType` is unused.

packages/elements/src/chart/index.ts
```diff
@@ -82,6 +104,9 @@ export class Chart extends BasicElement {
       plugins: {
         title: {
           display: false
+        },
+        efchart: {
```
wattachai-lseg: `ef-chart` instead per [chart.js doc](https://www.chartjs.org/docs/latest/developers/plugins.html#plugin-id?)?

packages/elements/src/chart/index.ts
```diff
@@ -116,6 +141,14 @@ export class Chart extends BasicElement {
     return '';
   }
 
+  /**
+   * Safely returns a dataset array
```
wattachai-lseg: Could you remove "Safely"? As the dataset is being returned as refernce not value, mutating the returned dataset creates side effects. Side effects are considered unsafe.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: It currently uses to let chart know the plugin. If we remove this, Typescript will say error that chartjs.options doesn't know the plugin.
```
declare module 'chart.js' {
  interface PluginOptionsByType<TType extends ChartType> {
    efchart?: {
      enable: boolean
    }
  }
}
```

wattachai-lseg: Indeed! TypeScript is acting up on my machine again. My apology.

packages/elements/src/chart/index.ts
```diff
@@ -148,6 +181,77 @@ export class Chart extends BasicElement {
...
+   * @returns {Plugin} plugin
+   */
+  private plugin (): Plugin {
+    const cssStyle = getComputedStyle(this);
+    // TODO: Try and remove the need for global object modification.
+    // It's easier to cover all areas by modifying the global object,
+    // however, if possible, we should look to try and just modify local configs.
+
+    // Set font globals
+    ChartJS.defaults.color = cssStyle.getPropertyValue('color');
```
wattachai-lseg: shall we move global config update out of the plugin?

packages/elements/src/chart/index.ts
```diff
@@ -148,6 +181,77 @@ export class Chart extends BasicElement {
     this.destroyChart();
   }
 
+  /**
+   * Create plugin to set our theme into chartjs lifecycle
+   * @returns {Plugin} plugin
+   */
+  private plugin (): Plugin {
```
wattachai-lseg: let's name function with verb for use getter pattern instead.

packages/elements/src/chart/index.ts
```diff
@@ -148,6 +181,77 @@ export class Chart extends BasicElement {
...
+      beforeUpdate: this.decorateColors
+    };
+  }
+
+  /**
+   * Themable parts of the config.
+   * This will be merged into the configuration object.
+   * @returns {ChartOptions} chart config with theme
+   */
+  protected get themableConfig (): ChartOptions {
```
wattachai-lseg: `themableChartOption` could be a better name.

packages/elements/src/chart/index.ts
```diff
@@ -160,6 +264,157 @@ export class Chart extends BasicElement {
     }
   }
 
+  /**
+   * Get as CSS variable and tries to convert it into a usable number
+   * @returns {(number|undefined)} The value as a number, or, undefined if NaN.
+   */
+  protected cssVarAsNumber (...args: string[]): number | undefined {
```
wattachai-lseg, line 276: this looks very "helper" too me.

packages/elements/src/chart/index.ts
```diff
@@ -181,6 +436,37 @@ export class Chart extends BasicElement {
...
+  protected generateColors (isArray: boolean, amount: number, shift: number): DatasetColors {
+    let color;
+    const solid = [];
+    const opaque = [];
+    const alpha = Number(this.getComputedVariable('--fill-opacity', '0.2'));
+
+    amount = isArray ? amount : 1;
+
+    for (let i = shift; i < amount + shift; i++) {
+      color = this.colors[i % this.colors.length];
```
wattachai-lseg: `color` can be `const`

packages/elements/src/chart/index.ts
```diff
@@ -10,19 +10,30 @@ import { customElement } from '@refinitiv-ui/core/decorators/custom-element.js';
 import { property } from '@refinitiv-ui/core/decorators/property.js';
 import { query } from '@refinitiv-ui/core/decorators/query.js';
 import { VERSION } from '../version.js';
+import { color as parseColor } from '@refinitiv-ui/utils/color.js';
+
+type DatasetColors = {
```
wattachai-lseg: move this to `types` helper?

packages/elements/src/chart/index.ts
```diff
@@ -160,6 +264,157 @@ export class Chart extends BasicElement {
...
+          colors = this.generateColors(true, dataset.data ? dataset.data.length : 1, index);
+          borderColor = isMultipleDatasets ? this.getComputedVariable('--multi-dataset-border-color', '#fff') : colors.solid;
+          backgroundColor = this.config?.type === 'polarArea' ? colors.opaque : colors.solid;
+          if (!dataset.borderColor) {
+            dataset.borderColor = borderColor;
+          }
+          if (!dataset.backgroundColor) {
+            dataset.backgroundColor = backgroundColor;
+          }
+          // Add more color if items doesn't enough
```
wattachai-lseg: pulral subject with sigular verb, grammar error here.

packages/elements/src/chart/index.ts
```diff
@@ -160,6 +264,157 @@ export class Chart extends BasicElement {
...
+          }
+          if (!(dataset as LineControllerDatasetOptions).pointBackgroundColor) {
+            (dataset as LineControllerDatasetOptions).pointBackgroundColor = colors.solid;
+          }
+          if (!(dataset as LineControllerDatasetOptions).pointBorderColor) {
+            (dataset as LineControllerDatasetOptions).pointBorderColor = colors.solid;
+          }
+          break;
+
+        // This type, Colors is set to an array.
```
wattachai-lseg:
```suggestion
        // These types, Colors is set to an array.
```

packages/elements/src/chart/index.ts
```diff
@@ -160,6 +264,157 @@ export class Chart extends BasicElement {
...
+          // Add more color if items doesn't enough
+          if (Array.isArray(dataset.borderColor) && Array.isArray(borderColor) && dataset.borderColor.length < borderColor.length) {
+            merge(dataset.borderColor, borderColor);
+          }
+          if (Array.isArray(dataset.backgroundColor) && Array.isArray(backgroundColor) && dataset.backgroundColor.length < backgroundColor.length) {
+            merge(dataset.backgroundColor, backgroundColor);
+          }
+          break;
+
+        // This type, Colors is possible to be string or array
```
wattachai-lseg:
```suggestion
        // These types, Colors could be string or array
```

packages/elements/src/chart/index.ts
```diff
@@ -160,6 +264,157 @@ export class Chart extends BasicElement {
...
+          // Add more color if items doesn't enough
+          if (Array.isArray(dataset.borderColor) && Array.isArray(borderColor) && dataset.borderColor.length < borderColor.length) {
+            merge(dataset.borderColor, borderColor);
+          }
+
+          if (Array.isArray(dataset.backgroundColor) && Array.isArray(backgroundColor) && dataset.backgroundColor.length < backgroundColor.length) {
+            merge(dataset.backgroundColor, backgroundColor);
+          }
+          break;
+        // For the other chart
```
wattachai-lseg:
```suggestion
        // For other types
```

packages/elements/src/chart/index.ts
```diff
@@ -170,7 +425,7 @@ export class Chart extends BasicElement {
     }
 
     // TODO: unknown type
```
wattachai-lseg: time to fix this unknown type!

packages/elements/src/chart/index.ts
```diff
@@ -160,6 +264,157 @@ export class Chart extends BasicElement {
...
+        });
+      }
+
+      return legends;
+    }
+
+    if (ChartJS.defaults.plugins.legend.labels.generateLabels) {
+      legends = ChartJS.defaults.plugins.legend.labels.generateLabels(chart);
+    }
+    legends.forEach((legend, i) => {
```
wattachai-lseg, line 402: call: only fillStyle is not covered by the built-in generateLabels fn?

packages/elements/src/chart/__demo__/index.html
```diff
@@ -493,24 +392,26 @@
...
               }
             }
           }
         }
       };
     </script>
   </demo-block>
-
-  <demo-block layout="normal" header="Doughnut Chart">
+-->
```
wattachai-lseg: we don't need this, right?

packages/elements/src/chart/index.ts
```diff
@@ -160,6 +264,157 @@ export class Chart extends BasicElement {
...
+      legend.lineWidth = Number(datasets[i].borderWidth) || 0;
+      switch (datasets[i].type || chartType) {
+        case 'line':
+        case 'radar':
+        case 'scatter':
+          legend.fillStyle = (datasets[i] as LineControllerDatasetOptions).borderColor as Color;
+          break;
+        // For the other chart
```
wattachai-lseg:
```suggestion
        // For other chart types
```

wattachai-lseg (changes requested):
Please check my response.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: It can't move to helper because this method use getComputedVariable. And getComputedVariable method can access only within a Class. 
```typescript
Property 'getComputedVariable' is protected and only accessible within class 'BasicElement' and its subclasses.ts(2445)
```

wattachai-lseg: Oh I missed that one. Marking as resolved.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated to be ef-chart

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Agree with you. So I updated to remove the word.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Agree that inside Plugin should not do anything globally.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: I renamed to be createPlugin. Is it good?

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated to be below.
// Add more colors if items aren't enough

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Fixed. I implemented merge function to receive type assertion.

packages/elements/src/chart/index.ts
wattachai-lseg: I don't think we should move global updates to `decorateConfig`: the function runs during chart render & create chart.  The global update should run only once.

packages/elements/src/chart/index.ts
wattachai-lseg: my bad, it should be
` these types, Colors are set to an array.`

packages/elements/src/chart/index.ts
wattachai-lseg: With this complexity, I think we better stick with `unknown` assertion. What do you think?

packages/elements/src/chart/index.ts
wattachai-lseg: `createPlugin` LGTM but it should not be the getter pattern.

wattachai-lseg (changes requested):
a few more things.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated by moving out to connectedCallback, And I create a new function `setGlobalConfig` to set global config.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: I am not sure with my new code too. So I agree to reverting the code.
Updated.

wattachai-lseg (approved):
LGTM

PR #623, opened 2023-03-10
fix(card): Card can not expand nested overlay-menu in touch devices

Nantawat-Poothong:
## Description
Card can't expand nested overlay-menu in touch devices.
In mouse devices it work fine when hover the parent item to expand sub items.
But not for touch devices, it should tap the parent item to expand the sub items.

Root cause:
- Card always close the overlay-menu when trigger item

Solution: 
- Add condition to close the overlay-menu only when trigger an item with value (That mean parent item shouldn't have a value)

Fixes # (issue)
https://jira.refinitiv.com/browse/STG-241

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/card/__test__/card.test.js
+++ b/packages/elements/src/card/__test__/card.test.js
@@ -10,7 +10,7 @@ import '@formatjs/intl-pluralrules/locale-data/en';
 import '@refinitiv-ui/elements/card';
 import '@refinitiv-ui/elemental-theme/light/ef-card';
 
-const menuData = [{ label: 'Spain', value: 'Spain'  }, { label: 'France',value: 'France', disabled: true }, { label: 'Italy', value: 'Italy' }];
+const menuData = [{ label: 'Spain', value: 'Spain'  }, { label: 'France',value: 'France', disabled: true }, { label: 'Italy', value: 'Italy' }, { label: 'Other', items: [{ label: 'Thailand', value: 'Thailand' }] }];
 
 describe('card/Card', () => {
   describe('DOM structure', () => {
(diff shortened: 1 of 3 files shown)


packages/elements/src/card/__test__/card.test.js
```diff
@@ -82,6 +82,36 @@ describe('card/Card', () => {
       await oneEvent(el, 'item-trigger');
       expect(menu.opened).to.equal(false, 'Menu should close when item is selected');
     });
+    it('Should open menu and not close when clicked on non value item', async () => {
```
wattachai-lseg: Let's avoid arrow functions as described in the [doc](https://mochajs.org/#arrow-functions). Note that, I don't think we should refactor everything. Just make sure that we don't use arrow functions on the new test cases.

packages/elements/src/overlay-menu/index.ts
```diff
@@ -1029,7 +1029,7 @@ export class OverlayMenu extends Overlay {
       .label=${label}
       .subLabel=${subLabel}
       .icon=${icon}
-      .value=${value}
+      .value=${ifDefined(value)}
```
wattachai-lseg: `ifDefined()` is no-op here as `.value` is a property not attribute.
https://lit.dev/docs/templates/directives/#ifdefined

We can remove it.

packages/elements/src/card/index.ts
```diff
@@ -155,10 +155,11 @@ export class Card extends BasicElement {
 
   /**
    * Close menu
+   * @param event Custom event
```
wattachai-lseg:
```suggestion
   * @param event ItemTriggerEvent
```

wattachai-lseg (changes requested):
Just a few things.

packages/elements/src/overlay-menu/index.ts
Nantawat-Poothong: if `ifDefined` has removed the value that return from item will be `string` 'undefined' instead of `null`

wattachai-lseg (approved):
LGTM

PR #632, opened 2023-03-20
v7 Simplify Web Test Runner config and CLI

Sakchai-Refinitiv:
## Description

v7 Simplify Web Test Runner config and CLI

- Prevent yargs CLI options crash with Web Test Runner CLI by using Node API
- Simplify CLI options by using sharable options for package and element CLI
- Simplify Web Test Runner config by moving yargs CLI out of the configs
- Show element name on BrowserStack if running test on a specific element

## Type of change

## Checklist

### DIFF
--- a/cli.mjs
+++ b/cli.mjs
@@ -8,7 +8,6 @@ import { PACKAGES_ROOT, errorHandler, getJSON } from './scripts/helpers/esm.mjs'
 
 const argvNoBin = hideBin(process.argv);
 const argv = yargs(argvNoBin)
-  .scriptName("refinitiv-ui")
   .command('$0 <reflect> <package>', 'reflect the command', yargs => {
     yargs.positional('reflect', {
       describe: 'npm command to reflect',
--- a/cli.mjs
+++ b/cli.mjs
@@ -10,7 +10,6 @@ const nodeArgv = process.argv.filter(item => {
   return true;
 });
 
-console.log(nodeArgv);
 const argvNoBin = hideBin(nodeArgv);
 const argv = yargs(argvNoBin)
   .scriptName("refinitiv-ui")
--- a/cli.mjs
+++ b/cli.mjs
@@ -35,13 +35,16 @@ try {
   // For workspace package real name is required
   const packageName = (await getJSON(path.resolve(PACKAGES_ROOT, workspace, 'package.json'), import.meta)).name;
   const command = ['npm', 'run', argv.reflect, `--workspace=${packageName}`];
-  elementName && command.push(elementName);
+
+  // Add element name to command
+  if (workspace === 'elements') {
+    command.push(elementName ? elementName : 'elements');
+  }
+
   options.length > 0 && command.push('--')
   command.push(...options);
 
-  execSync(command.join(' '), {
-    stdio: 'inherit'
-  });
+  execSync(command.join(' '), { stdio: 'inherit' });
 }
 catch (error) {
   errorHandler(error);
--- a/cli.mjs
+++ b/cli.mjs
@@ -6,11 +6,7 @@ import yargs from 'yargs/yargs';
 import { hideBin } from 'yargs/helpers';
 import { PACKAGES_ROOT, errorHandler, getJSON } from './scripts/helpers/esm.mjs';
 
-const nodeArgv = process.argv.filter(item => {
-  return true;
-});
-
-const argvNoBin = hideBin(nodeArgv);
+const argvNoBin = hideBin(process.argv);
 const argv = yargs(argvNoBin)
   .scriptName("refinitiv-ui")
   .command('$0 <reflect> <package>', 'reflect the command', yargs => {
@@ -24,7 +20,7 @@ const argv = yargs(argvNoBin)
     })
   })
   .demandCommand()
-  .help(false)
+  .help()
   .argv
 
 const options = argvNoBin.slice(2);
--- a/cli.mjs
+++ b/cli.mjs
@@ -6,8 +6,14 @@ import yargs from 'yargs/yargs';
 import { hideBin } from 'yargs/helpers';
 import { PACKAGES_ROOT, errorHandler, getJSON } from './scripts/helpers/esm.mjs';
 
-const argvNoBin = hideBin(process.argv);
+const nodeArgv = process.argv.filter(item => {
+  return true;
+});
+
+console.log(nodeArgv);
+const argvNoBin = hideBin(nodeArgv);
 const argv = yargs(argvNoBin)
+  .scriptName("refinitiv-ui")
   .command('$0 <reflect> <package>', 'reflect the command', yargs => {
     yargs.positional('reflect', {
       describe: 'npm command to reflect',
@@ -19,7 +25,7 @@ const argv = yargs(argvNoBin)
     })
   })
   .demandCommand()
-  .help()
+  .help(false)
   .argv
 
 const options = argvNoBin.slice(2);
--- a/cli.mjs
+++ b/cli.mjs
@@ -9,12 +9,10 @@ import { PACKAGES_ROOT, errorHandler, getJSON } from './scripts/helpers/esm.mjs'
 const argvNoBin = hideBin(process.argv);
 const argv = yargs(argvNoBin)
   .command('$0 <reflect> <package>', 'reflect the command', yargs => {
-    yargs.require('reflect')
     yargs.positional('reflect', {
       describe: 'npm command to reflect',
       type: 'string'
     })
-    yargs.require('package')
     yargs.positional('package', {
       describe: 'Package or element name',
       type: 'string'
--- a/packages/elements/scripts/cmd/test.mjs
+++ b/packages/elements/scripts/cmd/test.mjs
@@ -2,6 +2,7 @@
 import { execSync } from 'node:child_process';
 import { useTestOptions } from '../../../../scripts/tests/cli-options.mjs';
 import { getElements, errorHandler } from '../helpers/index.mjs';
+import { hideBin } from 'yargs/helpers';
 
 const elements =  ['elements', ...getElements()];
 
@@ -20,8 +21,8 @@ export const builder = yargs => {
   useTestOptions(yargs);
 };
 export const handler = (argv) => {
-  // Get params and forward to main test file
-  let params = process.argv.slice(3);
+  // Remove command and forward params only to main test file
+  let params = hideBin(process.argv).slice(1)
 
   try {
     // Build before run test everytime.
--- a/packages/elements/scripts/cmd/test.mjs
+++ b/packages/elements/scripts/cmd/test.mjs
@@ -20,8 +20,8 @@ export const builder = yargs => {
   // Use shared test options for the CLI
   useTestOptions(yargs);
 };
-export const handler = (argv) => {
-  // Remove command and forward params only to main test file
+export const handler = () => {
+  // Remove command and forward all test options to main test file
   let params = hideBin(process.argv).slice(1)
 
   try {
--- a/packages/elements/scripts/cmd/test.mjs
+++ b/packages/elements/scripts/cmd/test.mjs
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import { execSync } from 'node:child_process';
-import { DefaultBrowsers, BrowserStack } from '../../browsers.config.mjs';
+import { DefaultBrowsers, BrowserStack } from '../../../../browsers.config.mjs';
 import {
   getElements,
   info,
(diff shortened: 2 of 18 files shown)


scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+const testAll = packageName === 'all' || packageName === undefined;
+const basePath = path.join(PACKAGES_ROOT, testAll ? '*' : packageName);
+const watch = argv.watch;
+const snapshots = argv.updateSnapshots;
+const browserstack = argv.browserstack && !watch;
+const testCoverage = argv.includeCoverage;
+let testTarget = packageName;
+
+// Environment variables to use for overriding config script
+env.testCoverage = testCoverage;
```
wattachai-lseg: I don't think we need to use environment variable here. If the export has to be dynamic, we could export a function with a param instead.

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+
+// Test on BrowserStack`
+if (browserstack) {
+  const sharedCapabilities = {
+    'browserstack.user': env.BROWSERSTACK_USERNAME,
+    'browserstack.key': env.BROWSERSTACK_ACCESS_KEY,
+    project: env.BROWSERSTACK_PROJECT_NAME || 'Refinitiv UI',
+    name: testTarget,
+    build: `build ${env.BROWSERSTACK_BUILD || 'unknown'}`,
+    timeout: 1800, // Maximum
```
wattachai-lseg: intead of "maximum" implied by the prop name `timeout`, we could provide the number unit e.g. time in milliseconds

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+  const launchers = [];
+  argv.browserstack.forEach((option) => {
+    switch (option) {
+      case 'default':
+        BrowserStack.defaultBrowsers.forEach(browser => launchers.push(BrowserStack.config[browser]));
+        break;
+      case 'supported':
+        BrowserStack.supportedBrowsers.forEach(browser => {
+          // Disable testing on Safari, we have to check all test cases are passed before enabling it again
+          if (!browser.includes('safari')) {
```
wattachai-lseg: we should check for `safari_minus1` & `safari_minus2` as well. Also, we're going to enable testing on Safari this sprint. Does it mean this check would be removed very soon?

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+    build: `build ${env.BROWSERSTACK_BUILD || 'unknown'}`,
+    timeout: 1800, // Maximum
+  };
+
+  // Add BrowserStack launchers to config
+  const launchers = [];
+  argv.browserstack.forEach((option) => {
+    switch (option) {
+      case 'default':
+        BrowserStack.defaultBrowsers.forEach(browser => launchers.push(BrowserStack.config[browser]));
```
wattachai-lseg, line 81: why don't we exclude Safari in case of `default`? Plus, safari could be passed as `browserstack` argument direcly. We haven't handled such case.

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+    }
+  });
+
+  // Add BrowserStack launchers to config
+  const browsers = [];
+  launchers.forEach(launcher => {
+    browsers.push(browserstackLauncher({ capabilities: { ...sharedCapabilities, ...launcher } }));
+  })
+
+  config.browsers = browsers; // Set all browsers to use BrowserStack
```
wattachai-lseg, line 98: "Add BrowserStack launchers to config" should apply to this whole section (line 96 to line 101). "Set all browsers to use BrowserStack" seems unrelated and confusing here.

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+  testTarget = target;
+  env.testElement = target;
+} else if (packageName === 'elements') {
+  env.testElement = 'all';
+}
+
+const config = {
+  ...wtrConfig,
+  files: [path.join(basePath , '/__test__/**/*.test.js')],
+  watch: argv.watch,
```
wattachai-lseg:
```suggestion
  watch
```

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+  const browsers = [];
+  launchers.forEach(launcher => {
+    browsers.push(browserstackLauncher({ capabilities: { ...sharedCapabilities, ...launcher } }));
+  })
+
+  config.browsers = browsers; // Set all browsers to use BrowserStack
+}
+
+// Specify browser to run the unit test & convert browser naming to playwright's one
+const optionBrowser = argv.browsers;
```
wattachai-lseg: let's move this to line 25 section where arguments are extracted into variables.

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// Strip argv (options) out to prevent them pass to the web-test-runner cli
+process.argv = process.argv.slice(0, 2);
+
+info(watch ? `Start Dev Server: ${ testTarget }` : `Test: ${ testTarget }`);
+
+if (snapshots) {
+  info(`Update and prune snapshots: ${ testTarget }`);
+  process.argv.push('--update-snapshots');
```
wattachai-lseg, line 128: Could you add a comment explaining that `update-snapshots` must be passed as  `process.argv` only?

scripts/tests/run.mjs
```diff
@@ -0,0 +1,148 @@
...
+        break;
+      default:
+        browser = configBrowser.product;
+        break;
+    }
+    return optionBrowser.includes(browser);
+  });
+}
+
+// Strip argv (options) out to prevent them pass to the web-test-runner cli
```
wattachai-lseg:
```suggestion
// Strip argv (options) out to prevent web-test-runner picking them up
```

scripts/tests/cli-options.mjs
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env node
+
+import { BrowserStack, DefaultBrowsers } from '../../browsers.config.mjs';
+
+/**
+ * Use standard options for the yargs CLI
+ * @param {Argv} yargs argv
+ * @returns {Argv} yargs argv
+ */
+export const useTestOptions = (argv) => {
```
wattachai-lseg: this looks like a builder function. the first param is `yargs` not `argv`

packages/elements/scripts/cmd/test.mjs
```diff
@@ -1,94 +1,36 @@
...
-      type: 'string',
-      alias: 'o',
-      default: 'full',
-      choices: ['full', 'minimal'],
-      description: 'Print output to the console'
-    })
     .completion('completion', () => elements);
+
+  // Use shared test options for the CLI
+  useTestOptions(yargs);
```
wattachai-lseg, line 21: missing `return` statement
```suggestion
  return useTestOptions(yargs);
```

packages/elements/scripts/cmd/test.mjs
```diff
@@ -1,94 +1,36 @@
...
-  const browsers = argv.browsers ? argv.browsers.join(' '): '';
-  const browserstack = argv.browserstack ? argv.browserstack.join(' ') : '';
-
-  info(watch ? `Start Dev Server: ${ element }` : `Test: ${ element }`);
-
-  if (snapshots) {
-    info(`Update and prune snapshots: ${ element }`);
-  }
+  // Remove command and forward params only to main test file
+  let params = hideBin(process.argv).slice(1)
```
wattachai-lseg, line 25: yargs provides arguments for us with `argv` param. We should use it instead.

cli.mjs
```diff
@@ -8,6 +8,7 @@ import { PACKAGES_ROOT, errorHandler, getJSON } from './scripts/helpers/esm.mjs'
 
 const argvNoBin = hideBin(process.argv);
 const argv = yargs(argvNoBin)
+  .scriptName("refinitiv-ui")
```
wattachai-lseg: I don't think there is any command with such name.

packages/elements/web-test-runner.config.mjs
wattachai-lseg, line 1: it seems this file is not in use. Did I miss something?

wattachai-lseg (changes requested):
I have quite a number of concerns. Hope we can resolve them swiftly.

packages/elements/web-test-runner.config.mjs
Sakchai-Refinitiv: This config is use for running test element, it inherits options from the parent config

wattachai-lseg: Could you point me to its usage location?

packages/elements/scripts/cmd/test.mjs
Sakchai-Refinitiv: This method just modifies the object, so it does not need to return.

packages/elements/scripts/cmd/test.mjs
Sakchai-Refinitiv: The argv form yargs is already converted to an object that makes it harder to forward to the child process

scripts/tests/cli-options.mjs
Sakchai-Refinitiv: If you see the type in yargs it is `Yargs.Argv` interface

scripts/tests/run.mjs
wattachai-lseg: need to use it as WTR would search for its config in `elements` package. Environment variable would be the only way to pass value to the config.
Let's mention `elements` package here in the comment.

cli.mjs
Sakchai-Refinitiv: This is optional I add it to prevent confusion because we have the two cli.mjs files

packages/elements/web-test-runner.config.mjs
Sakchai-Refinitiv: Behavior is the same test group https://modern-web.dev/docs/test-runner/cli-and-configuration/#test-groups

scripts/tests/run.mjs
wattachai-lseg: Testing and ensuring green test result of Safari would be tracked separately. Marking as resolved.

scripts/tests/run.mjs
Sakchai-Refinitiv: Updated a comment.

scripts/tests/run.mjs
wattachai-lseg: Testing and ensuring green test result of Safari would be tracked separately. Marking as resolved.

scripts/tests/run.mjs
Sakchai-Refinitiv: Removed

scripts/tests/run.mjs
Sakchai-Refinitiv: Removed

scripts/tests/run.mjs
Sakchai-Refinitiv: Updated, Safari included to the default browsers in v7.

scripts/tests/run.mjs
Sakchai-Refinitiv: Fixed

scripts/tests/run.mjs
Sakchai-Refinitiv: Moved

scripts/tests/run.mjs
Sakchai-Refinitiv: Web Test Runner does not provide a config to update snapshots, so the CLI option is the only way.  I already added a code comment.

scripts/tests/run.mjs
Sakchai-Refinitiv: Updated

scripts/tests/cli-options.mjs
Sakchai-Refinitiv: Renamed to yargs

scripts/tests/run.mjs
Sakchai-Refinitiv: Update comment.

scripts/tests/run.mjs
wattachai-lseg: this `timeout` is optional?

cli.mjs
wattachai-lseg: image

the upper one show a result with `scriptName`. The lower one show a result without `scriptName`.

We should remove this `scriptName()` as it only creates confusion to anyone using `--help` option. 
While `node refinitiv-ui test core` will never work, `node cli.mjs test core` will do the job.

wattachai-lseg (changes requested):
only 2 more locations.

scripts/tests/run.mjs
Sakchai-Refinitiv: The `BrowserstackLauncher` class always overwrites `timeout` to 300 which means it does not need `timeout` here.

cli.mjs
Sakchai-Refinitiv: That's right. removed

wattachai-lseg (approved):
LGTM

PR #635, opened 2023-03-22
Documentation local components bundling

Tolomeo:
## Description

This PR adds a local bundle to the documentation build which is useful to support live code examples of components not yet deployed.
The bundle is created through a new documentation package script which was also queued to the build one.
Please see `sub-checkbox` page as a provided example of the working result.
Open doubts to be discussed:
- the entry file for esbuild is more a source than a script, although currently placed in the `scripts` folder
- esbuild compilation files end up in the resources folder, where other static assets are tracked. Depending on the answer to the first point, they should be added to gitignore
- the usage of a stylesheet link to import bundled css is due to the impossibility of using a @import in a css block, because that would need to be placed before any other css to work, but Pandora always puts default styles on top. It would improve this solution to modify that behaviour and have Pandora place defaults in a separate style tag, leaving to docs css blocks full control of their own

Fixes # (issue)
STG-293

## Type of change

## Checklist


Sarin-Udompanish:
@Tolomeo , I wanted to let you know that I made a small change to the code. I switched the theme import from the `tokens` package to `workspace-theme` instead.

For the first point, I agree that it would be better to put `component.js` inside `resources/components` and the bundle file inside `resources/components/bundles`. We can ignore the bundle file in `.gitignore` to prevent unnecessary commits.

As for the last point, I made some changes to the theme that I think make it unnecessary to use @import in a CSS block anymore. However, I'm not entirely sure if I understood your concern correctly. Can you please try pulling the latest code then run `npm run build` on the root project and checking the results again to make sure that everything is working as expected?

dtanp-rft:
Guys. This looks great! I will update the content to reflect current state of development e.g. add all components to the doc. I will do this ASAP so we could have a good baseline for the doc.

dtanp-rft:
I have committed initial content.

dtanp-rft (approved):

Sarin-Udompanish (approved):

PR #643, opened 2023-03-29
Updates demo to include controls and improve a11y

TremayneChrist:
This PR updates the demo page to include POC controls and adds more components to the sign-in pattern.

This fixes the previous accessibility issues and improves screen reader support.

Changes:
- Adds POC controls to demo page
- Adds more components to sign in pattern
- Adds `aria-hidden` to hide invaluable information from screen reader
- Adds `aria-hidden` temorary hack to icon. See next steps.
- Updates tokens to load them in correct order
- Partially fixes select `value-changed` event

Next steps:
- Update select to fix value get/set properly
- Update select to place chevron inside of button then add `aria-hidden` to icon.
- Remove icon `aria-hidden` hack
- Remove multi-level `aria-labelledby`/`aria-describedby`
- See if we need to remove A11y from `FormFieldElement`
- Stop using tap event, or fix it to support screen reader commands


wsuwt (approved):

wsuwt:
Merging this to `poc-dev` cc. @Sarin-Udompanish

PR #653, opened 2023-04-10
Docs: Remove IE from support browser

Nantawat-Poothong:
## Description

Just remove IE tab from support browser (บรัยส์)

Fixes # (issue)
https://jira.refinitiv.com/browse/STG-28


bualoy-napat (approved):
Look good to me

PR #668, opened 2023-04-26
refactor(chart): refactor import style and duplicate code

Theeraphat-Sorasetsakul:
## Description

- refactor module import arrangement
- refactor title code to using template
- fix to support plugins
- rephrase one sentence on doc 

## Type of change

## Checklist

### DIFF
--- a/documents/src/pages/elements/chart.md
+++ b/documents/src/pages/elements/chart.md
@@ -208,7 +208,7 @@ line.config = {
 ## Update chart data or configurations
 To update chart datasets or configurations, you can modify the value in the `config` property and then call `updateChart()`.
 
-However, you do not have to call `updateChart()` if you set a new `config` object for the chart.
+However, if you want to renew a `config` object to the chart, you do not need to call 'updateChart()'.
 
 ::
 ```javascript
--- a/documents/src/pages/elements/chart.md
+++ b/documents/src/pages/elements/chart.md
@@ -208,7 +208,7 @@ line.config = {
 ## Update chart data or configurations
 To update chart datasets or configurations, you can modify the value in the `config` property and then call `updateChart()`.
 
-However, if you want to renew a `config` object to the chart, you do not need to call 'updateChart()'.
+However, you do not have to call `updateChart()` if you set a new `config` object for the chart.
 
 ::
 ```javascript
--- a/packages/elements/src/chart/helpers/index.ts
+++ b/packages/elements/src/chart/helpers/index.ts
@@ -1,1 +1,2 @@
+export * from './types.js';
 export * from './merge.js';
--- a/packages/elements/src/chart/index.ts
+++ b/packages/elements/src/chart/index.ts
@@ -523,6 +523,7 @@ export class Chart extends BasicElement {
     // Stop any chart.js animations
     this.chart.stop();
     this.mergeConfigs();
+    this.requestUpdate();
 
     // Update the chart
     this.chart?.update(updateMode);
--- a/packages/elements/src/chart/index.ts
+++ b/packages/elements/src/chart/index.ts
@@ -36,7 +36,6 @@ import {
   DatasetColors
 } from './helpers/index.js';
 
-import type { Header } from '../header';
 import '../header/index.js';
 
 const CSS_COLOR_PREFIX = '--chart-color-';
@@ -535,11 +534,8 @@ export class Chart extends BasicElement {
    * @returns Header template from title of config
    */
   protected get titleTemplate (): TemplateResult | undefined {
-    if (this.chartTitle) {
-      return html`
-        <ef-header>${this.chartTitle}</ef-header>
-      `;
-    }
+    return this.chartTitle
+      ? html`<ef-header>${this.chartTitle}</ef-header>` : undefined;
   }
 
   /**
--- a/packages/elements/src/chart/index.ts
+++ b/packages/elements/src/chart/index.ts
@@ -13,35 +13,31 @@ import { VERSION } from '../version.js';
 import { color as parseColor } from '@refinitiv-ui/utils/color.js';
 
 import {
-  merge,
-  MergeObject
-} from './helpers/index.js';
-import type {
-  DatasetColors
-} from './helpers/types';
-
-import {
-  Chart as ChartJS,
-  Plugin,
-  ChartType,
+  ChartConfiguration,
   ChartDataset,
+  Chart as ChartJS,
+  ChartOptions,
+  ChartType,
   Color,
-  LineControllerDatasetOptions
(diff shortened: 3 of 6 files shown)


documents/src/pages/elements/chart.md
```diff
@@ -208,7 +208,7 @@ line.config = {
 ## Update chart data or configurations
 To update chart datasets or configurations, you can modify the value in the `config` property and then call `updateChart()`.
 
-However, you do not have to call `updateChart()` if you set a new `config` object for the chart.
+However, if you want to renew a `config` object to the chart, you do not need to call 'updateChart()'.
```
wattachai-lseg: I find the original one easier to understand. Let's hear what Mos thinks about this.

packages/elements/src/chart/helpers/index.ts
```diff
@@ -1 +1,2 @@
+export * from './types.js';
```
wattachai-lseg, line 1: would be part of `ef-chart` config? I can't find its usage in a user's perspective.

packages/elements/src/chart/index.ts
```diff
@@ -443,9 +428,21 @@ export class Chart extends BasicElement {
       return;
     }
 
+    let plugins: Plugin[] = [
+      this.createPlugin(),
+      doughnutCenterPlugin
+    ];
+    
+    if (this.config.plugins && this.config.plugins?.length > 0) {
```
wattachai-lseg: with `?`, there is no need to do turthy check on `this.config.plugins`.
```suggestion
    if (this.config.plugins?.length > 0) {
```

packages/elements/src/combo-box/index.ts
```diff
@@ -1219,10 +1218,11 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
       // benefit of being localised too
       if (this.focused || selectionLength > 1) {
         return html`
-        <ef-counter part="selection-badge" tabindex="-1" .value=${selectionLength} title=${ifDefined(selectionLength > 999 ? selectionLength.toLocaleString() : undefined)} max="999"></ef-counter>
-      `;
+          <ef-counter part="selection-badge" tabindex="-1" .value=${selectionLength} title=${ifDefined(selectionLength > 999 ? selectionLength.toLocaleString() : undefined)} max="999"></ef-counter>
```
wattachai-lseg: This line is too long. Let's add a new line to each atrribute.

wattachai-lseg (changes requested):
Just a few points.

wattachai-lseg (changes requested):
Just a few points.

documents/src/pages/elements/chart.md
Theeraphat-Sorasetsakul: Mos agree to use the old message. So I revert it back.

packages/elements/src/chart/helpers/index.ts
Theeraphat-Sorasetsakul: I don't think user need this type. This file only contains DatasetColors type which used in the chart element.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: I try your suggestion but typescript doesn't allow this. It may be that because of I try to using Spread syntax (...). So I updated this condition by checking it is as Array.

packages/elements/src/combo-box/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/helpers/index.ts
wattachai-lseg: I see. Let's remove this export.

packages/elements/src/chart/index.ts
wattachai-lseg: that makes sense.  Let's remove `?` as shown below as plugin can't be `falsy` anymore thanks to `isArray()`

```typescript
if (Array.isArray(this.config.plugins) && this.config.plugins.length > 0) {
```

wattachai-lseg (changes requested):
2 more things.

packages/elements/src/chart/helpers/index.ts
Theeraphat-Sorasetsakul: I think we can't remove it. It uses in index.ts.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: Updated.

packages/elements/src/chart/helpers/index.ts
wattachai-lseg: Your're right. My bad.

wattachai-lseg (approved):
LGTM

packages/elements/src/chart/index.ts
```diff
@@ -609,7 +592,7 @@ export class Chart extends BasicElement {
   protected render (): TemplateResult {
     return html`
       <div container>
```
wsuwt, line 594: Not sure why we were using `container` attribute to style `div`

packages/elements/src/combo-box/index.ts
```diff
@@ -1219,10 +1218,17 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
       // benefit of being localised too
       if (this.focused || selectionLength > 1) {
         return html`
-        <ef-counter part="selection-badge" tabindex="-1" .value=${selectionLength} title=${ifDefined(selectionLength > 999 ? selectionLength.toLocaleString() : undefined)} max="999"></ef-counter>
-      `;
+          <ef-counter
+            part="selection-badge"
+            tabindex="-1"
```
wsuwt, line 1223: `counter` is a `BasicElement` which do not have `tabIndex`. So I think we probably don't need this.

packages/elements/src/combo-box/index.ts
```diff
@@ -1219,10 +1218,17 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
...
+            .value=${selectionLength}
+            title=${ifDefined(selectionLength > 999 ? selectionLength.toLocaleString() : undefined)}
```
wsuwt, line 1225: Can we extract `999` to a constant.

packages/elements/src/chart/index.ts
Theeraphat-Sorasetsakul: It specify style in internal template like below.
```
[container] {
        position: absolute;
        top: 0;
        right: 0;
        bottom: 0;
        left: 0;
        display: flex;
        flex-direction: column;
}
```

PR #669, opened 2023-04-26
refactor(icon,flag): code cleanup

wattachai-lseg:
## Description

- icon,flag: remove irrelevant comment about polyfill browsers
- flag: mark `src` as deprecated
- icon,flag: replace templates when necessary only
- code cleanup

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/icon/index.ts
+++ b/packages/elements/src/icon/index.ts
@@ -4,8 +4,7 @@ import {
   css,
   CSSResultGroup,
   TemplateResult,
-  SVGTemplateResult,
-  PropertyValues
+  SVGTemplateResult
 } from '@refinitiv-ui/core';
 import { customElement } from '@refinitiv-ui/core/decorators/custom-element.js';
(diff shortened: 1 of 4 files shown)


packages/elements/src/icon/index.ts
```diff
@@ -113,16 +114,11 @@ export class Icon extends BasicElement {
...
-  protected firstUpdated (changedProperties: PropertyValues): void {
-    super.firstUpdated(changedProperties);
-    /**
-     * We have to call this here because
-     * polyfilled browsers only get variables at this point.
-     */
+  public connectedCallback (): void {
+    super.connectedCallback();
     this.setPrefix();
   }
```
TremayneChrist: This doesn't change much and means we set the prefix on every DOM connection. I'm not sure this is any better than before. Maybe we just leave it as it is?

wattachai-lseg: Indeed. I did confirm with others, let's keep it as is.

wsuwt (approved):

PR #694, opened 2023-05-17
feat(icon): introduce a new ef-config for ef-icon

Nantawat-Poothong:
## Description

Added an element ef-config to provide the data that consume by other elements

Fixes # (issue)
https://jira.refinitiv.com/browse/STG-423

## Type of change

## Checklist


packages/elements/src/config/elements/config.ts
```diff
@@ -0,0 +1,47 @@
...
+   * Element version number
+   * @returns version number
+   */
+  static get version (): string {
+    return VERSION;
+  }
+
+  @provide({ context: efConfig })
+  @state()
+    config: ConfigType = {};
```
Tolomeo: We don't provide a fully structure default config here. Does that create any problems with consumers?

packages/elements/src/icon/index.ts
```diff
@@ -112,13 +124,22 @@ export class Icon extends BasicElement {
     }
   }
 
+  /**
+   * Check if the autosuggest has content
+   * @returns content exists
+   */
+  private get iconMap (): string | null {
+    return this.icon && this.config?.icon?.map[this.icon] || null;
```
Tolomeo: This looks too interesting to be good.
Is there a way we can avoid making use of all those optional chaining?

Tolomeo (commented):
The PR looks nice and the changes are contained, as expected.
We could add tests and expand on the demo page though.
I left a couple (probably related) comments

Tolomeo:
As discussed earlier, there are a couple of concerns still to be fixed:
- **Context component naming**: we are currently making use of ef-config but there are doubts on that. An appreciated alternative seems to be `ef-provider`. Please try to discuss with the extended team on what could be a valid decision for this, but keep in mind that the final naming should reflect the intent of the component to provide shared configurations freely consumed by ef components (for example in the future we may decide to funnel the ui theme through the context).
- **EfConfig properties**: let’s follow already defined best practices to define them. That means we will require the developer to specify `config`.
- **Testing**: the test so far checks for the icon component receiving a context value. That doesn’t reflect the functionality provided though and the test should feed the context with a specified value and then assert on the rendered icon html matching an expected content. The idea is to test the feature using a black box approach.

PR #695, opened 2023-05-18
chore: update import paths in demo files

Sarin-Udompanish:
## Description

- Remove elemental theme from demo-block.
- Change import paths on demo files to import the themes from lib folder instead.
- Update theme-extractor script to accept path to element folder as a argument. 
- Remove inject native-styles in theme-extractor.
- Update build script of halo-theme and solar-theme packages to include theme-extractor script.


scripts/release/theme-extractor.mjs
```diff
@@ -89,8 +108,13 @@ const extractThemeDependency = (themePath) => {
...
-  return themeContent
-    .match(importRegex)
+  const matchedImports = themeContent.match(importRegex);
+
+  if (!matchedImports) {
+    return [];
+  }
+
+  return matchedImports
     .filter((matched) => !matched.includes('native-elements'))
```
wsuwt: We can safely remove this line.

scripts/release/theme-extractor.mjs
```diff
@@ -2,13 +2,12 @@
...
+import { hideBin } from 'yargs/helpers';
 
 import { log, errorHandler, success, ROOT, getJSON } from '../helpers/esm.mjs';
 import { ELEMENT_DIST, PACKAGE_ROOT, getElementList, getElementTagName } from './util.cjs';
 
-// List of themes to be extracted
-const THEMES = ['halo', 'solar'];
-
 // Element package scope
 const PACKAGE_NAME = (await getJSON(`${PACKAGE_ROOT}/package.json`, import.meta)).name;
```
wsuwt: This is the reason why theme extractor is currently giving wrong output, 

`import '@refinitiv-ui/halo-theme/icon/themes/halo/dark';`

instead of 

`import '@refinitiv-ui/elements/icon/themes/halo/dark';`

because it reads package name from `package.json` that it's executing. We could hard code this value as elements as it is not going to change so often.

wsuwt (approved):

PR #703, opened 2023-05-23
ci: add documents and __demo__ to .nxignore

Theeraphat-Sorasetsakul:
## Description

Ignore to files that should not run. Because it has nothing to test.

Fixes # (https://jira.refinitiv.com/browse/STG-436)

## Type of change


.nxignore
```diff
@@ -1 +1,6 @@
-samples/
\ No newline at end of file
+# develop process
+__demo__/
+
+# document files
```
wsuwt:
```suggestion
```

.nxignore
```diff
@@ -1 +1,6 @@
-samples/
\ No newline at end of file
+# develop process
```
wsuwt:
```suggestion
```

wsuwt (approved):

PR #710, opened 2023-05-25
feat(configuration): introduce new element with a support icon offline feature

Nantawat-Poothong:
## Description

Added an element ef-configuration to provide the data that consume by other elements

Fixes # (issue)
https://jira.refinitiv.com/browse/STG-423

## Type of change

## Checklist


documents/src/pages/elements/configuration.md
```diff
@@ -0,0 +1,47 @@
...
+<ef-configuration id="config">
+  <ef-icon icon="tick-base64"></ef-icon>
+</ef-configuration>
+```
+```javascript
+  const config = document.getElementById('config');
+  config.icon = { map: { 
+    "tick-base64": 'data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTYiIGhlaWdodD0iMTYiIHZpZXdCb3g9IjAgMCAxNiAxNiIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48cGF0aCBzdHJva2U9IiMwMDAiIGQ9Ik0xNCA0bC04LjI1IDguMjVMMiA4LjUiIGZpbGw9Im5vbmUiIGZpbGwtcnVsZT0iZXZlbm9kZCIvPjwvc3ZnPg==',
+  }}
+```
```
Tolomeo: Let's re-check the copy in this file.
We should also document how the only type of icon accepted by the icon map is a base64 encoded svg

packages/elements/src/configuration/__test__/configuration.test.js
```diff
@@ -0,0 +1,27 @@
+import { expect, fixture, nextFrame } from '@refinitiv-ui/test-helpers';
+import { tickSvgBase64, tickSvg, isEqualSvg } from './helpers/helpers.js'
+import '@refinitiv-ui/elements/configuration';
+import '@refinitiv-ui/elements/icon';
+import '@refinitiv-ui/elemental-theme/light/ef-icon.js';
+
+describe('configuration/Configuration', () => {
```
Tolomeo, line 4: Didn't we agree to move the tests in the icon tests suite?
Also, we should add at least one test covering for the icon rendering without being wrapped in the context. That will make sure we are not breaking the functionality for users not implementing ef-configuration

packages/utils/src/loader/svg-loader.ts
```diff
@@ -7,6 +7,13 @@ import { CDNLoader } from './cdn-loader.js';
  */
 const isUrl = (str: string): boolean => (/^(https?:\/{2}|\.?\/)/i).test(str);
 
+/**
+ * Checks a string to see if it's a base64 URL
+ * @param str String to test
+ * @returns is Base64
+ */
+const isBase64svg = (str: string): boolean => (/^(data:image\/[svg,svg+xml]+;base64,[a-zA-Z0-9,+,/]+={0,2})/i).test(str);
```
Tolomeo: Very terse 👍

packages/elements/src/configuration/elements/configuration.ts
```diff
@@ -0,0 +1,61 @@
...
+   * @ignore
+   */
+  @provide({ context: efConfig })
+  @state()
+  public config: ConfigType = DEFAULT_CONFIG;
+
+  @property({ type: Object })
+  set icon (icon: ConfigIcon) {
+    this.config = { ...this.config, icon };
+  }
```
Tolomeo: On this, I was made aware that we have predefined guidelines for the definition of public apis for components.
Please have a look at https://ui.refinitiv.com/custom-components/naming-conventions
We should align to those then

packages/utils/src/loader/svg-loader.ts
Nantawat-Poothong: You really mean it, don't you? 🤔

packages/elements/src/configuration/elements/configuration.ts
Nantawat-Poothong: Where are the predefined guidelines in the link? I have found nothing 😅

packages/utils/src/loader/svg-loader.ts
Tolomeo: Yes, I like it for real ❤️

packages/elements/src/configuration/elements/configuration.ts
Tolomeo: The link contains naming conventions already in use throughout the library.
It happens that we already use a `config` property which may be suitable to expose here.
Screenshot 2023-06-01 091605

Nantawat-Poothong: So that means we could be setting configuration though out config and icon properties?

packages/utils/src/loader/svg-loader.ts
Nantawat-Poothong: Thanks for this regex101 topic, there really save my day. hehe
https://regex101.com/library/XAQEkg?filterFlavors=javascript&filterFlavors=golang&orderBy=MOST_RECENT&search=

Nantawat-Poothong: By the way, this regular expression cause the 2 code smells on this PR 😅

packages/elements/src/configuration/elements/configuration.ts
Tolomeo: We should pass the entire config to the component. I understand it is a bit verbose.
I see 2 options: use `config` with the whole config obj  (we could also provide to consumers a utility fn that they will use to define only what they need and the fn will fill the rest with defaults), or we can discuss with the team whether they are fine with this component api, even if it doesn't strictly adhere to conventions

packages/utils/src/loader/svg-loader.ts
Tolomeo: Let's see what they say. Actually, by looking at it more carefully, the regex may be not entirely correct

packages/elements/src/configuration/__test__/configuration.test.js
dtanp-rft: is it too late to disagree? :) I rather feel that all configuration test should set in configuration. Otherwise, when we do more config in the future, the configuration test will be scattered all over the place.

Tolomeo: It is not too late, don't worry
The logic these tests cover live in the icon component, so related tests should be placed there. If we keep tests here, when a bug is introduced in icon a test in a different test suite will fail, when the logic changes the developer will need to reach for tests in a different component to update etc.
I think what makes people think these tests should be placed here is just the title. We should just create a new describe block in the icon test suite

Nantawat-Poothong: Is it too late for me to explain? 55555
@Tolomeo we had discussed this on the Bangkok site and agreed to move these test cases under the icon element.
But for the configuration element, we think it should have some unit test as well.

dtanp-rft: Yep. My bad :) Thanks for Atom for clarification.

packages/elements/src/icon/__test__/icon.test.js
```diff
@@ -13,234 +14,257 @@ import {
...
-      expect(checkRequestedUrl(server.requests, uniqueInvalidIconSrc)).to.equal(true, 'should try to request invalid icon');
-      expect(preloadedIcons[0].length > 0).to.equal(true, 'Should successfully preload icon by name with CDN prefix');
-      expect(preloadedIcons[1].length > 0).to.equal(true, 'Should successfully preload icons with src');
-      expect(preloadedIcons[2], 'Should not preload invalid icon').to.be.undefined;
-      el.setAttribute('icon', firstUniqueIcon);
-      await elementUpdated(el);
-
-      expect(server.requests.length).to.equal(3, 'no new requests should be made since icons are already preloaded');
+  
+    it('should not pass config when to icon incorrectly', async () => {
```
Tolomeo: The description is not very clear, can we re-phrase it?

packages/elements/src/icon/index.ts
```diff
@@ -112,13 +124,22 @@ export class Icon extends BasicElement {
     }
   }
 
+  /**
+   * Check if the autosuggest has content
```
Tolomeo: Are you sure the comment is relevant to this code?

Tolomeo (dismissed):
💥 
I left just a couple of comments addressing minor concerns

dtanp-rft (approved):

PR #715, opened 2023-05-29
fix(interactive-chart): reduce canvas memory size for safari

bualoy-napat:
## Description
There was an issue about the memory limit in Safari that cause some canvas to not render. So, I was purposing the workaround to reduce the size of the canvas that accumulates in memory and decrease chance of the canvas reaching the limit.

Fixes # (issue) 
STG-430

## Type of change

## Checklist


bualoy-napat:
As we discussed, it is better to handle in the application's lifecycle. The developer should resize the element size to 1px*1px before the page is unloaded to reduce the size of the canvas that accumulates in memory. Moreover, this issue will be fixed soon by the WebKit team (https://github.com/WebKit/WebKit/pull/14375).

PR #746, opened 2023-06-13
fix(chart): fix radar scale backdrop to be false

Theeraphat-Sorasetsakul:
## Description

- Fix Radar scale backdrop to be false.
- Update doughnutCenterLabel to export serveral types.


packages/elements/src/chart/plugins/doughnut-center-label.ts
```diff
@@ -19,25 +19,25 @@ interface Selectable {
...
+export interface CenterLabelConfig {
   defaultText: CenterLabel[];
   selected: Selected;
   onRenderLabel(chart: ChartJS, chartItems: ActiveElement[]): CenterLabel[];
 }
 
 declare module 'chart.js' {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  interface PluginOptionsByType<TType extends ChartType> {
+  interface PluginOptionsByType<TType extends ChartType = ChartType> {
```
wattachai-lseg, line 40: do we need this default value?

wattachai-lseg (commented):
1 thing.

packages/elements/src/chart/plugins/doughnut-center-label.ts
Theeraphat-Sorasetsakul: I just update to follow suggestion in Chart.js doc. https://www.chartjs.org/docs/latest/developers/plugins.html#typescript-typings

Theeraphat-Sorasetsakul: Updated to both interfaces.

wattachai-lseg (approved):

PR #751, opened 2023-06-14
Chore/ci-cache-per-workflow-only

wattachai-lseg:
## Description

* use  node18.x with action/setup-node built-in package data cache
* cache node_modules, built result and test coverage scoped with a workflow only

## Type of change

- [x] Refactor (improves code without changing its functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have checked my code and corrected any misspellings

PR #736, opened 2023-06-08
chore: syncs change from v6

bualoy-napat:
Syncs change from v6

PR #752, opened 2023-06-14
Temp-test-ci-caching-update

wattachai-lseg:
## Description

testing new ci caching update of #751 with change from #736


wattachai-lseg:
Test successful.

PR #757, opened 2023-06-16
fix(number-field): fix Floating point Division on validate method

Theeraphat-Sorasetsakul:
## Description

There is a red border around the input for valid numbers.

**Steps to replicate:**
1. Open https://elf.int.refinitiv.com/elements/coral-number-field.html page and find sandbox with element like this:
<coral-number-field no-spiner step="0.00001"></coral-number-field>
2. Enter a number to the input filed, like 111 then blur(click somewhere else) the input. Then repeat the previous step: 111111 and blur.

Fixes # ([STG-457](https://jira.refinitiv.com/browse/STG-457))

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/number-field/__test__/number-field.validity.test.js
+++ b/packages/elements/src/number-field/__test__/number-field.validity.test.js
@@ -1,4 +1,4 @@
-import { fixture, expect } from '@refinitiv-ui/test-helpers';
+import { fixture, expect, elementUpdated } from '@refinitiv-ui/test-helpers';
 
 import '@refinitiv-ui/elements/number-field';
 import '@refinitiv-ui/elemental-theme/light/ef-number-field';
@@ -22,5 +22,15 @@ describe('number-field/Validity', () => {
       const el = await fixture('<ef-number-field min="1" max="5" value="4.5"></ef-number-field>');
       expect(el.checkValidity()).to.be.equal(false);
     });
+  });
+});
+
+describe('Check Floating point', function () {
+  // Test Floating point precision issue that results approximation of real number. e.g. 1111111/0.00001 should equal to 111111100000. 
+  it('Input remains valid upon value update with a step of float value', async function () {
+    const el = await fixture('<ef-number-field step="0.00001" value="111111"></ef-number-field>');
+    el.value="1111111";
+    await elementUpdated(el);
+    expect(el.checkValidity()).to.be.equal(true);
   });
 });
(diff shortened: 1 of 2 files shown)


packages/elements/src/number-field/__test__/number-field.validity.test.js
```diff
@@ -24,3 +24,13 @@ describe('number-field/Validity', () => {
     });
   });
 });
+
+describe('Check Floating point', function () {
+  // Test Floating point precision issue that results approximation of real number. e.g. 1111111/0.00001 should equal to 111111100000. 
+  it('Input remains valid upon value update with a step of float value', async function () {
+    const el = await fixture('<ef-number-field step="0.00001" value="111111"></ef-number-field>');
```
wsuwt: Can we remove this 
```suggestion
    const el = await fixture('<ef-number-field step="0.00001"></ef-number-field>');
```
and add a comment that value change is needed in order to trigger validation. Also can we have at least one more test case with different number and decimal points ?

packages/elements/src/number-field/index.ts
```diff
@@ -532,6 +532,29 @@ export class NumberField extends FormFieldElement {
...
   }
 
+  /**
+   * Count precision number
+   * @param number value to count
+   * @returns precision number
+   */
+  private getPrecision (number: number): number {
+    const getDecimalPrecision = (number: string | number): number => {
+      const [whole, part] = number.toString().split('.');
```
wsuwt: Would this make it a bit more readable ?
```suggestion
      const [wholeNumber, decimalNumber] = number.toString().split('.');
```

packages/elements/src/number-field/__test__/number-field.validity.test.js
Theeraphat-Sorasetsakul: Added more test cases.

packages/elements/src/number-field/index.ts
```diff
@@ -532,6 +532,29 @@ export class NumberField extends FormFieldElement {
     return 0;
   }
 
+  /**
+   * Count precision number
+   * @param number value to count
+   * @returns precision number
+   */
+  private getPrecision (number: number): number {
+    const getDecimalPrecision = (number: string | number): number => {
```
wsuwt:
```suggestion
    const getDecimalPrecision = (number: string): number => {
```

packages/elements/src/number-field/index.ts
```diff
@@ -532,6 +532,29 @@ export class NumberField extends FormFieldElement {
...
+      const [wholeNumber, decimalNumber] = number.toString().split('.');
```
wsuwt:
```suggestion
      const [wholeNumber, decimalNumber] = number.split('.');
```

wsuwt (approved):

PR #761, opened 2023-06-20
ci: fix lerna to use from local

Theeraphat-Sorasetsakul:
## Description

Found that https://github.com/Refinitiv/refinitiv-ui/actions/runs/5318623946 this release failed because of using wrong lerna version. So it need to fix Lerna to use from our local dependency

image

## Type of change


documents/package.json
```diff
@@ -11,7 +11,7 @@
     "start:dev": "concurrently \"pandora src dist --watch\" \"serve dist -s\"",
     "start": "serve dist -s",
     "copy:resources": "cp src/_redirects dist/_redirects && cp -r resources dist/resources && cp src/theme-loader.js dist/theme-loader.js",
-    "build:api": "lerna run --scope @refinitiv-ui/elements api-analyzer --stream && node ./scripts/element.injector.js",
+    "build:api": "npx lerna run --scope @refinitiv-ui/elements api-analyzer --stream && node ./scripts/element.injector.js",
```
wattachai-lseg: don't think we need `npx` here.

Theeraphat-Sorasetsakul:
```suggestion
    "build:api": "lerna run --scope @refinitiv-ui/elements api-analyzer --stream && node ./scripts/element.injector.js",
```

wattachai-lseg (approved):

PR #768, opened 2023-06-22
chore(icon): remove preload deprecated notice

Nantawat-Poothong:
## Description

Remove preload deprecated notice, because preload is used by internal elements


packages/elements/src/icon/utils/IconLoader.ts
```diff
@@ -11,13 +10,6 @@ const iconLoaderInstance = new IconLoader();
...
 
-/**
-   * Deprecation notice displays a warning message
-   * when deprecated features are used.
-  */
-const deprecationNotice = new DeprecationNotice('Icon `preload()` is deprecated.');
-
-
 /**
  * @deprecated Icon `preload()` is deprecated.
```
wattachai-lseg: let's remove this line as well.

wattachai-lseg (approved):

PR #776, opened 2023-06-23
chore(text-field, password-field, email-field): update demo show error state

Nantawat-Poothong:
## Description

The text field element demo page error state style has disappeared due to the element not processing error validation as expected

** UPDATE
Found that the password-field and the email-field should update as well.

Solution:
Delay 50ms then set error true to show the error state for the elements


dtanp-rft (changes requested):
- Please fix code style e.g. missing semi colon
- Instead of explicit set `error`, I think we should setup field control with validation and use timeout to set invalid value. With this, we can both test error styles + validation logic of component. For example, text-field could be setup as below.
 
```html
<ef-text-field
  id="nickname"
  aria-describedby="error-text"
  pattern="[a-z]{4,8}"
  placeholder="Must be lowercase letters between 4-8 characters">
</ef-text-field>
```

Nantawat-Poothong:
> * Please fix code style e.g. missing semi colon

I agree.
> * Instead of explicit set `error`, I think we should setup field control with validation and use timeout to set invalid value. With this, we can both test error styles + validation logic of component. For example, text-field could be setup as below.

I have tried to do this but an invalid value alone can not change an element with either `disabled` or `readonly` attribute to show the error state.
.
FYI, checkValidity() with `disabled` or `readonly` attribute would always return true without doing anything else.
Ref: https://www.w3.org/TR/2011/WD-html5-20110405/association-of-controls-and-forms.html#dom-cva-checkvalidatity

dtanp-rft:
Shame but we could set disable and readonly after we set value? :D

Nantawat-Poothong:
> Shame but we could set disable and readonly after we set value? :D

Of course, it could be set disabled and readonly later but it still does not meet the idea to represent the test error styles and validation logic of the component as well 😵‍💫
.
IMO, I prefer to delay the error state rather than delay to set the disabled state or the readonly state later.

dtanp-rft:
> > Shame but we could set disable and readonly after we set value? :D
> 
> Of course, it could be set disabled and readonly later but it still does not meet the idea to represent the test error styles and validation logic of the component as well 😵‍💫 . IMO, I prefer to delay the error state rather than delay to set the disabled state or the readonly state later.

Yep, that's fine for me. The idea of error disable & readonly is to see if the color is right. Thanks Atom.

dtanp-rft (approved):

PR #781, opened 2023-06-26
refactor(utils): remove local cache utils

Nantawat-Poothong:
## Description

We provide the local cache to reduce network requests to solve too many network request problems for SVG loader that are related to ef-icon and ef-flag.
But when the ef-icon changed to using the sprite to improve loading performance then the local cache is no longer needed.

one element that is still used is the ef-flag but seems like the ef-flag doesn't have too much benefit from it. so we decide to remove the local cache utils because have/haven't local cache will not significant effect on the performance like the images below

### Solution: 
Remove the localcache utils

Fixes # (issue)
https://jira.refinitiv.com/browse/STG-424

## Type of change

#### Test with 258 flags
Screenshot 2566-06-23 at 14 09 34

#### Test results

|  Test | Duration  |
|---|---|
| 1. Cold start with local cache utils <br /> 2. Warm start with local cache utils  |  2.94s <br /> 1.45s |
|  3. Without local cache utils |  1.39s |

Screenshot 2566-06-23 at 13 51 58 copy

##### 1. Cold start with local cache utils
Screenshot 2566-06-23 at 14 07 08

##### 2. Warm start with local cache utils
Screenshot 2566-06-23 at 14 07 28

##### 3. Without local cache utils
Screenshot 2566-06-23 at 14 07 41


Theeraphat-Sorasetsakul (commented):
I think we doesn't use "idb" anymore. Could you remove it and update package.json and package-lock.json

Nantawat-Poothong:
> I think we doesn't use "idb" anymore. Could you remove it and update package.json and package-lock.json

Thanks, updated!

Theeraphat-Sorasetsakul (approved):
LGTM

PR #785, opened 2023-06-27
docs(number-field): show error state and update validate section

bualoy-napat:
## Description

Refactor content in input validity section and live demo

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/documents/src/pages/elements/number-field.md
+++ b/documents/src/pages/elements/number-field.md
@@ -132,7 +132,6 @@ To validate input from users, `ef-number-field` provides similar features to a n
 
 @> Validation of user input of `ef-number-field` is consistent with a native input. [See native input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number).
 
-
 Whenever input is invalid, the error attribute will be added to the element. You can use the `error` property to check if input is currently in the error state.
 
 ```html
--- a/documents/src/pages/elements/number-field.md
+++ b/documents/src/pages/elements/number-field.md
@@ -157,11 +157,6 @@ You can add the event listener `error-changed` to the element and it will dispat
 const element = document.getElementById('input');
 const errorChangedText = document.getElementById('error-text');
 element.value = "-1";
-
-setTimeout(() => {
-  element.reportValidity();
-}, 1000);
-
 element.addEventListener('error-changed', (e) => {
   if (e.detail.value) {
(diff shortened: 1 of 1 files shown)


documents/src/pages/elements/number-field.md
```diff
@@ -155,7 +156,11 @@ You can add the event listener `error-changed` to the element and it will dispat
 ::number-field::
 const element = document.getElementById('input');
 const errorChangedText = document.getElementById('error-text');
-element.value = "-1"
+element.value = "-1";
+
+setTimeout(() => {
+  element.reportValidity();
+}, 1000);
```
Nantawat-Poothong: I think setTimeout 1000ms is a bit unreal (too much delay time) if users really need to verify the input value. do you have any smooth approach to call reportValidity()? or just use 50ms instead.

dtanp-rft: Actually, this code doesn't make sense to show in docs. I'd prefer to just mention about `input.reportValidity()`. Do we really need to show this live sample?

dtanp-rft (changes requested):

documents/src/pages/elements/number-field.md
Nantawat-Poothong: I agree to not add reportValidity() to the live sample. 
I talked with Nut that setTimeout to call reportValidity() is so weird either, users may usually call reportValidity() inside their app lifecycle.

bualoy-napat:
@dtanp-rft In my opinion, it would be good if we can show the users how to apply input.reportValidity() during initialization. However, after I reconsider about this again, it would be better not to include it in the docs because we have already mentioned it in the first paragraph of this topic. Thank you for your suggestion.

documents/src/pages/elements/number-field.md
bualoy-napat: In my opinion, it would be good if we can show the users how to apply `input.reportValidity()` during initialization. However, after I reconsider about this again, it would be better not to include it in the docs because we have already mentioned it in the first paragraph of this topic. Thank you for your suggestion.

dtanp-rft: @Nantawat-Poothong Yeah exactly. Also, we dont' need to have live demo in everywhere.

dtanp-rft (approved):

PR #803, opened 2023-07-04
Refactor Regex and unit test in number field

bualoy-napat:
## Description

Simplify the pattern Regex and add some unit tests to cover the Regex pattern

[STG-490](https://jira.refinitiv.com/browse/STG-490)

## Type of change

## Checklist


bualoy-napat:
I already verified that `pattern` is unnecessary and not effected to the behavior of the screen reader. Moreover, we implemented our own validity check API instead of native API. Therefore, we will remove `pattern` from number-field.

PR #811, opened 2023-07-11
chore: update `mocha` version to latest version that support `IE11`

wsuwt:
## Description

Our test helper loads old version of mocha. It's got a dependency with security vulnerability in it. We should look to upgrade to avoid this and also modernise our development tech stack as much as possible.

- Update `mocha` to version 9, the latest version that support IE11
- Removing `esm` package and `--require esm` in mocha cli. ESM is natively supported by Node 16 and above. 

Below changes are in sync with `v7`

- Replacing `require` with `import`
- Changing test files extension to `.test.js` for consistency.
- Making `@refinitiv-ui/phrasebook` type module to support `mocha@9` which is an ESM package.

## Type of change

## Checklist


packages/phrasebook/__test__/locale.test.js
```diff
@@ -64,7 +64,7 @@ const assembleImports = () => {
       let componentsList = getDirList(`./${DEFAULT_LANG_DIR}/${lang}`);
       if(componentsList) {
         for (const component of componentsList) {
-          let imported = require(`../${DEFAULT_LANG_DIR}/${lang}/${component}`);
+          let imported = import(`../${DEFAULT_LANG_DIR}/${lang}/${component}`);
```
wattachai-lseg: dynamic import returns `Promise`. Shall we `await` here?

packages/test-helpers/package.json
```diff
@@ -28,8 +28,8 @@
     "access": "public"
   },
   "dependencies": {
-    "@open-wc/testing": "^3.0.0-next.5",
-    "@types/mocha": "^5.2.7",
-    "mocha": "^6.2.3"
+    "@open-wc/testing": "^3.2.0",
+    "@types/mocha": "^9.1.1",
```
wattachai-lseg, line 32: `@types/mocha` should be devDep instead?

wattachai-lseg (changes requested):
2 things plus a merge conflict.

packages/test-helpers/package.json
wsuwt: I think it should remain in dependencies. We ship mocha and type as a dependencies to users (mainly us).

packages/phrasebook/__test__/locale.test.js
wsuwt: I'm not sure if this is being resolved on `done()` after this function is called or not but this is how `v7` is currently. Let me do a bit more investigation on this.

wattachai-lseg: indeed, `v7` uses `await`. I still think this is the way. We're better off without `done()` 😄

wsuwt: Updated. I'll create another PR for `v7` to use `async/await` as well.

wattachai-lseg (approved):

PR #819, opened 2023-07-13
style: format code with prettier

wsuwt:
## Description

- integrates prettier
- format code at every precommit


packages/core/src/directives/template-map.ts
```diff
@@ -68,20 +72,18 @@ const setMapped = (element: Element, name: string, value: unknown, oldValue: unk
...
-      else {
+      } else {
         (element as VariableElement)[name] = value;
       }
       break;
     case MAP_TYPE.ATTRIBUTE:
     default:
-      if (value === null || value === undefined || value === false) { // remove with undefined to comply with property
+      if (value === null || value === undefined || value === false) {
+        // remove with undefined to comply with property
```
wattachai-lseg: This comment describes `if (value === null || value === undefined || value === false) {`, the code proceeding it. With this formatting, the comment would ostensibly describing its following line instead.

wsuwt: Agree. I'm going move comment to just above the `if`

wattachai-lseg: A comment might be describing something above rather than below as a result of Prettier formatting. As there is no code behavior change, marking as resolved.

wattachai-lseg (dismissed):
Based on my sampling, LGTM. 😄

wattachai-lseg (approved):

PR #827, opened 2023-07-14
[v7] feat(docs): add version dropdown

TremayneChrist:
Adds version dropdown and improves graphic loading by moving resource to load locally


documents/src/menu.md
```diff
@@ -2,6 +2,9 @@
 type: menu
  -->
 
+* v7.x
+  * [v6.x](https://ui.refinitiv.com/v6)
```
wattachai-lseg:
```suggestion
  * [v6.x](http://ui.refinitiv.com/v6)
```
we should use `http` instead of `https` to avoid client side navigation of Pandora itself.

TremayneChrist: I don't see any client-side navigation happening. Checking the network log, it does a full page load. Client-side should only happen with relative URLs.

TremayneChrist: Ok, so when the URLs have the same origin, I see the issue. This needs to be fixed in our documentation tool. Will add _workaround_ for now.

dtanp-rft (approved):

PR #834, opened 2023-07-18
fix(notification,toggle): make user-select overridable

Nantawat-Poothong:
## Description
Make user-select can override

Solution:
make `user-select: none` CSS can override in the :host level

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-2043

## Type of change


Theeraphat-Sorasetsakul (approved):
LGTM

PR #847, opened 2023-07-20
fix(datetime-picker): allow user to clear invalid value in input

bualoy-napat:
## Description

To allow user to clear invalid input, we need to accept empty string for `value` and empty array for `values` to be assigned when current input is not valid.

Fixes # (issue) 
[ELF-1715](https://jira.refinitiv.com/browse/ELF-1715)

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
+++ b/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
@@ -89,6 +89,7 @@ describe('datetime-picker/Value', () => {
       await elementUpdated(el);
 
       expect(el.inputEl.value).to.be.equal('');
+      expect(el.error).to.be.equal(false);
     });
     it('It should be able to clear input values when user type invalid format for range mode', async () => {
       const el = await fixture('<ef-datetime-picker lang="en-gb" range opened></ef-datetime-picker>');
@@ -110,6 +111,7 @@ describe('datetime-picker/Value', () => {
 
       expect(el.inputEl.value).to.be.equal('');
       expect(el.inputToEl.value).to.be.equal('');
+      expect(el.error).to.be.equal(false);
     });
     it('It should not be possible to set from value after to', async () => {
       const el = await fixture(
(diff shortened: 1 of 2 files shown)


Nantawat-Poothong (changes requested):
I think the `error` state should be false after resetting the value

bualoy-napat:
> I think the `error` state should be false after resetting the value

Good point! I will add code to vaildate the input again in order to reset `error` state.

Nantawat-Poothong:
How about this one? just accept an empty array and go with the flow.
```typescript
  public set values(values: string[]) {
    const oldValues = this._values;
    if (String(oldValues) !== String(values) || !values.length) {
      this._values = values;
      this.valuesToSegments();
      this.requestUpdate('_values', oldValues); /* segments are populated in update */
    }
  }
```

bualoy-napat:
> How about this one? just accept an empty array and go with the flow.
> 
> ```ts
>   public set values(values: string[]) {
>     const oldValues = this._values;
>     if (String(oldValues) !== String(values) || !values.length) {
>       this._values = values;
>       this.valuesToSegments();
>       this.requestUpdate('_values', oldValues); /* segments are populated in update */
>     }
>   }
> ```

To my mind, I also thought to accept empty array in that condition. But it would be better to split off the code in order to make it easy to understand.

Nantawat-Poothong:
> To my mind, I also thought to accept empty array in that condition. But it would be better to split off the code in order to make it easy to understand.

I don't think so,  Due to setting `this.inputValues = [];`  I think splitting off the code like this will disrupt a normal flow and won't align with normal updating value flow. it may lead to another problem in further.

bualoy-napat:
> > To my mind, I also thought to accept empty array in that condition. But it would be better to split off the code in order to make it easy to understand.
> 
> I don't think so, Due to setting `this.inputValues = [];` I think splitting off the code like this will disrupt a normal flow and won't align with normal updating value flow. it may lead to another problem in further.

I get it, I will update PR ASAP.

Nantawat-Poothong (approved):

Nantawat-Poothong:
LGTM, wait to verified then you can merge to main branch.

PR #865, opened 2023-07-26
fix(number-field): add aria-valuetext to prevent VO announce percentage

bualoy-napat:
## Description

VO always announce percentage value in number-field when open in Safari. To make VO+Safari announced correct value, the document in MDN suggests to include `aria-valuetext`  with `aria-valuenow` to prevent it from announced value as percentage when `aria-valuemin` and `aria-valuemax` are not defined.
REF: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuenow

Fixes # (issue)
[ELF-2192](https://jira.refinitiv.com/browse/ELF-2192)

## Type of change

## Checklist


Sakchai-Refinitiv (approved):
Good 👍

PR #868, opened 2023-07-27
fix(number-field): add aria-valuetext to prevent VO announce percentage

bualoy-napat:
Sync change from https://github.com/Refinitiv/refinitiv-ui/pull/865


Sakchai-Refinitiv (approved):
Good 👍

PR #887, opened 2023-08-07
refactor(interactive-chart): revise internal chart type and type export

Nantawat-Poothong:
## Description
This changes will effect to
1. Typescript type import from ef-interactive-chart document
2. User who using some exported types from interactive-chart

This changes including 
1. Refactor custom types
2. Removed some public custom types  and using from lightweight-charts instead
   - Remove `Time` type user can use `BusinessDay` from lightweight-chart instead
   - Unpublish `RowLegend` type due to protected variable usage
   - Unpublish `LegendStyle` type
   - Unpublish `SeriesList` type user can import necessary types from lightweight-chart instead
   - Remove `SeriesData` type
   - Remove `SeriesStyleOptions` type
4. Improve casting type 

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-2215

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/interactive-chart/helpers/types.ts
+++ b/packages/elements/src/interactive-chart/helpers/types.ts
@@ -8,7 +8,8 @@ import type {
   SeriesType
 } from 'lightweight-charts';
 
-// https://stackoverflow.com/a/50375286
+// convert `A | B | C` into `A & B & C`. For more info, check https://stackoverflow.com/a/50375286
+``
 type UnionToIntersection<U> = (U extends unknown ? (k: U) => void : never) extends (k: infer I) => void
   ? I
   : never;
--- a/packages/elements/src/interactive-chart/helpers/types.ts
+++ b/packages/elements/src/interactive-chart/helpers/types.ts
@@ -1,40 +1,23 @@
 import type { HSLColor, RGBColor } from '@refinitiv-ui/utils/color.js';
 import type {
-  AreaSeriesPartialOptions,
-  AreaStyleOptions,
-  BarData,
-  BarSeriesPartialOptions,
-  BarStyleOptions,
-  CandlestickSeriesPartialOptions,
-  CandlestickStyleOptions,
   ChartOptions,
   DeepPartial,
-  HistogramData,
-  HistogramSeriesPartialOptions,
-  HistogramStyleOptions,
   ISeriesApi,
-  LineData,
-  LineSeriesPartialOptions,
-  LineStyleOptions,
-  SeriesPartialOptions,
+  SeriesDataItemTypeMap,
+  SeriesOptionsMap,
   SeriesType
 } from 'lightweight-charts';
 
-type SeriesOptions =
-  | AreaSeriesPartialOptions
-  | BarSeriesPartialOptions
-  | CandlestickSeriesPartialOptions
-  | HistogramSeriesPartialOptions
-  | LineSeriesPartialOptions;
-type SeriesStyleOptions = LineStyleOptions &
-  AreaStyleOptions &
-  BarStyleOptions &
-  CandlestickStyleOptions &
-  HistogramStyleOptions;
+// https://stackoverflow.com/a/50375286
+type UnionToIntersection<U> = (U extends unknown ? (k: U) => void : never) extends (k: infer I) => void
+  ? I
+  : never;
 
-type SeriesData = LineData[] | BarData[] | HistogramData[];
+type SeriesOptions = UnionToIntersection<SeriesOptionsMap[keyof SeriesOptionsMap]>;
+
 type SeriesList = ISeriesApi<SeriesType>;
-type SeriesDataItem = BarData | LineData | HistogramData;
+
+type SeriesDataItem = SeriesDataItemTypeMap[keyof SeriesDataItemTypeMap];
 
 type RowLegend = NodeListOf<Element> | HTMLElement | null;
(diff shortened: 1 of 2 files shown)


packages/elements/src/interactive-chart/helpers/types.ts
```diff
@@ -1,40 +1,23 @@
...
+type UnionToIntersection<U> = (U extends unknown ? (k: U) => void : never) extends (k: infer I) => void
+  ? I
+  : never;
+
+type SeriesOptions = UnionToIntersection<SeriesOptionsMap[keyof SeriesOptionsMap]>;
 
-type SeriesData = LineData[] | BarData[] | HistogramData[];
 type SeriesList = ISeriesApi<SeriesType>;
-type SeriesDataItem = BarData | LineData;
+
```
wattachai-lseg: let's add a comment. This's not that straightforward.
```suggestion
// a type union of each prop in `SeriesDataItemTypeMap` interface i.e. `BarData |  CandlestickData | ...`
```

packages/elements/src/interactive-chart/index.ts
```diff
@@ -506,8 +488,7 @@ export class InteractiveChart extends ResponsiveElement {
 
       for (let index = 0; index < this.internalConfig.series.length; index++) {
         // Get seriesOptions and type
-        const seriesOptions =
-          (this.internalConfig.series[index].seriesOptions as ChartSeriesOptions<SeriesStyleOptions>) || {};
+        const seriesOptions = (this.internalConfig.series[index].seriesOptions as SeriesOptions) || {};
```
wattachai-lseg: Partial would be better here.
```suggestion
        const seriesOptions: Partial<SeriesOptions> = this.internalConfig.series[index].seriesOptions || {};
```

packages/elements/src/interactive-chart/index.ts
```diff
@@ -790,28 +767,11 @@ export class InteractiveChart extends ResponsiveElement {
          * Don't need to be updated if chart has no data.
          */
         /* c8 ignore start */
-        let value;
+        let value: SeriesDataItem | string | number | undefined;
```
wattachai-lseg: no need for `number` here, right?
```suggestion
        let value: SeriesDataItem | string | undefined;
```

packages/elements/src/interactive-chart/index.ts
```diff
@@ -873,10 +830,17 @@ export class InteractiveChart extends ResponsiveElement {
...
+        this.createTextOHLC(rowLegendElem, value as OhlcData, priceColor, index);
       }
     } else {
-      this.createTextPrice(rowLegendElem, value as number | string, priceColor, index);
+      let price: string | number;
+      if ((value as SingleValueData).value !== undefined) {
+        price = (value as SingleValueData).value;
+      } else {
+        price = value as number;
+      }
```
wattachai-lseg: this should be simpler.
```suggestion
      const subValue = (value as SingleValueData).value;
      const price = subValue !== undefined ? subValue : value as number; 
```

packages/elements/src/interactive-chart/helpers/types.ts
```diff
@@ -1,40 +1,23 @@
...
-  | BarSeriesPartialOptions
-  | CandlestickSeriesPartialOptions
-  | HistogramSeriesPartialOptions
-  | LineSeriesPartialOptions;
-type SeriesStyleOptions = LineStyleOptions &
-  AreaStyleOptions &
-  BarStyleOptions &
-  CandlestickStyleOptions &
-  HistogramStyleOptions;
+// https://stackoverflow.com/a/50375286
```
wattachai-lseg:
```suggestion
// convert `A | B | C` into `A & B & C`. For more info, check https://stackoverflow.com/a/50375286
``

**Comment by wattachai-lseg** [comment]
There are a few locations with usages of `LegendStyle.horizontal` and `LegendStyle.vertical`. They should be replaced as well.

File: packages/elements/src/interactive-chart/index.ts (hunk 4 above)
Comment (Nantawat-Poothong) [reply packages/elements/src/interactive-chart/index.ts]: How about this one?
```suggestion
     const price = (value as SingleValueData).value || value as number; 
```

wattachai-lseg: if `value.value` is `0`, `price` would set to `value` instead of `value.value`. This changes code behavior so we better avoid this.

Nantawat-Poothong: Then how about this one. 😁

```suggestion
     const price = (value as SingleValueData).value ?? value; 
```

Nantawat-Poothong:
> There are a few locations with usages of `LegendStyle.horizontal` and `LegendStyle.vertical`. They should be replaced as well.

I'm not found any LegendStyle enum usage 🤔

packages/elements/src/interactive-chart/index.ts
wattachai-lseg: That should work! Don't forget to check the type of `price` tho.

wattachai-lseg:
> > There are a few locations with usages of `LegendStyle.horizontal` and `LegendStyle.vertical`. They should be replaced as well.
> 
> I'm not found any LegendStyle enum usage 🤔

I couldn't find them anymore. I probably was looking at the old version. My bad.

wattachai-lseg (approved):
LGTM

PR #891, opened 2023-08-08
fix(icon): change sprite cdn domain

Nantawat-Poothong:
## Description

Change sprite cdn domain from `https://cdn.refinitiv.com` to `https://cdn.refinitiv.net`

### FYI
.net => work both public + private network
.com => only work on public network

## Type of change


bualoy-napat (dismissed):
LGTM!

dtanp-rft (changes requested):
--cdn-sprite-prefix is in all theme folder. Need to update in all themes.

dtanp-rft (approved):

PR #892, opened 2023-08-08
docs: update cdn domain name

wattachai-lseg:
## Description

Change sprite cdn domain from `https://cdn.refinitiv.com` to `https://cdn.refinitiv.net`

### FYI
.net => work both public + private network
.com => only work on public network


bualoy-napat (approved):
LGTM! all cdn domain are correct!

PR #899, opened 2023-08-11
ci: make labeler action condition to be more specific on test files

wsuwt:
## Description

Making action labeler to be more specific on test files.


Nantawat-Poothong (approved):
LGTM

PR #900, opened 2023-08-11
ci: make labeler action condition to be more specific on test files

wsuwt:
## Description

Making action labeler to be more specific on test files.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -3,7 +3,7 @@ v7:
 
 documentation:
   - head-branch: ['^doc']
-  - changed-files: '**/*.md'
+  - changed-files: ['**/*.md']
 
 dependencies:
   - head-branch: ['^dependencies', '^deps']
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -3,7 +3,7 @@ v7:
 
 documentation:
   - head-branch: ['^doc']
-  - changed-files: '**/*.md'
+  - changed-files: ['**/*.md']
```
Nantawat-Poothong, line 6: Is it can ignore test files like v6?

wsuwt: snapshots in `v7` is under `__test__` and not a markdown files. So I think this condition should be enough.

Nantawat-Poothong: I see

Nantawat-Poothong (approved):
LGTM

.github/labeler.yml
Nantawat-Poothong: Snapshots on v7 are using `.snap.js` as well, it won't confusing for sure.

PR #903, opened 2023-08-11
[DEBUG]: labeler config

Nantawat-Poothong:
This PR is create for debugging the labeler.

Labeler adding incorrect label tags to the PR
for example this PR should have labels `v6` `ci` and `unit-testing`


Nantawat-Poothong:
@wsuwt seems like it still added `documentation` label for the unit-testing PR 😂

Nantawat-Poothong:
I found that to use `all` with `!some_path` it could exclude some specific path  as well.
Ref: https://github.com/actions/labeler#create-githublabeleryml

Screenshot 2566-08-18 at 14 41 56


**PS.** I curious that `github-actions bot` is run from this PR or branch master? if it run from branch master that mean it will be useless to debug here.

dtanp-rft:
What if we do 'documentation' in more specific pattern? Really it's about /documents/ and README.md right?
cc @Nantawat-Poothong  @wsuwt

PR #914, opened 2023-08-21
fix(slider): display and initial to value incorrectly

Nantawat-Poothong:
Sync from v7: https://github.com/Refinitiv/refinitiv-ui/pull/913


bualoy-napat (approved):
LGTM

PR #915, opened 2023-08-21
feat(tree-select): add max property

Sarin-Udompanish:
## Description
- Introduce `max` as a new prop/attr in Tree Select
- When users choose item that more than `max`, Done button will be disabled
- When users unselected item so the selected item is less than or equals to max, the Done button is activated again.
- Default value of `max` is null and if `max` is set with negative, revert to default.

## Type of change

## Checklist


Theeraphat-Sorasetsakul (approved):
LGTM.

PR #916, opened 2023-08-23
fix(slider): page freeze when from/to overflow max value

Nantawat-Poothong:
## Description

Slider range mode got loop hole when from/to overflow from max value

Root cause: 
loop hole happen when element trying to update from/to value under the max/min range but the new value still overflow from max/min range.

Solution: 
set compare values under max/min value before do any comparation 

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-2241

## Type of change

## Checklist

### DIFF
--- a/packages/elements/src/slider/__test__/slider.test.js
+++ b/packages/elements/src/slider/__test__/slider.test.js
@@ -347,7 +347,7 @@ describe('slider/Slider', function () {
     });
   });
   describe('Min-range', function () {
-    it('Set from/to overflow to nearly max', async function () {
+    it('Should reset from/to value to not exceed max value and ensure it keep min range correctly', async function () {
       el = await fixture(
         '<ef-slider range min="0" max="10" from="25" to="75"  min-range="1"></ef-slider></ef-slider>'
       );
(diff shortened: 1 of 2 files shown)


packages/elements/src/slider/__test__/slider.test.js
```diff
@@ -347,6 +347,16 @@ describe('slider/Slider', function () {
     });
   });
   describe('Min-range', function () {
+    it('Set from/to overflow to nearly max', async function () {
```
Sarin-Udompanish:
```suggestion
    it('Should reset from/to value to not exceed max value and ensure it keep min range correctly', async function () {
```

packages/elements/src/slider/index.ts
```diff
@@ -1148,21 +1147,15 @@ export class Slider extends ControlElement {
   private isValueInBoundary(value: number, valueFor: string): boolean {
     if (this.minNumber < this.maxNumber) {
```
Sarin-Udompanish:
```suggestion
   if (this.minNumber > this.maxNumber) {
      return false;
    }
```

packages/elements/src/slider/__test__/slider.test.js
Nantawat-Poothong: Updated

packages/elements/src/slider/index.ts
Nantawat-Poothong: Updated

Sarin-Udompanish (approved):

bualoy-napat:
LGTM

PR #935, opened 2023-09-11
chore: update eslint import extensions to ignore package imports

wsuwt:
## Description

Update eslint import extensions from always forcing extension to only force extension when import relatively and ignoring package imports

### DIFF
--- a/packages/configurations/typescript.eslint-config.js
+++ b/packages/configurations/typescript.eslint-config.js
@@ -1,10 +1,11 @@
 module.exports = {
   extends: ['./eslint-config.js'],
+  rules: {},
   overrides: [
     {
       files: ['*.ts', '*.tsx'],
       parser: '@typescript-eslint/parser',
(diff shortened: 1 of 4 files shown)


packages/configurations/typescript.eslint-config.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
       rules: {
         '@typescript-eslint/unbound-method': 0,
         '@typescript-eslint/no-namespace': 0,
-        'import/extensions': [2, 'always'],
+        'import/extensions': [2, 'ignorePackages'],
```
wattachai-lseg: let's remove this rule from both overrides and add it to `eslint-config.js` instead.
This should not break anything as the new rule config is more permissive.

wsuwt: Agreed and updated.

wattachai-lseg (approved):

PR #939, opened 2023-09-12
fix(calendar): view does not scroll when set value after view changed

bualoy-napat:
## Description

`ef-calendar` doesn't switch to view of value when we change view of calendar to next/previous month or set `view` and set `value` to `ef-calendar`.

Fixes # (issue)
[ELF-2088](https://jira.refinitiv.com/browse/ELF-2088)

## Checklist


bualoy-napat:
Calendar doesn't dictate the behavior to switch view when new value set because there might be some cases that view is not supposed to change. Thus, it provides a `view` API for developer to call after setting the new value if they want to change view.

PR #945, opened 2023-09-13
fix: create efx doesn't upload tsconfig template to npm

wsuwt:
## Description

Fixes `.npmignore` glob where it ignores `tsconfig.json.template`


packages/create-efx/.npmignore
wattachai-lseg, line 1: this is a good chance to use `file` field in package.json instead of `.npmignore`. Let's do that!

wsuwt: Good idea kub, much more readable. I'll update to `file`

wattachai-lseg (approved):

PR #946, opened 2023-09-13
fix: `tap` event should not be triggerred on right click

wsuwt:
## Description

`TapEvent` should not be dispatch when middle or right mouse is clicked

Fix issue
https://jira.refinitiv.com/browse/ELF-2283

## Checklist


packages/core/src/events/TapEvent.ts
```diff
@@ -204,12 +204,17 @@ const applyEvent = (target: Global): void => {
...
    * has already been triggered from a touch event.
    */
   target.addEventListener(
     'mousedown',
     (event) => {
+      // if mouse is not left click
+      if (event.button !== 0) {
+        return;
+      }
+
```
TremayneChrist: We shouldn't be too specific on whether it's the left or the right button, as this can change. You could use the term 'main' instead and declare it above to be reused across both parts. Then we don't need the comments either.

```suggestion
      if (event.button !== MAIN_BUTTON) {
        return;
      }

```

packages/core/__test__/events/TapEvent.test.js
```diff
@@ -71,6 +72,20 @@ const click = async (element1, element2) => {
...
 };
 
+const middleClick = async (element1, element2) => {
+  dispatchMouseEvent(element1, 'mousedown', 1);
+  await nextFrame();
+  dispatchMouseEvent(element2, 'mouseup', 1);
+  await nextFrame();
+};
+
+const rightClick = async (element1, element2) => {
```
TremayneChrist:
```suggestion
const secondaryClick = async (element1, element2) => {
```

packages/core/__test__/events/TapEvent.test.js
```diff
@@ -71,6 +72,20 @@ const click = async (element1, element2) => {
   await nextFrame();
 };
 
+const middleClick = async (element1, element2) => {
```
TremayneChrist:
```suggestion
const auxiliaryClick = async (element1, element2) => {
```

Sarin-Udompanish (approved):

bualoy-napat:
LGTM!

PR #971, opened 2023-09-25
fix(list, tree): stateless list and tree should have `aria-readonly`="true"

wsuwt:
## Description

Adding `aria-reaonly=true` when list and tree are in stateless mode i.e., can be interacted but not updating value. `aria-readonly` is compatible with `role=listbox`.

Reference, https://www.w3.org/TR/wai-aria-1.2/#aria-readonly 

Fixes [ELF-2264](https://jira.refinitiv.com/browse/ELF-2264)

## Checklist

### DIFF
--- a/packages/elements/src/list/elements/list.ts
+++ b/packages/elements/src/list/elements/list.ts
@@ -654,6 +654,10 @@ export class List<T extends DataItem = ItemData> extends ControlElement {
       this.renderTimestamp.clear(); // force render of all items
       this.setAttribute('aria-multiselectable', this.multiple ? 'true' : 'false');
     }
+
+    if (changeProperties.has('stateless')) {
+      this.setAttribute('aria-readonly', this.stateless ? 'true' : 'false');
+    }
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/elements/src/list/elements/list.ts
```diff
@@ -654,6 +654,10 @@ export class List<T extends DataItem = ItemData> extends ControlElement {
       this.renderTimestamp.clear(); // force render of all items
       this.setAttribute('aria-multiselectable', this.multiple ? 'true' : 'false');
     }
+
+    if (changeProperties.has('stateless')) {
+      this.setAttribute('aria-readonly', this.stateless ? 'true' : 'false');
+    }
```
Sakchai-Refinitiv: I prefer to remove the `aria-readonly="false"` to make the html tag clean and simple.

```suggestion
    
    if (changeProperties.has('stateless')) {
      if (this.stateless) {
        this.setAttribute('aria-readonly', 'true');
      } else {
        this.removeAttribute('aria-readonly');
      }
    }
```

packages/elements/src/tree/__test__/__snapshots__/tree.test.snap.js
Sakchai-Refinitiv, line 1: I think we should add one the A11Y test case to checking the `aria-readonly="true"` when enable stateless mode.
It would be great if you are the first to using this https://modern-web.dev/docs/test-runner/commands/#accessibility-snapshot

wsuwt: We plan to do this as part of another ticket, double checking every elements test.

Sakchai-Refinitiv (approved):
Great 👍

bualoy-napat:
LGTM!

PR #979, opened 2023-10-02
docs: update field docs

dtanp-rft:
Revise validity section in all fields controls and make the sample show best practice of accessibility.

### DIFF
--- a/documents/src/pages/elements/email-field.md
+++ b/documents/src/pages/elements/email-field.md
@@ -60,7 +60,7 @@ The field's value can be accessed directly using the `value` property.
 ```
 
 ```javascript
-const emailField = document.querySelector("ef-email-field");
+const emailField = document.getElementById("email);
 console.log(emailField.value); // "awesome@tmail.com"
 ```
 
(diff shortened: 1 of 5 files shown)


documents/src/pages/elements/email-field.md
```diff
@@ -90,32 +100,13 @@ emailField?.addEventListener("value-changed", (event) => {
 ```
 
 ## Input validation
+`ef-email-field` has validation logic similar to a [native input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text). When a user types the invalid value into the control, error style will be shown to notify the user.
```
Sarin-Udompanish:
```suggestion
`ef-email-field` has validation logic similar to a [native input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email). When a user types the invalid value into the control, error style will be shown to notify the user.
```

Sarin-Udompanish (dismissed):

wattachai-lseg (dismissed):

Sarin-Udompanish (approved):

PR #983, opened 2023-10-03
feat(number-field): support long press at spinner

Theeraphat-Sorasetsakul:
## Description

Support long press up/down spinner in Number Field

Fixes # (issue)
https://jira.refinitiv.com/browse/ELF-2265

## Native Number field
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#try_it

## Checklist


dtanp-rft:
I could be mishearing it but did we find that long key press in input works but not consistent cross browsers?
If it so, just wonder if it's feasible to change it to use the same function as long tap.

Theeraphat-Sorasetsakul:
> I could be mishearing it but did we find that long key press in input works but not consistent cross browsers? If it so, just wonder if it's feasible to change it to use the same function as long tap.

I wonder we can't do that. Long key press works because OS fires keydown event repeatedly. It's called Repeat rate on Windows. You can set up it on your OS keyboard setting. But mousedown/tap behavior fires only once times.

packages/elements/src/number-field/index.ts
```diff
@@ -504,6 +559,16 @@ export class NumberField extends FormFieldElement {
     }
   }
 
+  /**
+   * Reset error state on input
+   * @returns {void}
+   */
+  private resetError(): void {
```
Nantawat-Poothong: it's not using anywhere

Theeraphat-Sorasetsakul: Thank. This should not be there. Already removed.

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,43 @@ describe('number-field/Step', function () {
       await expectValues(el, [-1, -2, -3, -4, -5], UP, -1.5);
     });
   });
+
+  describe('Support long press', function () {
```
Nantawat-Poothong:
```suggestion
  describe('Long press spinner', function () {
```

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,43 @@ describe('number-field/Step', function () {
...
+    let el;
+    const dispatchLongTapEvent = async (el, wait) => {
```
Nantawat-Poothong: How about default `wait` parameter at 1000ms?
```suggestion
    const dispatchLongTapEvent = async (el, wait = 1000) => {
```

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,43 @@ describe('number-field/Step', function () {
...
+        el.dispatchEvent(
+          new Event('tapend', {
+            bubbles: true
+          })
+        )
+      );
+      await oneEvent(el, 'tapend');
+    };
+
+    beforeEach(async function () {
```
Nantawat-Poothong: judge from this test file context, I think we don't need `beforeEach()` here.

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,43 @@ describe('number-field/Step', function () {
...
+      el = await fixture('<ef-number-field></ef-number-field>');
+    });
+    it('By tap, Step Up', async function () {
+      await dispatchLongTapEvent(el.spinnerUpEl, 1000);
```
Nantawat-Poothong:
```suggestion
      await dispatchLongTapEvent(el.spinnerUpEl);
```

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,43 @@ describe('number-field/Step', function () {
...
+      await elementUpdated(el);
+      expect(Number(el.value)).to.greaterThan(0);
+    });
+    it('By tap, Step Down', async function () {
+      await dispatchLongTapEvent(el.spinnerDownEl, 1000);
```
Nantawat-Poothong:
```suggestion
      await dispatchLongTapEvent(el.spinnerDownEl);
```

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,43 @@ describe('number-field/Step', function () {
...
+      await oneEvent(el, 'tapend');
+    };
+
+    beforeEach(async function () {
+      el = await fixture('<ef-number-field></ef-number-field>');
+    });
+    it('By tap, Step Up', async function () {
+      await dispatchLongTapEvent(el.spinnerUpEl, 1000);
+      await elementUpdated(el);
+      expect(Number(el.value)).to.greaterThan(0);
```
Nantawat-Poothong: I think`greaterThan(0)` and `lessThan(0)` here aren't cover the long press spinner use case.
Just a single tap can pass this test case as well.

packages/elements/src/number-field/__test__/number-field.step.test.js
Theeraphat-Sorasetsakul: Removed.

packages/elements/src/number-field/__test__/number-field.step.test.js
Theeraphat-Sorasetsakul: Agree with you. So I update it to expect greaterthan and lessthan to 3 and -3 accordingly.

packages/elements/src/number-field/index.ts
```diff
@@ -169,6 +179,26 @@ export class NumberField extends FormFieldElement {
   @query('[part=spinner-down]')
   private spinnerDownEl?: HTMLInputElement;
 
+  /**
+   * An object's returned from setTimeout to use with repeat delay.
+   */
+  private timerRepeatDelay: NodeJS.Timeout | undefined = undefined;
```
Nantawat-Poothong:
```suggestion
  private timerRepeatDelay: NodeJS.Timeout | undefined;
```

packages/elements/src/number-field/index.ts
```diff
@@ -169,6 +179,26 @@ export class NumberField extends FormFieldElement {
...
+
+  /**
+   * An object's returned from setInterval to use with repeat rate.
+   */
+  private timerRepeatRate: NodeJS.Timeout | undefined = undefined;
```
Nantawat-Poothong:
```suggestion
  private timerRepeatRate: NodeJS.Timeout | undefined;
```

packages/elements/src/number-field/index.ts
```diff
@@ -169,6 +179,26 @@ export class NumberField extends FormFieldElement {
...
+
+  /**
+   * Called after the component is first rendered
+   * @param changedProperties Properties which have changed
+   * @returns {void}
+   */
+  protected override firstUpdated(changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    document.addEventListener('tapend', this.resetTimer);
```
Nantawat-Poothong: I feel uncomfortable with `document.addEventListener` how about `this.addEventListener` does it have any concerns?

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,41 @@ describe('number-field/Step', function () {
...
+          })
+        )
+      );
+      await oneEvent(el, 'tapend');
+    };
+
+    it('By tap, Step Up', async function () {
+      const el = await fixture('<ef-number-field></ef-number-field>');
+      await dispatchLongTapEvent(el.spinnerUpEl);
+      await elementUpdated(el);
```
Nantawat-Poothong: Does element really need wait to update?

packages/elements/src/number-field/index.ts
Theeraphat-Sorasetsakul: Using `this.addEventListener` does `tapend` isn't fired. This causes number-field does infinite count up when long press and drag out of number-field.

packages/elements/src/number-field/__test__/number-field.step.test.js
Theeraphat-Sorasetsakul: It doesn't need anymore. Removed.

packages/elements/src/number-field/index.ts
Nantawat-Poothong: Oh I see, it necessary to detect tapend from anywhere inside DOM document.
Does will it cause any concerns in further? for example
1. when number-field in side shadow DOM
2. when have to many number field elements

Theeraphat-Sorasetsakul: 1. Tested on color-dialog and time-picker. It works well if they had spinner.
2. Added condition to not call clearTimer at every tapend.

Nantawat-Poothong (dismissed):
LGTM

Nantawat-Poothong (dismissed):

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,39 @@ describe('number-field/Step', function () {
...
+        el.dispatchEvent(
+          new Event('tapend', {
+            bubbles: true
+          })
+        )
+      );
+      await oneEvent(el, 'tapend');
+    };
+
+    it('By tap, Step Up', async function () {
```
wsuwt:
```suggestion
    it('Should increases value gradually on long press', async function () {
```

packages/elements/src/number-field/__test__/number-field.step.test.js
```diff
@@ -220,4 +220,39 @@ describe('number-field/Step', function () {
...
+      const el = await fixture('<ef-number-field></ef-number-field>');
+      await dispatchLongTapEvent(el.spinnerUpEl);
+      expect(Number(el.value)).to.greaterThan(3);
```
wsuwt: What does `3` number means ?

Nantawat-Poothong: number 3 is a magic number to assume that long press is working

wsuwt: In that case, wouldn't `greater than 1` already proves the functionality.

Nantawat-Poothong: Maaaaake sense, agree.

bualoy-napat: On this point, I think he want to check the value that should increase to `3` if user press spinner for 1 second.

Nantawat-Poothong: I think the exact value is unpredictable and depend on test device performance. but at least pressing spinner for one second, the value should literally greater than 1.

wsuwt: If so, we have could have 2 test cases. One just to test long press functionality. Another is to test incremental speed for long press. I feel the later test case is unnecessary, but could be added just in case.

bualoy-napat:
image
I think this hover style is not correct. 
You can follow this step to replicate the issue
1. Long press on up spinner
2. Move cursor to down spinner

Theeraphat-Sorasetsakul:
> image I think this hover style is not correct. You can follow this step to replicate the issue
> 
> 1. Long press on up spinner
> 2. Move cursor to down spinner

I think I disagree because this is normal behavior of hover css.

Hover CSS class do when users point at the element, but does not necessarily activate it.
ref: https://developer.mozilla.org/en-US/docs/Web/CSS/:hover

packages/elements/src/number-field/__test__/number-field.step.test.js
Theeraphat-Sorasetsakul: Agreed that '1' is enough to be a value for test. So, I updated.

packages/elements/src/number-field/index.ts
```diff
@@ -109,6 +109,16 @@ export class NumberField extends FormFieldElement {
     `;
   }
 
+  /**
+   * Time period (ms) between starting press and starting repeat.
```
wsuwt:
```suggestion
   * Time period (ms) before press repetition starts
```

packages/elements/src/number-field/index.ts
```diff
@@ -109,6 +109,16 @@ export class NumberField extends FormFieldElement {
...
+   */
+  private repeatDelay = 300;
+
+  /**
+   * Time period (ms) which it repeats.
```
wsuwt:
```suggestion
   * Time period (ms) between each repeat
```

packages/elements/src/number-field/index.ts
```diff
@@ -169,6 +179,26 @@ export class NumberField extends FormFieldElement {
...
+  private timerRepeatRate: NodeJS.Timeout | undefined;
+
+  /**
+   * Called after the component is first rendered
+   * @param changedProperties Properties which have changed
+   * @returns {void}
+   */
+  protected override firstUpdated(changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    document.addEventListener('tapend', this.clearTimer);
```
wsuwt, line 200:
```suggestion
    // To remove press repetition when tap event ends outside of the pressed button
    document.addEventListener('tapend', this.clearTimer);
```

wsuwt (changes requested):

packages/elements/src/number-field/index.ts
```diff
@@ -169,6 +179,27 @@ export class NumberField extends FormFieldElement {
   @query('[part=spinner-down]')
   private spinnerDownEl?: HTMLInputElement;
 
+  /**
+   * An object's returned from setTimeout to use with repeat delay.
+   */
+  private timerRepeatDelay: NodeJS.Timeout | undefined;
```
wsuwt:
```suggestion
  private repeatDelayTimer: NodeJS.Timeout | undefined;
```

packages/elements/src/number-field/index.ts
```diff
@@ -169,6 +179,27 @@ export class NumberField extends FormFieldElement {
...
+
+  /**
+   * An object's returned from setInterval to use with repeat rate.
+   */
+  private timerRepeatRate: NodeJS.Timeout | undefined;
```
wsuwt:
```suggestion
  private repeatRateTimer: NodeJS.Timeout | undefined;
```

wsuwt (dismissed):

wsuwt:
Looks good kub.

packages/elements/src/number-field/index.ts
```diff
@@ -330,13 +361,28 @@ export class NumberField extends FormFieldElement {
...
+    // Support long tap at a spinner
+    this.repeatDelayTimer = setTimeout(() => {
+      this.repeatRateTimer = setInterval(() => {
+        this.onApplyStep(direction);
+      }, this.repeatRate);
+    }, this.repeatDelay);
   }
 
+  /**
+   * Clear timerRepeatDelay and timerRepeatRate if exist
```
wsuwt:
```suggestion
   * Clear repeatDelayTimer and repeatRateTimer if exist
```

wsuwt (approved):

PR #989, opened 2023-10-09
fix: `ITimeScaleApi` type errors in `lightweight-charts` to 4.1.0

wsuwt:
## Description

Fixes build error caused by  `lightweight-charts@4.1.0`

## Checklist

### DIFF
--- a/packages/elements/src/interactive-chart/index.ts
+++ b/packages/elements/src/interactive-chart/index.ts
@@ -1,4 +1,4 @@
-import { ColorType, Time, createChart as chart } from 'lightweight-charts';
+import { ColorType, createChart as chart } from 'lightweight-charts';
 
 import {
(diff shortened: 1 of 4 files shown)


packages/elements/src/interactive-chart/index.ts
```diff
@@ -1,4 +1,4 @@
-import { ColorType, createChart as chart } from 'lightweight-charts';
+import { ColorType, Time, createChart as chart } from 'lightweight-charts';
```
Theeraphat-Sorasetsakul:
```suggestion
import { ColorType, createChart as chart } from 'lightweight-charts';
```
I saw `import type` at line number 33. We should do at that.

Theeraphat-Sorasetsakul (approved):

PR #998, opened 2023-10-18
fix(list): `selectItem()` should always require a param

wsuwt:
## Description

`selectItem()` should be requiring an item param.

## Checklist


Nantawat-Poothong (dismissed):
LGTM

Nantawat-Poothong:
I have a question, isn't it will break backward compatible?

wsuwt:
> I have a question, isn't it will break backward compatible?

Yes, it's possible. Though, I can't imagine why would someone use `selectItem()` without item as a parameter. This would always return `false` which is the reason I think this is meant to be bug fix rather than breaking features.

Nantawat-Poothong:
> > I have a question, isn't it will break backward compatible?
> 
> Yes, it's possible. Though, I can't imagine why would someone use `selectItem()` without item as a parameter. This would always return `false` which is the reason I think this is meant to be bug fix rather than breaking features.

I see, I digged deep down to see the reason of this optional parameter item, seems like this 3 years code was invented by inattention code style not a feature.

Nantawat-Poothong (approved):
LGTM

PR #1025, opened 2023-11-16
fix: voiceover and safari does not annouce selected value

wsuwt:
## Description

Move `role=combobox` to internal `div` that has the same level as `role=listbox`. This way Safari and VoiceOver will recognize the relationship between the two.  

- Added basic accessibility test cases
- Update documentation content and live examples
- Update demo page to be accessible

Fixes [ELF-2131](https://jira.refinitiv.com/browse/ELF-2131)

## Checklist


packages/elements/src/select/index.ts
```diff
@@ -189,7 +178,7 @@ export class Select extends ControlElement implements MultiValue {
    * Placeholder to display when no value is set
    */
   @property({ type: String })
-  public placeholder = '';
+  public override placeholder = '';
```
Sakchai-Refinitiv: Why it needs to be overrided with the same value?

packages/elements/src/select/index.ts
```diff
@@ -201,13 +190,13 @@ export class Select extends ControlElement implements MultiValue {
...
   @property({ type: Boolean, reflect: true })
-  public error = false;
+  public override error = false;
 
   /**
    * Set warning state
    */
   @property({ type: Boolean, reflect: true })
-  public warning = false;
+  public override warning = false;
```
Sakchai-Refinitiv: Why it needs to be overrided with the same value?

packages/elements/src/select/index.ts
```diff
@@ -201,13 +190,13 @@ export class Select extends ControlElement implements MultiValue {
    * Set error state
    */
   @property({ type: Boolean, reflect: true })
-  public error = false;
+  public override error = false;
```
Sakchai-Refinitiv: Why it needs to be overrided with the same value?

Sakchai-Refinitiv:
I can't reproduce this bug on the Safari 16.5.1 on the BrowserStack we should recheck it on real device.

Screenshot 2566-11-27 at 09 56 55

wsuwt:
> I can't reproduce this bug on the Safari 16.5.1 on the BrowserStack we should recheck it on real device.
> 
> Screenshot 2566-11-27 at 09 56 55

It should be reproducible. The correct way to reproduce is focusing on `select` with popup closed and it will not annouce selected value.

documents/src/pages/elements/select.md
```diff
@@ -178,70 +167,35 @@ Groups are also defined using `ef-item`. The only difference is that we add a `t
...
-  <ef-item value="3">Water</ef-item>
-  <ef-item type="header">Ice Cream</ef-item>
-  <ef-item value="4">Vanilla</ef-item>
-  <ef-item value="5">Chocolate</ef-item>
-  <ef-item value="6">Strawberry</ef-item>
-  <ef-item value="7">Raspberry</ef-item>
-</ef-select>
-```
-
-## Adding a placeholder
```
Sakchai-Refinitiv, line 216: Why we have to remove this use case from the docs?

documents/src/pages/elements/select.md
```diff
@@ -15,49 +15,35 @@ language_tabs: [javascript, typescript]
...
-    <ef-item type="header">Ice Cream</ef-item>
-    <ef-item value="4">Vanilla</ef-item>
-    <ef-item value="5">Chocolate</ef-item>
-    <ef-item value="6">Honey &amp; Walnut</ef-item>
-    <ef-item value="7">Raspberry</ef-item>
-  </ef-select>
-  <ef-select disabled>
-    <ef-item selected>Lemonade</ef-item>
+  <label for="desserts">Desserts </label>
+  <ef-select id="desserts" placeholder="Pick your preferred desserts" opened>
```
Sakchai-Refinitiv: The placeholder is always hidden because of the selected option. I think the placeholder should be removed.

packages/elements/src/select/__demo__/index.html
```diff
@@ -29,32 +29,49 @@
...
-        <ef-item role="option" value="10">Pistachio</ef-item>
-        <ef-item role="option" value="11">Salted Caramel</ef-item>
-        <ef-item role="option" value="12">Mint Choc Chip</ef-item>
-        <ef-item role="option" value="13">Hazelnut</ef-item>
-        <ef-item role="option" value="14">Strawberry</ef-item>
-        <ef-item role="option" value="15">Raspberry</ef-item>
-      </ef-select>
+      <section>
+        <label for="desserts">Desserts:</label>
+        <ef-select id="desserts" placeholder="Pick your preferred desserts...">
```
Sakchai-Refinitiv: The placeholder is always hidden because of the selected option. I think the placeholder should be removed.

packages/elements/src/select/index.ts
```diff
@@ -1074,11 +1035,25 @@ export class Select extends ControlElement implements MultiValue {
...
-    return html` <div id="box">
+    return html`<div
+        id="box"
+        tabindex="0"
+        role="combobox"
+        placeholder=${this.placeholder || nothing}
+        aria-haspopup="listbox"
+        aria-controls="menu"
+        aria-invalid=${this.error ? 'true' : 'false'}
+        aria-expanded=${this.opened ? 'true' : 'false'}
```
Sakchai-Refinitiv, line 1046: We can remove short if by do this.
```suggestion
        aria-invalid=${String(this.error)}
        aria-expanded=${String(this.opened)}
```

wsuwt: Anything can be input as `error` or `opened` so I think it is better to be explicit that these attributes only accept `'true'` or `'false'`

packages/elements/src/select/__demo__/index.html
wsuwt: Remove this one's placeholder and updated `currency` demo to show placeholder kub.

documents/src/pages/elements/select.md
wsuwt: Updated.

packages/elements/src/select/index.ts
Sakchai-Refinitiv: It's OK.

Sakchai-Refinitiv (approved):
Nice 👍

PR #1040, opened 2023-11-29
feature(slider): tap on label should select value

Sarin-Udompanish:
## Description
* Show pointer cursor when hover on label so users know that it's clickable.
* Click on label should set slider's value with the value of marker.
* If value of marker is not valid e.g. if slider has step = 5 and label is set to show at 8, clicking on the label should set slider with the nearest value – in this case, it should set 10.
* If sliders in range mode with min-range set, slider's value should be set to the nearest value satisfying all criteria.

## Checklist

### DIFF
--- a/packages/elements/src/slider/elements/slider.ts
+++ b/packages/elements/src/slider/elements/slider.ts
@@ -439,6 +439,30 @@ export class Slider extends ControlElement {
   }
 
   /**
+   * Finds the closest ancestor ef-slider-marker in the DOM tree.
+   * @param startingElement The HTML element to start searching from.
+   * @returns The found marker, or null if not found.
+   */
+  private findClosestMarker(startingElement: EventTarget | Node | null): SliderMarker | null {
+    if (!startingElement) {
+      return null;
+    }
+
+    let currentNode: Node | null = startingElement as Node | null;
+    const markers = this.getMarkerElements();
+
+    while (currentNode && currentNode !== this) {
+      if (currentNode instanceof SliderMarker && markers.includes(currentNode)) {
+        return currentNode;
+      }
+
+      currentNode = currentNode.parentNode;
+    }
+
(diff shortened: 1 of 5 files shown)


packages/elements/src/slider/elements/slider.ts
```diff
@@ -438,6 +438,30 @@ export class Slider extends ControlElement {
...
+   * @param startingElement The HTML element to start searching from.
+   * @returns The found marker, or null if not found.
+   */
+  private findClosestMarker(startingElement: EventTarget | Node | null): SliderMarker | null {
+    if (!startingElement) {
+      return null;
+    }
+
+    let currentNode: Node | null = startingElement as Node | null;
+    const markers = this.getMarkerElements();
```
wattachai-lseg, line 452: `startingElement` can't be `null` due to the falsy check above.
```suggestion
    let currentNode: Node | null = startingElement as Node;
```

packages/elements/src/slider/elements/slider.ts
```diff
@@ -438,6 +438,30 @@ export class Slider extends ControlElement {
     return markers;
   }
 
+  /**
+   * Finds the closest ancestor ef-slider-marker in the DOM tree.
+   * @param startingElement The HTML element to start searching from.
+   * @returns The found marker, or null if not found.
+   */
+  private findClosestMarker(startingElement: EventTarget | Node | null): SliderMarker | null {
```
wattachai-lseg: No need to accept `Node` here, right?
```suggestion
  private findClosestMarker(startingElement: EventTarget | null): SliderMarker | null {
```

wattachai-lseg (changes requested):
Let's add some new test cases checking clicks on the label marker.

wattachai-lseg (approved):
LGTM

PR #1042, opened 2023-11-30
fix: voiceover and safari does not annouce selected value

wsuwt:
## Description

Move `role=combobox` to internal `div` that has the same level as `role=listbox`. This way Safari and VoiceOver will recognize the relationship between the two.  

- Added basic accessibility test cases
- Update documentation content and live examples
- Update demo page to be accessible

Fixes [ELF-2131](https://jira.refinitiv.com/browse/ELF-2131)

## Checklist

### DIFF
--- a/packages/elements/src/select/index.ts
+++ b/packages/elements/src/select/index.ts
@@ -40,6 +40,13 @@ const observerOptions = {
   attributeFilter: ['label', 'value', 'selected', 'disabled', 'readonly']
 };
 
+// the same event listener options should be used with both addEventListener() & removeEventListener()
+// https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal
+const eventListenerOptions = {
+  /* need this for IE11, otherwise the event is not removed */ capture: true,
+  passive: true
+};
+
 const LABEL_SEPARATOR = ', '; // TODO: for multiselect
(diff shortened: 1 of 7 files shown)


packages/elements/src/select/index.ts
```diff
@@ -1043,49 +994,73 @@ export class Select extends ControlElement implements MultiValue {
...
+  /**
+   * Area that allows select to be clickable and toggle between open and close
+   */
+  private get triggerTemplate(): TemplateResult {
+    return html` <div id="trigger" @tapstart="${this.toggleOpened}"></div>`;
   }
 
   /**
-   * A `TemplateResult` that will be used
-   * to render the updated internal template.
```
Sakchai-Refinitiv: It should not be removed

packages/elements/src/select/index.ts
Sakchai-Refinitiv, line 1: Please refactor the `eventOptions` for the scroll event to be the same v7.

packages/elements/src/select/index.ts
wsuwt: Updated.

packages/elements/src/select/index.ts
Sakchai-Refinitiv: Updated

Sakchai-Refinitiv (approved):
Nice 👍

PR #1046, opened 2023-12-06
fix(combo-box): cannot set value in free text when data is empty

bualoy-napat:
## Description

In free text mode, a developer cannot set any value to the combo-box when data is empty. Therefore, I propose the solution that when a developer sets a value, it will save the value as the free text value.

Fixes # (issue)
ELF-2349

## Checklist

### DIFF
--- a/packages/elements/src/combo-box/__test__/combo-box.value.test.js
+++ b/packages/elements/src/combo-box/__test__/combo-box.value.test.js
@@ -132,7 +132,7 @@ describe('combo-box/Value', function () {
       el.value = '';
       expect(el.value).to.equal('', 'Value must be empty string when reset value on free text mode');
     });
-    it('Cannot set any value without free text mode when data is empty', async function () {
+    it('Should not update value when data is empty', async function () {
       const el = await fixture('<ef-combo-box lang="en"></ef-combo-box>');
       el.value = 'AL';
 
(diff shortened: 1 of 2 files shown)


packages/elements/src/combo-box/__test__/combo-box.value.test.js
```diff
@@ -132,5 +132,19 @@ describe('combo-box/Value', function () {
       el.value = '';
       expect(el.value).to.equal('', 'Value must be empty string when reset value on free text mode');
     });
+    it('Cannot set any value without free text mode when data is empty', async function () {
```
wattachai-lseg: It seems this test suite breaks our "it should" convention. Let's gradually fix this be applying the convention to new test cases.
```suggestion
    it('Should not update value when data is empty', async function () {
```

packages/elements/src/combo-box/__test__/combo-box.value.test.js
```diff
@@ -132,5 +132,19 @@ describe('combo-box/Value', function () {
...
+      const el = await fixture('<ef-combo-box lang="en"></ef-combo-box>');
+      el.value = 'AL';
+
+      await elementUpdated(el);
+      expect(el.value).to.equal('');
+    });
+    it('Set any value with free text mode when data is empty', async function () {
```
wattachai-lseg: It seems this test suite breaks our "it should" convention. Let's gradually fix this be applying the convention to new test cases.
```suggestion
    it('Should update value in free text mode when data is empty', async function () {
```

wattachai-lseg:
LGTM. Let's proceed to verification.

wattachai-lseg (approved):

PR #1050, opened 2023-12-07
fix(i18n): resolve import order issue on demo page

wattachai-lseg:
## Description

resolve import order issue caused by prettier

## Checklist

### DIFF
--- a/packages/i18n/__demo__/index.html
+++ b/packages/i18n/__demo__/index.html
@@ -12,9 +12,10 @@
     }
   </style>
   <body>
+    <!-- prettier-ignore -->
     <script type="module">
-      import 'https://cdn.jsdelivr.net/npm/benchmark@2.1.4/benchmark.min.js';
       import 'https://cdn.jsdelivr.net/npm/lodash@4.17.21/lodash.min.js';
+      import 'https://cdn.jsdelivr.net/npm/benchmark@2.1.4/benchmark.min.js';
 
       import '@refinitiv-ui/demo-block';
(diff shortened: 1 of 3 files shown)


packages/i18n/__demo__/index.html
```diff
@@ -12,9 +12,10 @@
     }
   </style>
   <body>
+    <!-- prettier-ignore -->
     <script type="module">
-      import 'https://cdn.jsdelivr.net/npm/benchmark@2.1.4/benchmark.min.js';
       import 'https://cdn.jsdelivr.net/npm/lodash@4.17.21/lodash.min.js';
+      import 'https://cdn.jsdelivr.net/npm/benchmark@2.1.4/benchmark.min.js';
```
wsuwt: So does this means that to use benchmark.js as cdn, import order matters ? In that case, would it be better to just include it as a part of dev dependencies because we feature this in demo page anyways ?

wattachai-lseg: Yes, import order matters here. Benchmark is seldomly used that's why it has been added this way instead of relying on dev dependencies.

wsuwt: My opinion is if it's in demo page, it is being used every now and then. This is the only demo that import packages using cdn.

wattachai-lseg: Changed to dev dependencies. No noticeable build time impact. :)

wsuwt (approved):

PR #1057, opened 2023-12-22
docs: revise Angular tutorial contents and example code

wsuwt:
## Description

Revise Angular tutorial contents and example code.

- codesandbox is now updated to EFv7 with Angular 16
- removed installation and how to create Angular app from tutorial to simplify and focus only on integration part
- update ControlValueAccessor example to be using `ef-slider` instead

Make React tutorial more consistent and correct version.


bualoy-napat:
I think we should remove installation part of Vue same as we did in Angular.

kosol-komenek-rft (approved):
Look good for me

bualoy-napat:
LGTM.

PR #1062, opened 2024-01-04
chore: bump deps versions in package.json to match with installed package

dtanp-rft:


Theeraphat-Sorasetsakul (approved):
Tested and build passed.

PR #1066, opened 2024-01-09
fix(autosuggest): incorrect mark style

bualoy-napat:
## Description

`mark` color on suggestion list is not correct when attach the `ef-autosuggest` inside the shadow DOM. The root cause of this issue is that the `mark` element is not within the scope of the `ef-autosuggest` theme. It is slotted inside the `ef-item`. Therefore, moving the `mark` style to the scope of `ef-item` will fix the issue.

Fixes # (issue)
ELF-2366

## Checklist


kosol-komenek-rft (dismissed):
Look good for me.

packages/elements/src/autosuggest/__demo__/index.html
```diff
@@ -665,66 +667,21 @@
...
-                },
-                {
-                  suggest: 'Cornelius Martin'
-                }
-              ]
-            }
-          };
-        })();
+        const wrapper = document.getElementById('search-suggestion');
+        let searchField = document.createElement('ef-search-field');
```
wattachai-lseg: let's replace `let` with `const` if there is no reassignment.

wattachai-lseg (changes requested):
2 more things:
* Could you replace relative path import (`import <something> from './../../../lib/autosuggest/<path>`) with module import (`import <something> from @refinitiv-ui/elements/autosuggest/<path>`) instead?
* I couldn't find a description for "inside shadow root" demo we discussed earlier.

packages/elements/src/autosuggest/__demo__/index.html
bualoy-napat: Updated!

packages/elements/src/autosuggest/__demo__/index.html
```diff
@@ -665,66 +667,25 @@
...
+        const searchField = document.createElement('ef-search-field');
+        const autoSuggest = document.createElement('ef-autosuggest');
+        autoSuggest.attach = searchField;
+        autoSuggest.addEventListener('suggestions-fetch-requested', (event) => {
+          const query = event.detail.query;
+          window.filterData(query).then(({ data }) => {
+            autoSuggest.suggestions = data;
+          });
+        });
+        wrapper.appendChild(searchField);
```
wattachai-lseg: Elements should be appended to wrapper element's shadow root not the element itself!

wattachai-lseg (approved):

PR #1083, opened 2024-01-17
refactor(combo-box): move placeholder property to document comment

Theeraphat-Sorasetsakul:
## Description

Found that ComboBox placeholder property didn't align with other FormFieldElement. The property declared inside the class but other components, ex select, described at JSDoc document comment.

### AC
- placeholder should display on document at property API reference section.
- placeholder should work by the same.

### DIFF
--- a/packages/elements/src/combo-box/index.ts
+++ b/packages/elements/src/combo-box/index.ts
@@ -757,7 +757,7 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
    * @returns {void}
    */
   protected restrictPopupWidth(): void {
-    /* istanbul ignore next */
+    /* c8 ignore start */
     if (this.offsetWidth === 0) {
       // this code might happen only when opened has been set during initialisation
       // or when display is set to none
@@ -770,7 +770,7 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
       });
       return;
     }
-
+    /* c8 ignore stop */
     const maxWidth = parseFloat(this.getComputedVariable('--list-max-width', 'none'));
     let minWidth = this.offsetWidth;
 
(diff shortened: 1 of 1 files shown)


packages/elements/src/combo-box/index.ts
```diff
@@ -760,7 +757,7 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
    * @returns {void}
    */
   protected restrictPopupWidth(): void {
-    /* c8 ignore start */
+    /* istanbul ignore next */
```
wattachai-lseg: It should be `c8` not `istantbul` for v7.

packages/elements/src/combo-box/index.ts
```diff
@@ -1183,12 +1182,12 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
   protected reTargetEvent(event: KeyboardEvent, target: HTMLElement): CustomKeyboardEvent {
     const path = event.composedPath();
 
-    /* c8 ignore start */
+    /* istanbul ignore next */
```
wattachai-lseg: It should be `c8` not `istantbul` for v7.

packages/elements/src/combo-box/index.ts
```diff
@@ -1094,9 +1091,11 @@ export class ComboBox<T extends DataItem = ItemData> extends FormFieldElement {
       case 'Enter':
         this.enter(event);
         break;
+      case 'Up':
```
wattachai-lseg: Why are you adding legacy key value back in???
https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#navigation_keys

wattachai-lseg (approved):

PR #1085, opened 2024-01-17
docs: replace using individual element templates with one template

dtanp-rft:
All individual element template to inject importing component/theme are obsoleted once we moved to bundle everything to one file.


kosol-komenek-rft (approved):
Look good for me krub. 👍

PR #1108, opened 2024-02-23
docs(heatmap): fix label-width description

Theeraphat-Sorasetsakul:
## Description

Fix wrong description

### DIFF
--- a/packages/elements/src/heatmap/index.ts
+++ b/packages/elements/src/heatmap/index.ts
@@ -118,7 +118,7 @@ export class Heatmap extends ResponsiveElement {
 
   /**
    * Set maximum text length to be shown on cells. Unit in pixel.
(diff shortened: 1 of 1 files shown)


packages/elements/src/heatmap/index.ts
```diff
@@ -118,7 +118,7 @@ export class Heatmap extends ResponsiveElement {
 
   /**
    * Set maximum text length to be shown on cells. Unit in pixel.
-   * e.g. label-width = 30; cell's label is hidden when text length less than 30px.
+   * e.g. label-width = 30; cell's label is hidden when text length larger than 30px.
```
wattachai-lseg:
```suggestion
   * e.g. label-width = 30; cell's label is hidden when text length is larger than 30px.
```

wattachai-lseg (approved):

PR #1117, opened 2024-03-07
fix(datetime-picker): validate values for weekendsonly and weekdaysonly

bualoy-napat:
## Description

Previously, datetime-picker was not validating values that within condition of `weekends-only` and `weekdays-only`. To fix this issue, we just simply add validation of `weekdays-only` and `weekends-only` to `validateInput` method. This method will be calling every time that input has change.

Expected behavior of this change:
1. `weekdaysOnly` or `weekendsOnly` behavior should be same to min/max property.
2. error property should be true if user select disabled dates through typing or setting value via API.
3. UI still show selected cell but control also show error state (red border)
4. Incorrect value is still set to `value`. It's now assume that this will be handle by app e.g. to revert the value back or to set value to empty, etc.

Fixes # (issue)
DME-9462

## Checklist

### DIFF
--- a/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
+++ b/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
@@ -303,7 +303,7 @@ describe('datetime-picker/Value', function () {
       await elementUpdated(el);
       expect(el.error).to.be.equal(true);
     });
-    it('It must error when weekdays-only is set and values are not in weekday period with range mode', async function () {
+    it('It should error when weekdays-only is true and values are not in weekday period with range mode', async function () {
       const el = await fixture('<ef-datetime-picker weekdays-only lang="en-gb" opened></ef-datetime-picker>');
 
       el.values = ['2024-03-04', '2024-03-08'];
--- a/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
+++ b/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
@@ -261,7 +261,7 @@ describe('datetime-picker/Value', function () {
       await elementUpdated(el);
       expect(el.error).to.be.equal(false);
     });
-    it('It must error when weekdays-only is set and value is not in weekday period', async function () {
+    it('It should error when weekdays-only is true and value is not in weekday period', async function () {
       const el = await fixture('<ef-datetime-picker weekdays-only lang="en-gb" opened></ef-datetime-picker>');
 
       el.value = '2024-03-01';
--- a/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
+++ b/packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
@@ -246,5 +246,89 @@ describe('datetime-picker/Value', function () {
       typeText(timePicker, '16:17:18');
       expect(el.value).to.equal('2020-04-21T16:17:18');
     });
+    it('It must error when weekends-only is set and value is not in weekend period', async function () {
+      const el = await fixture('<ef-datetime-picker weekends-only lang="en-gb" opened></ef-datetime-picker>');
+
+      el.value = '2024-03-02';
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(false);
+
+      el.value = '2024-03-01';
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(true);
+
+      el.value = '2024-03-02';
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(false);
+    });
+    it('It must error when weekdays-only is set and value is not in weekday period', async function () {
+      const el = await fixture('<ef-datetime-picker weekdays-only lang="en-gb" opened></ef-datetime-picker>');
+
+      el.value = '2024-03-01';
+      await elementUpdated(el);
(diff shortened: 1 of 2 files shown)


packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
```diff
@@ -246,5 +246,89 @@ describe('datetime-picker/Value', function () {
       typeText(timePicker, '16:17:18');
       expect(el.value).to.equal('2020-04-21T16:17:18');
     });
+    it('It must error when weekends-only is set and value is not in weekend period', async function () {
```
Theeraphat-Sorasetsakul:
```suggestion
    it('It should error when weekends-only is true and value is not in weekend period', async function () {
```

packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
```diff
@@ -246,5 +246,89 @@ describe('datetime-picker/Value', function () {
...
+
+      el.value = '2024-03-01';
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(true);
+
+      el.value = '2024-03-02';
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(false);
+    });
+    it('It must error when weekdays-only is set and value is not in weekday period', async function () {
```
Theeraphat-Sorasetsakul:
```suggestion
    it('It should error when weekdays-only is true and value is not in weekday period', async function () {
```

packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
```diff
@@ -246,5 +246,89 @@ describe('datetime-picker/Value', function () {
...
+
+      el.value = '2024-03-02';
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(true);
+
+      el.value = '2024-03-01';
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(false);
+    });
+    it('It must error when weekends-only is set and values are not in weekend period with range mode', async function () {
```
Theeraphat-Sorasetsakul:
```suggestion
    it('It should error when weekends-only is true and values are not in weekend period with range mode', async function () {
```

packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
```diff
@@ -246,5 +246,89 @@ describe('datetime-picker/Value', function () {
...
+
+      el.values = ['2024-03-02', '2024-03-03'];
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(false);
+
+      el.values = ['2024-03-04', '2024-03-05'];
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(true);
+    });
+    it('It must error when weekdays-only is set and values are not in weekday period with range mode', async function () {
```
Theeraphat-Sorasetsakul:
```suggestion
    it('It should error when weekdays-only is true and values are not in weekday period with range mode', async function () {
```

packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
```diff
@@ -246,5 +246,109 @@ describe('datetime-picker/Value', function () {
...
+
+      el.values = ['2024-03-03', '2024-03-09'];
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(true, 'error state should be true when both inputs value are invalid');
+
+      el.values = ['2024-03-04', '2024-03-08'];
+      await elementUpdated(el);
+      expect(el.error).to.be.equal(
+        false,
+        'error state should return to false when both inputs value are valid'
```
Theeraphat-Sorasetsakul, line 376: Shouldn't do multiple expect in one test case. And this Sound like multiple test cases.

Theeraphat-Sorasetsakul (dismissed):

packages/elements/src/datetime-picker/index.ts
```diff
@@ -1115,12 +1116,34 @@ export class DatetimePicker extends FormFieldElement implements MultiValue {
     return true;
   }
 
+  /**
+   * Check if `values` correspond to dates that are allowed within the conditions of weekdays or weekends
+   * @returns false if `values` don't correspond to dates that are allowed within the conditions of weekdays or weekends.
+   */
+  private isAllowSelectedDay(): boolean {
```
wattachai-lseg:
```suggestion
  private isValidDay(): boolean {
```

packages/elements/src/datetime-picker/index.ts
wattachai-lseg, line 1: `shouldValidateValue()` should detect change of `weekdaysOnly` & `weekendsOnly` as well.

wattachai-lseg (changes requested):

packages/elements/src/datetime-picker/__test__/datetime-picker.value.test.js
wattachai-lseg, line 1: Let's add a comment describing that <input> value update testing relies on blur event which can't be simulated due to the implementation of focus registry. There should be some existing similar comments in unit test suite. Let's copy it over.

wattachai-lseg (approved):

PR #1129, opened 2024-03-18
chore: update follow-redirects package

bualoy-napat:
`follow-redirects` need to update to latest version (1.15.6) according to the security issue reported here https://github.com/Refinitiv/refinitiv-ui/security/dependabot/121


wattachai-lseg:
only `follow-redirects` should be updated.

wattachai-lseg (approved):

PR #1145, opened 2024-04-17
docs: add Tree Node doc and its API reference

wattachai-lseg:
## Description

* Add Tree Node doc and its API reference
* handle optional context param when it's not provided to default Tree renderer

## Checklist


documents/src/pages/utils/tree-node.md
```diff
@@ -0,0 +1,147 @@
...
+];
+console.log(tree.values); // Expected output: ['3']
+
+const treeNodes = tree.manager.getTreeNodes();
+const selectedNodes = treeNodes.filter(treeNode => treeNode.icon);
+selectedNodes.forEach(node => node.selected = true);
+console.log(tree.values); // Expected output: ['2','3','4']
+```
+
+If the earlier example were [Tree Select](/elements/tree-select), `values` would remain unchanged as the selection needs to be confirmed first. This could be done by using "Done" button or calling `persistSelection()` method.
```
Theeraphat-Sorasetsakul, line 49:
```suggestion
If the earlier example was [Tree Select](/elements/tree-select), `values` would remain unchanged as the selection needs to be confirmed first. This could be done by using "Done" button or calling `persistSelection()` method.
```

packages/elements/src/tree/helpers/renderer.ts
```diff
@@ -30,12 +30,12 @@ export const createTreeRenderer = <T extends TreeDataItem = TreeDataItem>(
     element: HTMLElement = document.createElement('ef-tree-item')
   ): HTMLElement => {
     // cast type to element to not break List api.
-    const _context = context as RendererScope<T>;
+    const _context = context as RendererScope<T> | undefined;
```
Theeraphat-Sorasetsakul, line 33: What is the case that context is undefined?

wattachai-lseg: `context` is an optional parameter.

documents/src/pages/utils/tree-node.md
wattachai-lseg: "were" is already correct. Similar to "If I were you, ..."

Theeraphat-Sorasetsakul (approved):

PR #1137, opened 2024-03-29
chore: update target test version of Android to 12 - 14

wattachai-lseg:
## Description

* chore: update target test version of Android to 12 - 14
* chore: test iOS 17 on iPhone 15 Pro for better performance

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have checked my code and corrected any misspellings

PR #1146, opened 2024-04-18
fix(core): prevent error notice when define duplicate custom element into the registry

bualoy-napat:
## Description
To prevent element to throw error when custom element into registry on the production, we need to check environment before throw error to notice the developer.

Fixes # (issue)
DME-14858

## Checklist

### DIFF
--- a/packages/core/__test__/registries/CustomStyleRegistry.test.js
+++ b/packages/core/__test__/registries/CustomStyleRegistry.test.js
@@ -2,18 +2,8 @@ import { expect } from '@refinitiv-ui/test-helpers';
 
 import { DuplicateStyleError } from '../../lib/errors/DuplicateStyleError.js';
 import { CustomStyleRegistry } from '../../lib/registries/CustomStyleRegistry.js';
-import { mockCssString } from '../helper.js';
+import { isLocalhost, mockCssString } from '../helper.js';
 
-const duplicateStyleErrorMessage = `Only one theme file can be loaded per element
-
-[TestCustomStyleRegistry_2] has already been defined.
-
-Potential causes:
-1. You are trying to load a multiple variants of a theme
-2. You have loaded multiple or duplicate themes in your application bundle
-
-https://ui.refinitiv.com/kb/duplicate-styles
-`;
 describe('TestCustomStyleRegistry', function () {
   let testNum = 0;
   const baseName = 'TestCustomStyleRegistry_';
@@ -39,14 +29,17 @@ describe('TestCustomStyleRegistry', function () {
     expect(fetchedCssString).to.equal('');
   });
 
-  it('Test define twice same name', async function () {
+  it('Test define twice same name', function () {
     CustomStyleRegistry.define(testName, mockCssString);
 
-    try {
-      CustomStyleRegistry.define(testName, mockCssString);
-    } catch (error) {
-      expect(error).instanceOf(DuplicateStyleError);
-      await expect(error.message).to.equal(duplicateStyleErrorMessage);
+    if (isLocalhost) {
+      expect(() => {
+        CustomStyleRegistry.define(testName, mockCssString);
+      }).to.throw(DuplicateStyleError);
+    } else {
+      expect(() => {
+        CustomStyleRegistry.define(testName, mockCssString);
+      }).not.to.throw(DuplicateStyleError);
     }
   });
 });
--- a/packages/core/__test__/registries/ElementRegistry.test.js
+++ b/packages/core/__test__/registries/ElementRegistry.test.js
@@ -1,9 +1,9 @@
-import { expect, html } from '@refinitiv-ui/test-helpers';
+import { expect, html, nextFrame } from '@refinitiv-ui/test-helpers';
 
 import { BasicElement, css, unsafeCSS } from '../../lib/index.js';
 import { CustomStyleRegistry } from '../../lib/registries/CustomStyleRegistry.js';
 import { ElementRegistry } from '../../lib/registries/ElementRegistry.js';
-import { asyncFrames, getErrors, mockCssString, setErrors } from '../helper.js';
+import { getErrors, isLocalhost, mockCssString, setErrors } from '../helper.js';
 
 const staticMockCssString = ':host { padding: 0; }';
 const duplicationMessage = `Only one version of a Custom Element can be registered in the browser
@@ -109,12 +109,17 @@ describe('TestElementRegistry', function () {
     ElementRegistry.define(testName, MockBasicElement);
     ElementRegistry.define(testName, MockBasicElement);
 
-    await asyncFrames();
+    await nextFrame(2);
 
     const { errorCount, errorMessage } = getErrors();
 
-    await expect(errorMessage).to.equal(duplicationMessage);
-    expect(errorCount).to.equal(1);
+    if (isLocalhost) {
+      expect(errorCount).to.equal(1, 'Error not thrown');
+      expect(errorMessage).to.equal(duplicationMessage, 'duplication error not thrown in dev environment');
+    } else {
+      expect(errorCount).to.equal(0, 'Error thrown');
+      expect(errorMessage).to.equal('', 'duplication error thrown in other environment that was not dev');
+    }
 
     setErrors();
   });
@@ -185,5 +190,24 @@ describe('TestElementRegistry', function () {
     const element = document.createElement('div');
 
     expect(() => ElementRegistry.disconnect(element)).to.not.throw();
+  });
+  it('Test duplication define element', async function () {
+    let MockBasicElement = createMockClass();
+
+    customElements.define(testName, MockBasicElement);
+
+    ElementRegistry.define(testName, MockBasicElement);
+    CustomStyleRegistry.define(testName, mockCssString);
+
+    await nextFrame(2);
+
+    const { errorObject } = getErrors();
+    if (isLocalhost) {
+      expect(errorObject).to.be.instanceOf(DOMException, 'Error not thrown in dev environment');
+    } else {
+      expect(errorObject).to.equal(undefined, 'Error thrown in other environment');
+    }
+
+    setErrors();
   });
 });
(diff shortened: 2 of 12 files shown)


packages/core/src/registries/CustomStyleRegistry.ts
```diff
@@ -2,6 +2,7 @@ import { DuplicateStyleError } from '../errors/DuplicateStyleError.js';
 import { ready } from '../utils/elementReady.js';
 
 const register = new Map<string, string>();
+const DEV_ENV = /^(localhost|127\.0\.0\.1)$/.test(location.hostname);
```
wattachai-lseg: dev environment should be defined in one place only.

packages/test-helpers/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@refinitiv-ui/test-helpers",
-  "version": "7.1.0",
+  "version": "7.1.1-next.0",
```
wattachai-lseg: This is about registries update of `core` package. `test-helpers`' version should not be bumped. Let's find out what's going on.

wattachai-lseg (changes requested):

packages/core/src/registries/CustomStyleRegistry.ts
bualoy-napat: Fixed!

packages/test-helpers/package.json
bualoy-napat: I guess that it relates to https://github.com/Refinitiv/refinitiv-ui/pull/1137 where we update the dependency of test-helpers, and we didn't release it yet.

packages/core/src/utils/helpers.ts
```diff
@@ -1,6 +1,7 @@
 import type { BasicElement } from '../elements/BasicElement';
 
 const BasicElementSymbol = Symbol('BasicElement');
+const isDevEnvironment = /^(localhost|127\.0\.0\.1)$/.test(location.hostname);
```
wattachai-lseg:
```suggestion
const isLocalhost = /^(localhost|127\.0\.0\.1)$/.test(location.hostname);
```

packages/core/src/utils/elementReady.ts
```diff
@@ -15,9 +17,11 @@ export const ready = function (name: string, callback?: ReadyHandler): void {
     try {
       callbackCollection.forEach((callback) => callback && callback());
     } catch (e) {
-      setTimeout(() => {
-        throw e;
-      });
+      if (isDevEnvironment) {
```
wattachai-lseg: ~I don't think we need this if statement. Could you double check it?~

Update: this is actually needed as there will be multiple instance of EFs in micro-frontend setup.
However, the condition should be more specific so that it catches duplicate element errors only.

packages/core/src/utils/helpers.ts
bualoy-napat: Fixed!

packages/core/src/utils/elementReady.ts
bualoy-napat: Fixed!

packages/core/src/utils/elementReady.ts
```diff
@@ -15,9 +17,19 @@ export const ready = function (name: string, callback?: ReadyHandler): void {
...
+        if ((isLocalhost && isCustomElementRegistryError) || !isCustomElementRegistryError) {
+          setTimeout(() => {
+            throw e;
+          });
+        }
+      } else {
+        setTimeout(() => {
+          throw e;
+        });
+      }
```
wattachai-lseg:
```suggestion
      if (!(isLocalhost && e instanceof DOMException &&
        e.name === 'NotSupportedError' &&
        e.message.includes('CustomElementRegistry'))) {
          setTimeout(() => {
            throw e;
          });
      }
```

packages/core/__test__/registries/ElementRegistry.test.js
```diff
@@ -104,6 +104,8 @@ describe('TestElementRegistry', function () {
   });
 
   it('Test define twice same name', async function () {
+    // eslint-disable-next-line no-console
+    console.info('hostname:', location.hostname);
```
wattachai-lseg: let's forget about BrowserStack and update test case itself instead. Expect exceptions from localhost only + expect no exceptions from non-localhost hostname.

packages/core/src/utils/elementReady.ts
wattachai-lseg: one more thing, have you confirmed that the error message for duplicate custom elements of all 3 major browsers contain 'CustomElementRegistry'?

packages/core/src/registries/ElementRegistry.ts
```diff
@@ -17,7 +18,10 @@ const register = new Map<string, ElementRegistrationItem>();
 
 const upgrade = (name: string, definition: ElementConstructor): void => {
   definition.applyThemeStyles(CustomStyleRegistry.get(name));
-  customElements.define(name, definition);
+  const isDefined = customElements.get(name);
+  if (isLocalhost || !isDefined) {
```
wattachai-lseg: let's add a comment here: `For localhost, define custom element even when it's been defined already to inform developers about the issue.`

packages/core/__test__/utils/elementReady.test.js
```diff
@@ -55,8 +55,8 @@ describe('TestReady', function () {
 
     const { errorMessage, errorCount } = getErrors();
 
-    expect(errorMessage).to.equal('test');
-    expect(errorCount).to.equal(1);
+    expect(errorCount).to.equal(1, 'Error not thrown');
+    await expect(errorMessage).to.equal('test');
```
wattachai-lseg: This looks incomplete. Did you forget to push your change?

packages/core/src/registries/NativeStyleRegistry.ts
```diff
@@ -14,9 +15,12 @@ export abstract class NativeStyleRegistry {
    * @returns {void}
    */
   public static define(name: string, css: string): void {
-    if (register.has(name)) {
+    if (register.has(name) && isLocalhost) {
       throw new DuplicateStyleError(name);
+    } else if (register.has(name) && !isLocalhost) {
```
wattachai-lseg: There is no need to check `register.has(name)` twice.

packages/core/src/registries/CustomStyleRegistry.ts
```diff
@@ -15,8 +16,10 @@ export abstract class CustomStyleRegistry {
    * @returns {void}
    */
   public static define(name: string, css: string): void {
-    if (register.has(name)) {
+    if (register.has(name) && isLocalhost) {
       throw new DuplicateStyleError(name);
+    } else if (register.has(name) && !isLocalhost) {
```
wattachai-lseg: There is no need to check `register.has(name)` twice.

packages/core/src/registries/ElementRegistry.ts
bualoy-napat: Fixed

packages/core/__test__/utils/elementReady.test.js
bualoy-napat: Fixed

packages/core/src/registries/NativeStyleRegistry.ts
bualoy-napat: Fixed

packages/core/src/registries/CustomStyleRegistry.ts
bualoy-napat: Fixed

packages/core/__test__/registries/CustomStyleRegistry.test.js
```diff
@@ -39,14 +39,14 @@ describe('TestCustomStyleRegistry', function () {
     expect(fetchedCssString).to.equal('');
   });
 
-  it('Test define twice same name', async function () {
+  it('Test define twice same name', function () {
     CustomStyleRegistry.define(testName, mockCssString);
 
     try {
```
wattachai-lseg: we should use `expect` with callback similar to `packages/core/__test__/decorators/CustomElement.test.js` to test that there is no error thrown in non localhost environment as well.

packages/core/__test__/registries/NativeStyleRegistry.test.js
```diff
@@ -38,14 +38,14 @@ describe('TestNativeStyleRegistry', function () {
     expect(fetchedCssString).to.equal('');
   });
 
-  it('Test define twice same name', async function () {
+  it('Test define twice same name', function () {
     NativeStyleRegistry.define(testName, mockCssString);
 
     try {
```
wattachai-lseg: we should use `expect` with callback similar to `packages/core/__test__/decorators/CustomElement.test.js` to test that there is no error thrown in non localhost environment as well.

packages/core/__test__/utils/elementReady.test.js
```diff
@@ -55,8 +55,8 @@ describe('TestReady', function () {
 
     const { errorMessage, errorCount } = getErrors();
 
+    expect(errorCount).to.equal(1, 'Error not thrown');
     expect(errorMessage).to.equal('test');
```
wattachai-lseg: missing `message` for test failure.

wattachai-lseg (changes requested):

packages/core/__test__/utils/elementReady.test.js
```diff
@@ -55,8 +55,8 @@ describe('TestReady', function () {
 
     const { errorMessage, errorCount } = getErrors();
 
-    expect(errorMessage).to.equal('test');
-    expect(errorCount).to.equal(1);
+    expect(errorCount).to.equal(1, 'Error not thrown');
+    expect(errorMessage).to.equal('test', 'duplication error not thrown in dev environment');
```
wattachai-lseg: Forgot to update message for test failure?

bualoy-napat: Fixed ;)

wattachai-lseg:
LGTM

wattachai-lseg (approved):

Theeraphat-Sorasetsakul:
Verified.

PR #1156, opened 2024-05-02
fix(core): prevent error notice when define duplicate custom element into the registry

bualoy-napat:
## Description
To prevent element to throw error when custom element into registry on the production, we need to check environment before throw error to notice the developer.

Fixes # (issue)
DME-14858

## Checklist

### DIFF
--- a/packages/core/__test__/utils/elementReady.test.js
+++ b/packages/core/__test__/utils/elementReady.test.js
@@ -56,7 +56,7 @@ describe('TestReady', function () {
     const { errorMessage, errorCount } = getErrors();
 
     expect(errorCount).to.equal(1, 'Error not thrown');
-    expect(errorMessage).to.equal('test', 'duplication error not thrown');
+    expect(errorMessage).to.equal('test', 'test error not thrown');
 
     setErrors();
   });
--- a/packages/core/__test__/utils/elementReady.test.js
+++ b/packages/core/__test__/utils/elementReady.test.js
@@ -1,7 +1,7 @@
-import { expect } from '@refinitiv-ui/test-helpers';
+import { expect, nextFrame } from '@refinitiv-ui/test-helpers';
 
 import { ready } from '../../lib/utils/elementReady.js';
-import { asyncFrames, getErrors, setErrors } from '../helper.js';
+import { getErrors, setErrors } from '../helper.js';
 
 describe('TestReady', function () {
   const baseName = 'TestReady_';
@@ -51,12 +51,12 @@ describe('TestReady', function () {
 
     expect(callbackCallCount).to.equal(1, 'First callback was not called before throw error');
 
-    await asyncFrames();
+    await nextFrame(2);
 
     const { errorMessage, errorCount } = getErrors();
 
-    expect(errorMessage).to.equal('test');
-    expect(errorCount).to.equal(1);
+    expect(errorCount).to.equal(1, 'Error not thrown');
+    expect(errorMessage).to.equal('test', 'duplication error not thrown');
 
     setErrors();
   });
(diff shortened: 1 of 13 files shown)


packages/core/__test__/utils/elementReady.test.js
```diff
@@ -51,12 +51,12 @@ describe('TestReady', function () {
...
 
-    await asyncFrames();
+    await nextFrame(2);
 
     const { errorMessage, errorCount } = getErrors();
 
-    expect(errorMessage).to.equal('test');
-    expect(errorCount).to.equal(1);
+    expect(errorCount).to.equal(1, 'Error not thrown');
+    expect(errorMessage).to.equal('test', 'duplication error not thrown');
```
wattachai-lseg:
```suggestion
    expect(errorMessage).to.equal('test', 'test error not thrown');
```

packages/core/src/registries/CustomStyleRegistry.ts
```diff
@@ -16,8 +17,13 @@ export abstract class CustomStyleRegistry {
    */
   public static define(name: string, css: string): void {
     if (register.has(name)) {
-      throw new DuplicateStyleError(name);
+      if (isLocalhost) {
```
wattachai-lseg, line 21: `/* istanbul ignore else  */` on top of if statement instead of `/* istanbul ignore next */` should be better.

packages/core/src/registries/NativeStyleRegistry.ts
```diff
@@ -16,8 +17,13 @@ export abstract class NativeStyleRegistry {
    */
   public static define(name: string, css: string): void {
     if (register.has(name)) {
-      throw new DuplicateStyleError(name);
+      if (isLocalhost) {
```
wattachai-lseg, line 21: `/* istanbul ignore else  */` on top of if statement instead of `/* istanbul ignore next */` should be better.

wattachai-lseg (changes requested):
A few more things on test coverage.

packages/core/src/registries/CustomStyleRegistry.ts
bualoy-napat: Done

packages/core/__test__/utils/elementReady.test.js
bualoy-napat: Done

packages/core/src/registries/NativeStyleRegistry.ts
bualoy-napat: Done

packages/core/src/registries/CustomStyleRegistry.ts
wattachai-lseg: Looks like we need both. 🤦  Could you bring back `/* istanbul ignore next */` as well.

packages/core/src/registries/NativeStyleRegistry.ts
wattachai-lseg: Looks like we need both. 🤦  Could you bring back `/* istanbul ignore next */` as well.

packages/core/src/registries/ElementRegistry.ts
```diff
@@ -17,7 +18,11 @@ const register = new Map<string, ElementRegistrationItem>();
 
 const upgrade = (name: string, definition: ElementConstructor): void => {
   definition.applyThemeStyles(CustomStyleRegistry.get(name));
-  customElements.define(name, definition);
+  const isElementDefined = customElements.get(name);
+  // For localhost, define custom element even when it's been defined already to inform developers about the issue.
+  if (isLocalhost || !isElementDefined) {
```
wattachai-lseg, line 23: This should improve test coverage.
```suggestion
  if (isLocalhost ||
      /* istanbul ignore next */ !isElementDefined) {
```

wattachai-lseg: confirmed that it doesn't improve test coverage. Marking as resolved.

wattachai-lseg (approved):

Theeraphat-Sorasetsakul:
Verified.

PR #1161, opened 2024-05-13
fix(chart): fix gridlines colors

Theeraphat-Sorasetsakul:
## Description

In chart, when create chart elements more than one. The gridline of other charts except the first don't show the same style.

Fixes # ([DME-7098](https://jira.refinitiv.com/browse/DME-7098))

## Checklist

### DIFF
--- a/packages/elements/src/chart/__test__/chart.test.js
+++ b/packages/elements/src/chart/__test__/chart.test.js
@@ -352,6 +352,24 @@ describe('chart/Chart', function () {
         expect(labelstrokeStyle).to.equal(labels[0].strokeStyle);
       }
     });
+
+    it('Should be able to change grid line color', async function () {
+      const gridLineColor = '#FF0000';
+      el = await fixture(`<ef-chart style="--grid-line-color: '${gridLineColor}'"></ef-chart>`);
+      el.config = config.line;
+      await chartRendered(el);
+      expect(el.chart.options.scales.x.grid.color).to.equal(el.getComputedVariable('--grid-line-color'));
+      expect(el.chart.options.scales.y.grid.color).to.equal(el.getComputedVariable('--grid-line-color'));
+    });
+
+    it('Should be able to change zero grid line color', async function () {
+      const zeroLineColor = '#FF0000';
+      el = await fixture(`<ef-chart style="--grid-line-color: '${zeroLineColor}'"></ef-chart>`);
+      el.config = config.line;
+      await chartRendered(el);
+      expect(el.chart.options.scales.x.border.color).to.equal(el.getComputedVariable('--zero-line-color'));
+      expect(el.chart.options.scales.y.border.color).to.equal(el.getComputedVariable('--zero-line-color'));
+    });
   });
 
   describe('Plugins', function () {
--- a/packages/elements/src/chart/__test__/chart.test.js
+++ b/packages/elements/src/chart/__test__/chart.test.js
@@ -364,11 +364,11 @@ describe('chart/Chart', function () {
 
     it('Should be able to change zero grid line color', async function () {
       const zeroLineColor = '#FF0000';
-      el = await fixture(`<ef-chart style="--grid-line-color: '${zeroLineColor}'"></ef-chart>`);
+      el = await fixture(`<ef-chart style="--zero-line-color: '${zeroLineColor}'"></ef-chart>`);
       el.config = config.line;
       await chartRendered(el);
-      expect(el.chart.options.scales.x.border.color).to.equal(el.getComputedVariable('--zero-line-color'));
-      expect(el.chart.options.scales.y.border.color).to.equal(el.getComputedVariable('--zero-line-color'));
+      expect(el.chart.options.scales.x.border.color).to.equal(zeroLineColor);
+      expect(el.chart.options.scales.y.border.color).to.equal(zeroLineColor);
     });
   });
 
--- a/packages/elements/src/chart/__test__/chart.test.js
+++ b/packages/elements/src/chart/__test__/chart.test.js
@@ -358,8 +358,8 @@ describe('chart/Chart', function () {
       el = await fixture(`<ef-chart style="--grid-line-color: '${gridLineColor}'"></ef-chart>`);
       el.config = config.line;
       await chartRendered(el);
-      expect(el.chart.options.scales.x.grid.color).to.equal(el.getComputedVariable('--grid-line-color'));
-      expect(el.chart.options.scales.y.grid.color).to.equal(el.getComputedVariable('--grid-line-color'));
+      expect(el.chart.options.scales.x.grid.color).to.equal(gridLineColor);
+      expect(el.chart.options.scales.y.grid.color).to.equal(gridLineColor);
     });
 
     it('Should be able to change zero grid line color', async function () {
--- a/packages/elements/src/chart/elements/chart.ts
+++ b/packages/elements/src/chart/elements/chart.ts
@@ -1,5 +1,6 @@
 import { Chart as ChartJS } from 'chart.js';
 import type {
+  CartesianScaleOptions,
   ChartConfiguration,
   ChartDataset,
   ChartOptions,
@@ -290,13 +291,6 @@ export class Chart extends BasicElement {
       | 'initial'
       | 'inherit'
       | undefined;
-    // Set global grid color
-    ChartJS.defaults.scale.grid.color = (line) => {
-      return line.index === 0
-        ? this.getComputedVariable('--zero-line-color', 'transparent')
-        : this.getComputedVariable('--grid-line-color', 'transparent');
-    };
-
     if (ChartJS.defaults.scales.radialLinear) {
       ChartJS.defaults.scales.radialLinear.ticks.showLabelBackdrop = false;
     }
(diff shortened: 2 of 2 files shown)


packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,29 @@ export class Chart extends BasicElement {
...
+    this.decorateGridColors(chart);
   };
 
+  /**
+   * Set grid color in beforeUpdate event from Chartjs
+   * @param chart Chart.js instance
+   * @returns {void}
+   */
+  private decorateGridColors(chart: ChartJS): void {
+    for (const scale in chart.scales) {
```
wattachai-lseg: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#iterating_over_the_properties_of_an_object
```suggestion
    for (const scale of Object.keys(chart.scales)) {
```

packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,29 @@ export class Chart extends BasicElement {
...
+      const axis = chart.options?.scales?.[scale] as CartesianScaleOptions;
+      const userGridColor = this.config?.options?.scales?.[scale]?.grid?.color as Color;
```
wattachai-lseg: no need to type cast here, right?
```suggestion
      const userGridColor = this.config?.options?.scales?.[scale]?.grid?.color;
```

packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,29 @@ export class Chart extends BasicElement {
...
   };
 
+  /**
+   * Set grid color in beforeUpdate event from Chartjs
+   * @param chart Chart.js instance
+   * @returns {void}
+   */
+  private decorateGridColors(chart: ChartJS): void {
+    for (const scale in chart.scales) {
+      const axis = chart.options?.scales?.[scale] as CartesianScaleOptions;
```
wattachai-lseg: `axis` could be `undefined`. 
```suggestion
      const axis = chart.options?.scales?.[scale] as CartesianScaleOptions | undefined;
```

packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,29 @@ export class Chart extends BasicElement {
...
+      const userGridColor = this.config?.options?.scales?.[scale]?.grid?.color as Color;
+      // rgba(0,0,0,0.1) is unset grid color
```
wattachai-lseg: It seems `rgba(0,0,0,0.1)` is the default grid color. We're checking for that value so the grid color would be updated when it's necessary only, right? Let's comment about that.

packages/elements/src/chart/elements/chart.ts
```diff
@@ -512,6 +527,7 @@ export class Chart extends BasicElement {
    * Merges all the different layers of the config.
    * @returns {void}
    */
+  // refactor this func to immutable that return merged config
```
wattachai-lseg: Not sure I understand. A TODO? It should elaborate more.

wattachai-lseg (changes requested):

packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,29 @@ export class Chart extends BasicElement {
...
     this.decorateColors(chart);
+    this.decorateGridColors(chart);
   };
 
+  /**
+   * Set grid color in beforeUpdate event from Chartjs
+   * @param chart Chart.js instance
+   * @returns {void}
+   */
+  private decorateGridColors(chart: ChartJS): void {
```
wattachai-lseg, line 338: There is no unit test for this function, right? Let's exclude this function from test coverage.

packages/elements/src/chart/elements/chart.ts
Theeraphat-Sorasetsakul: removed.

packages/elements/src/chart/elements/chart.ts
Theeraphat-Sorasetsakul: Fixed.

packages/elements/src/chart/elements/chart.ts
Theeraphat-Sorasetsakul: Added test for this case.

packages/elements/src/chart/elements/chart.ts
wattachai-lseg: It's still there...

packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,30 @@ export class Chart extends BasicElement {
...
+    for (const scale of Object.keys(chart.scales)) {
+      const axis = chart.options?.scales?.[scale] as CartesianScaleOptions | undefined;
+      const userAxis = this.config?.options?.scales?.[scale] as CartesianScaleOptions | undefined;
+      const userGridColor = userAxis?.grid?.color;
+      const userZeroGridColor = userAxis?.border?.color;
```
wattachai-lseg: Zero grid color is no where to be found.
```suggestion
      const userBorderColor = userAxis?.border?.color;
```

packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,30 @@ export class Chart extends BasicElement {
...
+      // Change grid color if the color is default which rgba(0,0,0,0.1) is default grid color.
```
wattachai-lseg:
```suggestion
      // Change grid color if the color is the default value of `rgba(0,0,0,0.1)`
```

packages/elements/src/chart/elements/chart.ts
```diff
@@ -333,8 +327,30 @@ export class Chart extends BasicElement {
...
+      if (axis && userGridColor === 'rgba(0,0,0,0.1)') {
+        axis.grid.color = this.getComputedVariable('--grid-line-color', 'transparent');
+      }
+      if (axis && userZeroGridColor === 'rgba(0,0,0,0.1)') {
```
wattachai-lseg: It seems the default value also applies here. Let's add a similar comment for border color as well.

packages/elements/src/chart/elements/chart.ts
Theeraphat-Sorasetsakul: Oh sorry. Going to remove immediately. Removed.

packages/elements/src/chart/elements/chart.ts
```diff
@@ -512,6 +528,7 @@ export class Chart extends BasicElement {
    * Merges all the different layers of the config.
    * @returns {void}
    */
+  // refactor this func to immutable that return merged config
```
Theeraphat-Sorasetsakul:
```suggestion
```

packages/elements/src/chart/__test__/chart.test.js
```diff
@@ -352,6 +352,24 @@ describe('chart/Chart', function () {
...
       }
     });
+
+    it('Should be able to change grid line color', async function () {
+      const gridLineColor = '#FF0000';
+      el = await fixture(`<ef-chart style="--grid-line-color: '${gridLineColor}'"></ef-chart>`);
+      el.config = config.line;
+      await chartRendered(el);
+      expect(el.chart.options.scales.x.grid.color).to.equal(el.getComputedVariable('--grid-line-color'));
+      expect(el.chart.options.scales.y.grid.color).to.equal(el.getComputedVariable('--grid-line-color'));
```
wattachai-lseg: Checking the computed value would not be a test as it would be always `true`.
```suggestion
      expect(el.chart.options.scales.x.grid.color).to.equal(gridLineColor);
      expect(el.chart.options.scales.y.grid.color).to.equal(gridLineColor);
```

packages/elements/src/chart/__test__/chart.test.js
```diff
@@ -352,6 +352,24 @@ describe('chart/Chart', function () {
...
+    });
+
+    it('Should be able to change zero grid line color', async function () {
+      const zeroLineColor = '#FF0000';
+      el = await fixture(`<ef-chart style="--grid-line-color: '${zeroLineColor}'"></ef-chart>`);
+      el.config = config.line;
+      await chartRendered(el);
+      expect(el.chart.options.scales.x.border.color).to.equal(el.getComputedVariable('--zero-line-color'));
+      expect(el.chart.options.scales.y.border.color).to.equal(el.getComputedVariable('--zero-line-color'));
```
wattachai-lseg: Similar to above with an addition of incorrect custom property name.
```suggestion
      el = await fixture(`<ef-chart style="--zero-line-color: '${zeroLineColor}'"></ef-chart>`);
      el.config = config.line;
      await chartRendered(el);
      expect(el.chart.options.scales.x.border.color).to.equal(zeroLineColor);
      expect(el.chart.options.scales.y.border.color).to.equal(zeroLineColor);
```

wattachai-lseg (changes requested):

packages/elements/src/chart/elements/chart.ts
Theeraphat-Sorasetsakul: Fixed.

wattachai-lseg (approved):
LGTM

bualoy-napat:
LGTM

PR #1198, opened 2024-07-23
fix(slider): sync emitted value and input value

prima-lseg:
## Description

Emitted value doesn't match with value that show on input.

Fixes # (issue)
[DME-17777](https://jira.refinitiv.com/browse/DME-17777)

## Checklist

### DIFF
--- a/packages/elements/src/slider/elements/slider.ts
+++ b/packages/elements/src/slider/elements/slider.ts
@@ -954,8 +954,7 @@ export class Slider extends FormFieldElement {
     const valueNumber = Number(value);
     shouldUpdate = this.isValueInBoundary(valueNumber, name);
 
-    // Calculate updated value based on the current slider value.
-    // From and to value is calculate with min range
+    // adjust update-pending from/to value based on min range
     const calculateValue = (currentValue: string): string => {
       const _currentValue = Number(currentValue);
       if (name === SliderDataName.from && _currentValue === this.toNumber) {
--- a/packages/elements/src/slider/elements/slider.ts
+++ b/packages/elements/src/slider/elements/slider.ts
@@ -971,7 +971,8 @@ export class Slider extends FormFieldElement {
       this.notifyPropertyChange(name, this[name]);
     }
 
-    // Revert number field value back to the pre-update value
+    // Sync number field value with its component property counterpart.
+    // It's crucial especially when the value has been adjusted above.
     const inputName = `${name}Input`;
     this[inputName as NumberFieldName].value = this[name];
   }
--- a/packages/elements/src/slider/elements/slider.ts
+++ b/packages/elements/src/slider/elements/slider.ts
@@ -970,10 +970,10 @@ export class Slider extends FormFieldElement {
     if (shouldUpdate) {
       this[name] = calculateValue(this.format(valueNumber));
       this.notifyPropertyChange(name, this[name]);
-    } else {
-      const inputName = `${name}Input`;
-      this[inputName as NumberFieldName].value = this[name];
     }
+
+    const inputName = `${name}Input`;
+    this[inputName as NumberFieldName].value = this[name];
   }
 
   /**
@@ -1315,9 +1315,15 @@ export class Slider extends FormFieldElement {
       return false;
     }
     if (this.range) {
-      if (valueFor === SliderDataName.to && value < this.fromNumber) {
+      if (
+        valueFor === SliderDataName.to &&
+        (value < this.fromNumber || (this.minRangeNumber > 0 && value <= this.fromNumber))
+      ) {
         return false;
-      } else if (valueFor === SliderDataName.from && value > this.toNumber) {
+      } else if (
+        valueFor === SliderDataName.from &&
+        (value > this.toNumber || (this.minRangeNumber > 0 && value >= this.toNumber))
+      ) {
         return false;
       }
     }
--- a/packages/elements/src/slider/elements/slider.ts
+++ b/packages/elements/src/slider/elements/slider.ts
@@ -954,23 +954,24 @@ export class Slider extends FormFieldElement {
     const valueNumber = Number(value);
     shouldUpdate = this.isValueInBoundary(valueNumber, name);
 
-    // adjust update-pending from/to value based on min range
-    const calculateValue = (currentValue: string): string => {
-      const _currentValue = Number(currentValue);
+    // Adjust update-pending from/to value based on min range
+    const adjustValue = (value: string): string => {
+      const _currentValue = Number(value);
       if (name === SliderDataName.from && _currentValue === this.toNumber) {
         return String(_currentValue - this.minRangeNumber);
       }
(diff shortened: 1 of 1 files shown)


packages/elements/src/slider/elements/slider.ts
```diff
@@ -1263,7 +1273,7 @@ export class Slider extends FormFieldElement {
    * @returns {void}
    */
   private onFromValueChange(): void {
-    if (this.isValueInBoundary(this.fromNumber, SliderDataName.from)) {
+    if (this.isValueInBoundary(this.fromNumber + this.minRangeNumber, SliderDataName.from)) {
```
Theeraphat-Sorasetsakul, line 1287: Can you move this logic to `isValueInBoundary`? Do that to both onFromChange and onTochange.

packages/elements/src/slider/elements/slider.ts
```diff
@@ -951,16 +951,26 @@ export class Slider extends FormFieldElement {
...
+      }
+      return currentValue;
+    };
 
     if (shouldUpdate) {
-      this[name] = value;
-      this.notifyPropertyChange(name, value);
+      const adjustedValue = calculateValue(this.format(valueNumber));
+      this[name] = adjustedValue;
+      this.notifyPropertyChange(name, adjustedValue);
```
Theeraphat-Sorasetsakul:
```suggestion
      this[name] = calculateValue(this.format(valueNumber));
      this.notifyPropertyChange(name, this[name]);
```

packages/elements/src/slider/elements/slider.ts
Theeraphat-Sorasetsakul: I mean move only logic of the new change.

packages/elements/src/slider/elements/slider.ts
```diff
@@ -951,20 +951,29 @@ export class Slider extends FormFieldElement {
...
-    if (name === SliderDataName.to) {
-      shouldUpdate = this.isValueInBoundary(Number(value), SliderDataName.to);
-    } else {
-      shouldUpdate = this.isValueInBoundary(Number(value), '');
-    }
+    const valueNumber = Number(value);
+    shouldUpdate = this.isValueInBoundary(valueNumber, name);
+
+    // Calculate updated value based on the current slider value.
+    // From and to value is calculate with min range
```
wattachai-lseg:
```suggestion
    // adjust update-pending from/to value based on min range
```

packages/elements/src/slider/elements/slider.ts
```diff
@@ -951,20 +951,29 @@ export class Slider extends FormFieldElement {
...
+    const calculateValue = (currentValue: string): string => {
```
wattachai-lseg: * `adjustValue` should be a better name than `calculateValue`.
* keep it simple with `value` instead of `currentValue`?

packages/elements/src/slider/elements/slider.ts
```diff
@@ -951,20 +951,29 @@ export class Slider extends FormFieldElement {
...
     if (shouldUpdate) {
-      this[name] = value;
-      this.notifyPropertyChange(name, value);
-    } else {
-      const inputName = `${name}Input`;
-      this[inputName as NumberFieldName].value = this[name];
+      this[name] = calculateValue(this.format(valueNumber));
+      this.notifyPropertyChange(name, this[name]);
     }
+
```
wattachai-lseg, line 973: let's add a comment: "revert number field value back to the pre-update value".

packages/elements/src/slider/elements/slider.ts
```diff
@@ -1304,9 +1315,15 @@ export class Slider extends FormFieldElement {
...
+        valueFor === SliderDataName.to &&
+        (value < this.fromNumber || (this.minRangeNumber > 0 && value <= this.fromNumber))
+      ) {
         return false;
-      } else if (valueFor === SliderDataName.from && value > this.toNumber - this.minRangeNumber) {
+      } else if (
+        valueFor === SliderDataName.from &&
+        (value > this.toNumber || (this.minRangeNumber > 0 && value >= this.toNumber))
+      ) {
         return false;
```
wattachai-lseg: As both conditions return `false`, let's combine them.
Here's an idea:

```javascript
const isFromOutOfBound = ...;
const isToOutOfBound = ...;
if (isFromOutOfBound || isToOutOfBound) { return false; }
```

packages/elements/src/slider/elements/slider.ts
```diff
@@ -1318,7 +1335,8 @@ export class Slider extends FormFieldElement {
    * @returns {void}
    */
   private onToValueChange(): void {
-    if (this.isValueInBoundary(this.toNumber, SliderDataName.to)) {
+    // recalculate value and check boundary before rounding
```
wattachai-lseg: Let's replace the current "what" comment why "comment" as written below.
```javascript
/** 
 * This method handles "to" value change triggered by 
 * both user interaction and programmatic update.
 * 
 * In case of user interaction update, 
 * the value would be adjusted by min range before entering this method. 
 * That adjustment needs to be reverted.
 * 
 * In case of programmatic value update, 
 * This method could handle the revert although there is no prior adjustment.
*/

File: packages/elements/src/slider/elements/slider.ts [hunk 8]
```diff
@@ -1263,7 +1272,8 @@ export class Slider extends FormFieldElement {
    * @returns {void}
    */
   private onFromValueChange(): void {
-    if (this.isValueInBoundary(this.fromNumber, SliderDataName.from)) {
+    // recalculate value and check boundary before rounding
```
wattachai-lseg: Let's replace the current "what" comment why "comment" as written below.
```javascript
/** 
 * This method handles "from" value change triggered by 
 * both user interaction and programmatic update.
 * 
 * In case of user interaction update, 
 * the value would be adjusted by min range before entering this method. 
 * That adjustment needs to be reverted.
 * 
 * In case of programmatic value update, 
 * This method could handle the revert although there is no prior adjustment.
*/

**Review by wattachai-lseg — CHANGES_REQUESTED** [review]

File: packages/elements/src/slider/elements/slider.ts [hunk 9]
```diff
@@ -951,20 +951,29 @@ export class Slider extends FormFieldElement {
...
-      this[name] = value;
-      this.notifyPropertyChange(name, value);
-    } else {
-      const inputName = `${name}Input`;
-      this[inputName as NumberFieldName].value = this[name];
+      this[name] = adjustValue(this.format(valueNumber));
+      this.notifyPropertyChange(name, this[name]);
     }
+
+    // Revert number field value back to the pre-update value
```
wattachai-lseg: Actually, "revert" doesn't completely cover what is happening here. My bad. Let's revise the comment.
```suggestion
    // Sync number field value with its component property counterpart.
    // It's crucial especially when the value has been adjusted above.
```

wattachai-lseg (approved):

PR #1276, opened 2025-04-22
fix(tooltip): fix tooltip reopen when mouse leaves

Theeraphat-Sorasetsakul:
## Description

tooltips sometimes reappear when mouse pointer leaves the frame.

Fixes # ([DME-48231](https://jira.refinitiv.com/browse/DME-48231))

## Checklist

### DIFF
--- a/packages/elements/src/tooltip/index.ts
+++ b/packages/elements/src/tooltip/index.ts
@@ -46,6 +46,10 @@ class Tooltip extends BasicElement {
   private clicked = false;
   private timerTimeout?: number;
   private contentNodes?: Node[];
+  /**
+   * Boolean to check if the timeout is pending
+   */
+  private timeoutPending = false;
 
   protected override readonly defaultRole: string | null = 'tooltip';
 
@@ -220,14 +224,14 @@ class Tooltip extends BasicElement {
   public override connectedCallback(): void {
     super.connectedCallback();
     register(this, {
-      mousemove: this.reset,
+      mousemove: this.reset.bind(null, true),
       mousemoveThrottled: this.onMouseMove,
       click: this.onClick,
       mouseout: this.onMouseOut,
-      mouseleave: this.resetTooltip,
-      wheel: this.resetTooltip,
-      keydown: this.resetTooltip,
-      blur: this.resetTooltip
+      mouseleave: this.resetTooltip.bind(null, false),
+      wheel: this.resetTooltip.bind(null, false),
+      keydown: this.resetTooltip.bind(null, false),
+      blur: this.resetTooltip.bind(null, false)
     });
   }
 
@@ -235,7 +239,7 @@ class Tooltip extends BasicElement {
     deregister(this);
     this.setOpened(false);
 
-    this.reset();
+    this.reset(false);
     this.matchTarget = null;
     this.matchTargetRect = null;
     this.positionTarget = null;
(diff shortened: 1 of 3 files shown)


packages/elements/src/tooltip/index.ts
```diff
@@ -252,9 +256,11 @@ class Tooltip extends BasicElement {
 
   /**
    * Clear all timers
+   * @param enter Indicates whether the tooltip should be reset due to a mouse leave event
    * @returns {void}
    */
-  private reset = (): void => {
+  private reset = (enter?: boolean): void => {
```
wattachai-lseg: use default value instead
```suggestion
  private reset = (enter = false): void => {
```

packages/elements/src/tooltip/index.ts
```diff
@@ -380,10 +386,11 @@ class Tooltip extends BasicElement {
 
   /**
    * Hide tooltip
+   * @param leave Indicates whether the tooltip should be reset due to any events
    * @returns {void}
    */
-  private hideTooltip(): void {
-    this.reset();
+  private hideTooltip(leave?: boolean): void {
```
wattachai-lseg:
```suggestion
  private hideTooltip(leave = false): void {
```

packages/elements/src/tooltip/index.ts
```diff
@@ -252,9 +256,11 @@ class Tooltip extends BasicElement {
 
   /**
    * Clear all timers
+   * @param enter Indicates whether the tooltip should be reset due to a mouse leave event
    * @returns {void}
    */
-  private reset = (): void => {
+  private reset = (enter?: boolean): void => {
+    this.timeoutPending = !!enter;
```
wattachai-lseg: The timeout is going to be cleared anyway. `timeoutPending` is not a right name for this flag.

packages/elements/src/tooltip/index.ts
```diff
@@ -392,10 +399,11 @@ class Tooltip extends BasicElement {
...
   /**
    * Reset tooltip
+   * @param timeoutPending Indicates whether the tooltip should be reset due to any events
    * @returns {void}
    */
-  private resetTooltip = (): void => {
-    this.hideTooltip();
+  private resetTooltip = (timeoutPending?: boolean): void => {
+    this.hideTooltip(timeoutPending);
   };
```
wattachai-lseg: Remove this method as it's just an alias of `hideTooltip`.

wattachai-lseg (changes requested):

packages/elements/src/tooltip/managers/tooltip-manager.ts
```diff
@@ -66,6 +66,7 @@ class TooltipManager {
    * @returns {void}
    */
   private onMouseOut = (event: MouseEvent): void => {
+    this.titleThrottler.task?.cancel();
```
wattachai-lseg: Let's refactor this into a method with the following comment:


> Cancel the pending task of throttled `mousemove` event handler.
> This prevents the task from running out of order compared to the event dispatching sequence due to its `setTimeout` usage.
> Any event handler hiding the tooltip should call this method.

wattachai-lseg (changes requested):
one thing.

packages/elements/src/tooltip/managers/tooltip-manager.ts
wattachai-lseg: one more thing, click event handler should call this method as well.

wattachai-lseg (approved):

PR #1284, opened 2025-06-25
fix(icon): throw error once per session if sprite icon fails to load

bualoy-napat:
## Description

Update SpriteLoader to throw error message once per session.

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -47,6 +47,7 @@ export class LitElement extends BasicElement {
  */
 export { WarningNotice } from './notices/WarningNotice.js';
 export { DeprecationNotice } from './notices/DeprecationNotice.js';
+export { ErrorNotice } from './notices/ErrorNotice.js';
 
 /**
  * Export events
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -47,7 +47,7 @@ export class LitElement extends BasicElement {
  */
 export { WarningNotice } from './notices/WarningNotice.js';
 export { DeprecationNotice } from './notices/DeprecationNotice.js';
-export { ErrorNotice } from './notices/ErrorNotice.js';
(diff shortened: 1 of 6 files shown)


packages/elements/src/icon/utils/SpriteLoader.ts
```diff
@@ -25,7 +25,7 @@ class SpriteLoader extends SVGLoader {
 
   /**
    * Load and cache the DOM sprite svg
-   * Get an svg fragment of DOM sprite svg
+   * Get a svg fragment of DOM sprite svg
```
wattachai-lseg: https://www.pristineword.com/grammar-an-svg/

packages/core/src/index.ts
```diff
@@ -47,6 +47,7 @@ export class LitElement extends BasicElement {
  */
 export { WarningNotice } from './notices/WarningNotice.js';
 export { DeprecationNotice } from './notices/DeprecationNotice.js';
+export { ErrorNotice } from './notices/ErrorNotice.js';
```
wattachai-lseg: This sounds like it works the same way with other Notice classes above. Let's come up with a new distinguished name.

packages/core/src/notices/ErrorNotice.ts
```diff
@@ -0,0 +1,47 @@
...
+   */
+  protected message: string;
+
+  /**
+   * Create an error notice to show in the console.
+   * @param message Error message to show in the console
+   */
+  constructor(message: string) {
+    this.message = message;
+  }
```
wattachai-lseg: In this situation, your static map keeping track of messages to be displayed is actually better than a simple flag. The reason is that this class handles rejected promises not a fixed error to be thrown. Give me a ping if you need more clue.

wattachai-lseg (approved):

PR #1309, opened 2026-08-06
fix(flag, icon): resolve empty value returned by getComputedStyle on Chromium 151+

wattachai-lseg:
## Description

Chromium only issue: starting from version 151, when Icon is slotted into an unregistered custom element, `getComputedStyle()` will always return empty string as a style value. It needs to wait for the registration of the parent custom element first. In practice, this happens when Icon class and its style are imported before the parent's ones as following:

```javascript
import '@refinitiv-ui/elements/icon';
import '@refinitiv-ui/elements/panel';

import '@refinitiv-ui/elements/icon/themes/halo/dark';
import '@refinitiv-ui/elements/panel/themes/halo/dark'; 
```

Additionally, this PR includes the following change:

* [Icon, Flag] Introduce fallback CDN prefix value
* [Flag] Delete duplicate quote removal from CSS custom property retrieval
* Add predefined prompt: `local-code-review.prompt.md`

## Checklist

### DIFF
--- a/.github/prompts/local-code-review.prompt.md
+++ b/.github/prompts/local-code-review.prompt.md
@@ -1,0 +1,35 @@
+---
+agent: agent
+description: 'Self-review current vs v7 branch - identify KISS violations, code issues, and PR feedback items'
+---
+
+# Self-Review: Current vs v7 Branch
+
+Please perform a comprehensive self-review of the current compared to v7 branch:
+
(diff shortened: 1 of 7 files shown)


Sarin-Udompanish (dismissed):
LGTM

Sarin-Udompanish (approved):

